PR #1, opened 2025-11-24
Add plugin manager API. 

nvddr:


examples/oxr/python/test_synthetic_hands.py
```diff
@@ -0,0 +1,143 @@
...
+import sys
+import time
+import os
+from pathlib import Path
+
+try:
+    import teleopcore.xrio as xrio
+    import teleopcore.oxr as oxr
+    import teleopcore.plugin_manager as pm
+except ImportError as e:
```
aristarkhovNV: most users should be more familiar with the default import error behavior, there is no need for customized handling

src/core/plugin_manager/cpp/plugin_manager.cpp
```diff
@@ -0,0 +1,205 @@
...
+    {
+        ROOT,
+        DEVICES,
+        DEVICE_ITEM
+    };
+
+    ParseState state = ParseState::ROOT;
+    DeviceInfo current_device;
+
+    while (std::getline(file, line))
```
aristarkhovNV: oh parsing yaml could be rabbit hole, consider using an existing library

examples/oxr/python/test_synthetic_hands.py
```diff
@@ -0,0 +1,143 @@
...
+    print()
+
+    try:
+        # 3. Initialize Plugin
+        print("[Step 3] Initializing plugin...")
+        
+        plugin.initialize("synthetic_hands")
+        print("  ✓ Plugin initialized and running")
+        
+        devices = plugin.query()
```
aristarkhovNV: `query_devices()` might be more accurate name for what it does

examples/oxr/python/test_synthetic_hands.py
```diff
@@ -0,0 +1,143 @@
...
+    print("[Step 2] Loading plugin...")
+    plugin = manager.load_plugin("controller_synthetic_hands")
+    print("  ✓ Plugin loaded")
+    print()
+
+    try:
+        # 3. Initialize Plugin
+        print("[Step 3] Initializing plugin...")
+        
+        plugin.initialize("synthetic_hands")
```
aristarkhovNV: why have separate initialize step after loading? Can the load perform initialization as well so you'd get a usable plugin right away?

src/core/plugin_manager/cpp/inc/plugin_manager/plugin.hpp
```diff
@@ -0,0 +1,58 @@
...
+    {
+        return m_name;
+    }
+
+private:
+    void start_process();
+    void stop_process();
+    bool is_running() const;
+
+    std::string m_name;
```
aristarkhovNV: this should be the job of plugin manager to carry the names for the plugins. The name seems to have no use to the plugin itself?

src/core/plugin_manager/cpp/inc/plugin_manager/plugin_manager.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+class PluginManager
+{
+public:
+    PluginManager();
+
+    /**
+     * @brief Add a directory to search for plugins.
+     */
+    void add_search_path(const std::string& path);
```
aristarkhovNV: consider providong all paths at once to the constructor to avoid managing the mutable state

src/core/plugin_manager/cpp/plugin_manager.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+PluginManager::PluginManager()
+{
+}
+
+void PluginManager::add_search_path(const std::string& path)
+{
+    if (fs::exists(path) && fs::is_directory(path))
+    {
+        m_search_paths.push_back(path);
```
aristarkhovNV: this should perhaps invalidate discovered plugins? But better consider not having dynamic add at all

src/core/plugin_manager/cpp/plugin_manager.cpp
```diff
@@ -0,0 +1,205 @@
...
+        {
+            std::cerr << "Error scanning directory " << base_path << ": " << e.what() << std::endl;
+        }
+    }
+    return names;
+}
+
+std::shared_ptr<Plugin> PluginManager::load_plugin(const std::string& name)
+{
+    auto it = m_discovered_plugins.find(name);
```
aristarkhovNV: this function has an implicit dependency on a previois discover_plugins() call which is makes the API more error prone. To enforce the correct usage `discover_plugins()` should return an immutable object that would expose the `load` function.

src/core/plugin_manager/cpp/inc/plugin_manager/plugin_manager.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "plugin.hpp"
+
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace teleop
+{
+namespace plugin_manager
```
aristarkhovNV: `namespace plugin_manager` seems unnecessary for `class PluginManager`. Placing the class inside the top level `teleop` as `teleop::PluginManager` could be sufficient

src/core/plugin_manager/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+cmake_minimum_required(VERSION 3.20)
+
+# Plugin Manager library
+add_library(teleop_plugin_manager STATIC
+    plugin.cpp
+    plugin_manager.cpp
+)
+
+# Include directory
+target_include_directories(teleop_plugin_manager PUBLIC
```
aristarkhovNV:
```suggestion
target_include_directories(teleop_plugin_manager INTERFACE
```

src/core/plugin_manager/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+)
+
+# Set standard C++17
+target_compile_features(teleop_plugin_manager PUBLIC cxx_std_17)
+
+# Link dependencies
+find_package(Threads REQUIRED)
+target_link_libraries(teleop_plugin_manager PUBLIC
+    Threads::Threads
+    yaml-cpp::yaml-cpp
```
aristarkhovNV: could be private since we are only using it from cpp
```suggestion
    PRIVATE
        yaml-cpp::yaml-cpp
```

src/plugins/controller_synthetic_hands/controller_synthetic_hands.cpp
```diff
@@ -1,209 +1,59 @@
...
-constexpr bool USE_SPACE_BASED_INJECTION = false;
-
-// Debug: Test curling with a hardcoded value (set to 0.0 to use real trigger input)
-constexpr float DEBUG_CURL_OVERRIDE = 0.0f; // Set to 0.5 or 1.0 to test curling
+#include <memory>
 
-// Keyboard fallback: Press 'L' for left hand curl, 'R' for right hand curl
-// (Useful if controller triggers aren't working)
-constexpr bool USE_KEYBOARD_FALLBACK = true;
+std::atomic<bool> g_running{ true };
```
aristarkhovNV:
```suggestion
static_assert(ATOMIC_BOOL_LOCK_FREE, "lock-free atomic bool is required for signal safety");
std::atomic<bool> g_running{ true };
```

src/plugins/controller_synthetic_hands/controller_synthetic_hands.cpp
```diff
@@ -1,209 +1,59 @@
...
 
 void signal_handler(int signal)
 {
     if (signal == SIGINT)
     {
+        std::cout << "\nReceived interrupt signal, shutting down..." << std::endl;
```
aristarkhovNV: `std::cout` is not signal safe

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+std::unique_ptr<SyntheticHandsPlugin> SyntheticHandsPlugin::Create(const std::string& plugin_root_id)
+{
+    try
+    {
+        // Use unique_ptr with custom deleter to allow private constructor
+        std::unique_ptr<SyntheticHandsPlugin> plugin(new SyntheticHandsPlugin(plugin_root_id));
+        return plugin;
+    }
+    catch (const std::exception& e)
```
aristarkhovNV: let the exception propagate to the caller so they can inspect the `what()`

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
```diff
@@ -0,0 +1,190 @@
...
+    // Initialize session
+    SessionConfig config;
+    config.app_name = "ControllerSyntheticHands";
+    config.extensions = { XR_NVX1_DEVICE_INTERFACE_BASE_EXTENSION_NAME, "XR_MND_headless", "XR_EXTX_overlay" };
+    config.use_overlay_mode = true; // Required for headless mode
+
+    m_session = Session::Create(config);
+    if (!m_session)
+    {
+        throw std::runtime_error("Failed to create session");
```
aristarkhovNV: the `Session::Create` could have thrown it for us

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    std::cout << "SyntheticHandsPlugin initialized and running" << std::endl;
+}
+
+SyntheticHandsPlugin::~SyntheticHandsPlugin()
+{
+    std::cout << "Shutting down SyntheticHandsPlugin..." << std::endl;
+
+    m_running = false;
+    if (m_thread.joinable())
```
aristarkhovNV: it must always be joinable

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
```diff
@@ -0,0 +1,190 @@
...
+    {
+        m_thread.join();
+    }
+
+    delete m_injector;
+    delete m_controllers;
+    delete m_session;
```
aristarkhovNV: use a smart pointer for these

src/core/plugin_manager/cpp/inc/plugin_manager/plugin_manager.hpp
nvddr: Our current top-level namespace is oxr. Will send a folow-up to rename.

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <iostream>
+
+// Toggle between space-based and pose-based hand injection
+// true = use controller spaces directly (primary method)
+// false = read controller poses and generate in world space (secondary method)
+constexpr bool USE_SPACE_BASED_INJECTION = false;
+
+std::unique_ptr<SyntheticHandsPlugin> SyntheticHandsPlugin::Create(const std::string& plugin_root_id)
+{
+    // Use unique_ptr with custom deleter to allow private constructor
```
aristarkhovNV: I don't see a custom deleter here, might just do a `return std::make_unique<...`. And with that a create function might be unnecessary, users could manage the class directly however they prefer - on stack, via unique or shared ptr, etc.

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+        m_injector.reset(HandInjector::Create(
+            handles.instance, handles.session, m_controllers->left_aim_space(), m_controllers->right_aim_space()));
+    }
+    else
+    {
+        m_injector.reset(
+            HandInjector::CreateWithReferenceSpace(handles.instance, handles.session, handles.reference_space));
+    }
+
+    if (!m_injector)
```
aristarkhovNV: this should never happen, the above code either creates one or throws

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+}
+
+SyntheticHandsPlugin::SyntheticHandsPlugin(const std::string& plugin_root_id) : m_root_id(plugin_root_id)
+{
+    std::cout << "Initializing SyntheticHandsPlugin with root: " << m_root_id << std::endl;
+
+    // Initialize session
+    SessionConfig config;
+    config.app_name = "ControllerSyntheticHands";
+    config.extensions = { XR_NVX1_DEVICE_INTERFACE_BASE_EXTENSION_NAME, "XR_MND_headless", "XR_EXTX_overlay" };
```
aristarkhovNV: are there defines for "XR_MND_headless", "XR_EXTX_overlay"?

examples/oxr/python/test_synthetic_hands.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # 4. Create Local Session for Reading
+        print("[Step 4] Creating reader session...")
+        extensions = [
+            "XR_KHR_convert_timespec_time",
+            "XR_MND_headless",
+            "XR_EXT_hand_tracking"
+        ]
+
+        with oxr.OpenXRSession.create("HandReader", extensions) as oxr_session:
```
aristarkhovNV: nit: this block could be combined with the outer `with` above for fewer nesting levels

```
with (
    manager.start(plugin_name, plugin_root_id) as plugin,
    oxr.OpenXRSession.create("HandReader", extensions) as oxr_session
):
```

src/core/plugin_manager/cpp/inc/plugin_manager/plugin.hpp
```diff
@@ -0,0 +1,49 @@
...
+    void stop();
+
+private:
+    void start_process();
+    void stop_process();
+    bool is_running() const;
+
+    std::string m_command;
+    std::string m_working_dir;
+    std::string m_plugin_root_id;
```
aristarkhovNV: these vars are only used once in the start_process, no need to keep then in the class

src/core/plugin_manager/cpp/plugin.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+void Plugin::stop()
+{
+    stop_process();
+}
+
+void Plugin::start_process(const std::string& command, const std::string& working_dir, const std::string& plugin_root_id)
+{
+#ifndef _WIN32
+    if (is_running())
```
aristarkhovNV, line 38: cannot be true

PR #4, opened 2025-11-26
Monitor plugin crash and throw exception

nvddr:
Monitors the plugin binary for crashes and propagates an exception signal through to python.

### DIFF
--- a/examples/oxr/python/test_synthetic_hands.py
+++ b/examples/oxr/python/test_synthetic_hands.py
@@ -10,6 +10,9 @@ This test:
 2. Queries available devices from the manifest
 3. Starts the OpenXR session
 4. Uses HandTracker to read the injected hands
+5. Periodically polls plugin health
+
+Note: Plugin crashes will raise pm.PluginCrashException
 """
 
 import sys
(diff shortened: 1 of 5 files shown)


examples/oxr/python/test_synthetic_hands.py
```diff
@@ -38,6 +39,16 @@ def run_test():
         print("Please build and install the project first.")
         return
 
+    try:
+        run_test_impl()
+    except RuntimeError as e:
+        print(f"\n✗ Plugin crashed: {e}")
+        print("This exception was raised when checking plugin health")
```
aristarkhovNV: it could be for other reasons too

nvddr: Made a custom exception for plugin crash

src/core/plugin_manager/python/plugin_manager_bindings.cpp
```diff
@@ -10,8 +10,25 @@ PYBIND11_MODULE(_plugin_manager, m)
...
+            try
+            {
+                if (p)
+                    std::rethrow_exception(p);
+            }
+            catch (const PluginCrashException& e)
+            {
+                plugin_crash_exc(e.what());
+            }
+        });
```
aristarkhovNV: according to https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html, for simple cases this could seem to be a one liner:

`py::register_exception<PluginCrashException>(module, "PluginCrashException");`

PR #5, opened 2025-12-02
Add controller tracker

nvddr:


examples/oxr/python/test_controller_tracker.py
```diff
@@ -0,0 +1,244 @@
...
+"""
+Test script for ControllerTracker with simplified API.
+
+Demonstrates:
+- Snapshot mode for continuous state queries
+- Event stream mode for input changes
+- Separate pose queries
+"""
+
+import sys
```
jiwenc-nv, line 12: Not an immediate blocker: but it would be nice to run this as a pytest that can be executed in CI. Is the blocker mainly around tooling to have runtime being available in CI?

src/core/oxr/cpp/oxr_session.cpp
```diff
@@ -173,4 +179,26 @@ bool OpenXRSession::create_reference_space()
...
+    if (session_ == XR_NULL_HANDLE)
+    {
+        std::cerr << "Cannot begin session: session handle is null" << std::endl;
+        return false;
+    }
+
+    XrSessionBeginInfo begin_info{ XR_TYPE_SESSION_BEGIN_INFO };
+    begin_info.primaryViewConfigurationType = XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO;
+
+    XrResult result = xrBeginSession(session_, &begin_info);
```
jiwenc-nv: wait, i thought headless mode doesn't require us to call `xrBeginSession` anymore.

it's kind of weird that controller API needs to explicit set:
`begin_info.primaryViewConfigurationType = XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO;`

src/core/xrio/cpp/controllertracker.cpp
```diff
@@ -0,0 +1,608 @@
...
+{
+    auto& queue = (hand == Hand::Left) ? left_event_queue_ : right_event_queue_;
+    auto& mutex = (hand == Hand::Left) ? left_queue_mutex_ : right_queue_mutex_;
+
+    std::lock_guard<std::mutex> lock(mutex);
+
+    // If queue is full, drop the oldest event
+    if (queue.size() >= MAX_QUEUE_SIZE)
+    {
+        queue.pop_front();
```
aristarkhovNV: Can loosing older messages be destructive? Consider blocking the enqueuing thread until more space is freed up.

src/core/xrio/cpp/controllertracker.cpp
```diff
@@ -0,0 +1,608 @@
...
+      thumbstick_action_(thumbstick_action),
+      thumbstick_click_action_(thumbstick_click_action),
+      squeeze_value_action_(squeeze_value_action),
+      trigger_value_action_(trigger_value_action),
+      left_grip_space_(left_grip_space),
+      right_grip_space_(right_grip_space),
+      left_aim_space_(left_aim_space),
+      right_aim_space_(right_aim_space),
+      left_hand_path_(left_hand_path),
+      right_hand_path_(right_hand_path),
```
aristarkhovNV: there is some code duplication between left and right handling across the impl. Consider making a dedicated tracker per hand.

src/core/xrio/cpp/controllertracker.cpp
```diff
@@ -0,0 +1,608 @@
...
+        core_funcs.xrDestroySpace(left_aim_space);
+        core_funcs.xrDestroySpace(right_aim_space);
+        core_funcs.xrDestroyActionSet(action_set);
+        throw;
+    }
+}
+
+ControllerTracker::Impl::~Impl()
+{
+    cleanup();
```
aristarkhovNV: cleanup() seems never being called elsewhere. Consider inlining cleanup body here in the dtor so we know it executes exactly once at the end

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+
+    std::shared_ptr<ITrackerImpl> initialize(const OpenXRSessionHandles& handles) override;
+
+private:
+    // Implementation class declaration (Pimpl idiom)
+    class Impl : public ITrackerImpl
+    {
+    public:
+        // Factory function for creating the Impl - returns nullptr on failure
+        static std::unique_ptr<Impl> create(const OpenXRSessionHandles& handles);
```
aristarkhovNV: use the constructor directly, throw in case of error

src/core/xrio/cpp/controllertracker.cpp
```diff
@@ -0,0 +1,608 @@
...
+    return "ControllerTracker";
+}
+
+const ControllerSnapshot& ControllerTracker::get_snapshot(Hand hand) const
+{
+    static const ControllerSnapshot empty_snapshot{};
+    auto impl = cached_impl_.lock();
+    if (!impl)
+        return empty_snapshot;
+    return impl->get_snapshot(hand);
```
aristarkhovNV, line 340: this is one reason to make the impl external and accessible directly - not having to forward a bunch of functions to the internal one

src/core/xrio/python/xrio_bindings.cpp
```diff
@@ -107,6 +108,69 @@ PYBIND11_MODULE(_xrio, m)
...
+        .def("get_snapshot", &core::ControllerTracker::get_snapshot, py::arg("hand"),
+             py::return_value_policy::reference_internal,
+             "Get current controller snapshot for specified hand (snapshot mode)")
+        .def("get_grip_pose", &core::ControllerTracker::get_grip_pose, py::arg("hand"),
+             py::return_value_policy::reference_internal, "Get grip pose for specified hand")
+        .def("get_aim_pose", &core::ControllerTracker::get_aim_pose, py::arg("hand"),
+             py::return_value_policy::reference_internal, "Get aim pose for specified hand")
+        .def("get_events", &core::ControllerTracker::get_events, py::arg("hand"),
+             "Get all queued input events for specified hand (clears the queue)")
+        .def("clear_events", &core::ControllerTracker::clear_events, py::arg("hand"), "Clear events for specified hand");
```
aristarkhovNV: this is redundant if get_events() also clears the queue

examples/oxr/python/test_controller_tracker.py
```diff
@@ -0,0 +1,244 @@
...
+        print()
+        
+        # Test 7: Run tracking loop
+        print("[Test 7] Running controller tracking loop (10 seconds)...")
+        print("Press buttons or move controls to see events!")
+        print()
+        
+        # Clear any initial events
+        controller_tracker.clear_events(xrio.Hand.Left)
+        controller_tracker.clear_events(xrio.Hand.Right)
```
aristarkhovNV: why

examples/oxr/python/test_controller_tracker.py
```diff
@@ -0,0 +1,244 @@
...
+print(f"Required extensions: {required_extensions if required_extensions else 'None (uses core OpenXR)'}")
+
+# Create OpenXR session
+oxr_session = oxr.OpenXRSession.create("ControllerTrackerTest", required_extensions)
+if oxr_session is None:
+    print("❌ Failed to create OpenXR session")
+    sys.exit(1)
+
+# Use context managers for proper RAII cleanup
+with oxr_session:
```
aristarkhovNV: Typically the `with:` both creates and manages the resource in one line
```
with create(blah) as x:
```

examples/oxr/python/test_controller_tracker.py
```diff
@@ -0,0 +1,244 @@
...
+    handles = oxr_session.get_handles()
+    
+    # Create teleop session
+    session = builder.build(handles)
+    if session is None:
+        print("❌ Failed to initialize")
+        sys.exit(1)
```
aristarkhovNV, line 56: We should really just let a c++ exception to propagate to python and handle all errors in a single `except` block. And in case of a sample like - not handle them explicitly at all since the default behavior to print the `what()` and terminate is all we need

examples/oxr/python/test_controller_tracker.py
```diff
@@ -0,0 +1,244 @@
...
+        
+        def print_controller_summary(hand_name, hand):
+            snapshot = controller_tracker.get_snapshot(hand)
+            print(f"  {hand_name} Controller:")
+            if snapshot.is_active:
+                print(f"    Status: ACTIVE")
+                
+                # Pose queries (independent of snapshot)
+                grip = controller_tracker.get_grip_pose(hand)
+                aim = controller_tracker.get_aim_pose(hand)
```
aristarkhovNV: what is the point of snapshot then. Also there is a race condition - live poses may not be active anymore at the time of querying

examples/oxr/python/test_controller_tracker.py
```diff
@@ -0,0 +1,244 @@
...
+print("✅ All tests passed")
+print("=" * 80)
+print()
+print("API Benefits:")
+print("  ✓ Simple: One tracker per controller")
+print("  ✓ Clean: All inputs as floats in array")
+print("  ✓ Event stream: get_events() returns all changes")
+print("  ✓ Snapshot: get_snapshot() for current state")
+print("  ✓ Poses: Separate from input events")
+print("=" * 80)
```
aristarkhovNV: seems redundant

src/core/oxr/cpp/oxr_session.cpp
```diff
@@ -62,6 +62,11 @@ std::shared_ptr<OpenXRSession> OpenXRSession::Create(const std::string& app_name
         return nullptr;
     }
 
+    if (!session->begin())
+    {
+        return nullptr;
```
aristarkhovNV: would be good to somehow report the reason, consider converting this entire function to throwing exceptions for each case above. Also may use ctor directly instead of Create()

src/core/oxr/cpp/oxr_session.cpp
```diff
@@ -173,4 +179,26 @@ bool OpenXRSession::create_reference_space()
     return true;
 }
 
+bool OpenXRSession::begin()
+{
+    if (session_ == XR_NULL_HANDLE)
```
aristarkhovNV: `OpenXRSession` should never exist with `session_` being null

src/core/oxr/cpp/oxr_session.cpp
```diff
@@ -173,4 +179,26 @@ bool OpenXRSession::create_reference_space()
...
+    begin_info.primaryViewConfigurationType = XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO;
+
+    XrResult result = xrBeginSession(session_, &begin_info);
+    if (XR_FAILED(result))
+    {
+        std::cerr << "Failed to begin OpenXR session: " << result << std::endl;
+        return false;
+    }
+
+    std::cout << "OpenXR session begun successfully" << std::endl;
```
aristarkhovNV: remove

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+        return "unknown";
+    }
+}
+
+// Simple snapshot of all controller input states
+// Used for snapshot mode - query current state at any time
+struct ControllerInputState
+{
+    // All inputs stored in array, indexed by ControllerInput enum
+    float values[static_cast<size_t>(ControllerInput::COUNT)];
```
aristarkhovNV: not using bools for buttons?

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+        return values[to_index(input)];
+    }
+
+    float& operator[](ControllerInput input)
+    {
+        return values[to_index(input)];
+    }
+
+    // Equality operator for detecting state changes
+    bool operator==(const ControllerInputState& other) const
```
aristarkhovNV: might use `= default` with operator spaceship `<=>`

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+    // Array access operators for convenience
+    float operator[](ControllerInput input) const
+    {
+        return values[to_index(input)];
+    }
+
+    float& operator[](ControllerInput input)
+    {
+        return values[to_index(input)];
+    }
```
aristarkhovNV: any benefit to array access with enum vs individual fields? especially if we add non-float data

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+    // All inputs stored in array, indexed by ControllerInput enum
+    float values[static_cast<size_t>(ControllerInput::COUNT)];
+
+    ControllerInputState()
+    {
+        for (size_t i = 0; i < static_cast<size_t>(ControllerInput::COUNT); ++i)
+        {
+            values[i] = 0.0f;
+        }
+    }
```
aristarkhovNV: skip manual ctor, initialize values on declaration with `= {};

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+    float get(ControllerInput input) const
+    {
+        return values[to_index(input)];
+    }
+
+    // Set value by enum
+    void set(ControllerInput input, float value)
+    {
+        values[to_index(input)] = value;
+    }
```
aristarkhovNV: why duplicate with []

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+// Controller pose data
+struct ControllerPose
+{
+    float position[3]; // x, y, z in meters
+    float orientation[4]; // x, y, z, w (quaternion)
+    bool is_valid;
+
+    ControllerPose() : position{ 0.0f, 0.0f, 0.0f }, orientation{ 0.0f, 0.0f, 0.0f, 1.0f }, is_valid(false)
+    {
+    }
```
aristarkhovNV: init fields on declaration

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+        XrSpace left_aim_space_;
+        XrSpace right_aim_space_;
+
+        // Paths for both hands
+        XrPath left_hand_path_;
+        XrPath right_hand_path_;
+
+        // Controller data for both hands
+        ControllerSnapshot left_snapshot_;
+        ControllerSnapshot right_snapshot_;
```
aristarkhovNV, line 124: if returned the whole snapshot from the update we wouldn't hane to cache it here

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+
+        // Previous state for detecting changes (event stream mode)
```
aristarkhovNV: should we be able to get input events directly from openxr?

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+        ControllerInputState prev_left_inputs_;
+        ControllerInputState prev_right_inputs_;
+
+        // Event queues for event stream mode (thread-safe) - separate for each hand
+        mutable std::mutex left_queue_mutex_;
+        mutable std::mutex right_queue_mutex_;
```
aristarkhovNV: seems no need to be mutable

src/core/xrio/cpp/controllertracker.cpp
```diff
@@ -0,0 +1,608 @@
...
+
+const ControllerPose& ControllerTracker::Impl::get_grip_pose(Hand hand) const
+{
+    return (hand == Hand::Left) ? left_snapshot_.grip_pose : right_snapshot_.grip_pose;
+}
+
+const ControllerPose& ControllerTracker::Impl::get_aim_pose(Hand hand) const
+{
+    return (hand == Hand::Left) ? left_snapshot_.aim_pose : right_snapshot_.aim_pose;
+}
```
aristarkhovNV: redundant if `get_snapshot` already exposes everything

src/core/xrio/cpp/controllertracker.cpp
```diff
@@ -0,0 +1,608 @@
...
+
+} // anonymous namespace
+
+// ============================================================================
+// ControllerTracker::Impl Implementation
+// ============================================================================
+
+// Factory function for creating the Impl
+std::unique_ptr<ControllerTracker::Impl> ControllerTracker::Impl::create(const OpenXRSessionHandles& handles)
```
aristarkhovNV: any chance to convert `create()` to ctor?

examples/oxr/python/test_controller_tracker.py
```diff
@@ -0,0 +1,244 @@
...
+    sys.exit(1)
+
+print("=" * 80)
+print("Controller Tracker Test - Simplified API")
+print("=" * 80)
+print()
+
+# Test 1: Create controller tracker (handles both controllers)
+print("[Test 1] Creating controller tracker...")
+controller_tracker = xrio.ControllerTracker()
```
aristarkhovNV, line 30: let `xrio.ControllerTracker()` take the hand argument and maintain separate trackers throughout the code

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+{
+    ControllerPose grip_pose;
+    ControllerPose aim_pose;
+    ControllerInputState inputs;
+    bool is_active;
+    XrTime timestamp;
+
+    ControllerSnapshot() : is_active(false), timestamp(0)
+    {
+    }
```
aristarkhovNV: init fields on declaration instead

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+// An event is just an input component and its new value
+struct ControllerInputEvent
+{
+    ControllerInput input;
+    float value;
+    XrTime timestamp;
+
+    ControllerInputEvent(ControllerInput i, float v, XrTime t) : input(i), value(v), timestamp(t)
+    {
+    }
```
aristarkhovNV: init fields on declaration instead

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+    std::vector<ControllerInputEvent> get_events(Hand hand);
+
+    // Clear queued events
+    void clear_events(Hand hand);
+
+protected:
+    // Internal lifecycle methods - only accessible via friend classes
+    friend class TeleopSession;
+
+    std::shared_ptr<ITrackerImpl> initialize(const OpenXRSessionHandles& handles) override;
```
aristarkhovNV, line 80: Something like `create_impl` might be a better name. `initialize()` usually implies an action to the object itself

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
```diff
@@ -0,0 +1,301 @@
...
+
+        // Previous state for detecting changes (event stream mode)
+        ControllerInputState prev_left_inputs_;
+        ControllerInputState prev_right_inputs_;
+
+        // Event queues for event stream mode (thread-safe) - separate for each hand
+        mutable std::mutex left_queue_mutex_;
+        mutable std::mutex right_queue_mutex_;
+        std::deque<ControllerInputEvent> left_event_queue_;
+        std::deque<ControllerInputEvent> right_event_queue_;
```
aristarkhovNV: these could have been vectors if we didn't have to cap the queue size, would be useful for no-copy return from `get_events`

src/core/xrio/cpp/controllertracker.cpp
```diff
@@ -0,0 +1,608 @@
...
+{
+    // Load core OpenXR functions dynamically using the provided xrGetInstanceProcAddr
+    OpenXRCoreFunctions core_funcs;
+    if (!core_funcs.load(handles.instance, handles.xrGetInstanceProcAddr))
```
aristarkhovNV: convert to ctor, throw on failure

src/core/xrio/cpp/controllertracker.cpp
```diff
@@ -0,0 +1,608 @@
...
+                                              right_aim_space, left_hand_path, right_hand_path, core_funcs));
+    }
+    catch (...)
+    {
+        // Clean up on construction failure
+        core_funcs.xrDestroySpace(left_grip_space);
+        core_funcs.xrDestroySpace(right_grip_space);
+        core_funcs.xrDestroySpace(left_aim_space);
+        core_funcs.xrDestroySpace(right_aim_space);
+        core_funcs.xrDestroyActionSet(action_set);
```
aristarkhovNV: convert these to smart pointers so no explicit `try-catch` is required

examples/oxr/python/test_controller_tracker.py
nvddr: I will defer refactoring tests to a folllowup.

src/core/xrio/cpp/controllertracker.cpp
nvddr: no, this is intentional. Similar in line to what Jakob was thinking for the openxr events API, it buffers events but if you're querying it with a large delay you will lose messages. Overall lossy, but best effort to be non-lossy.

src/core/xrio/cpp/controllertracker.cpp
nvddr: This is by design and matches other trackers. Will make it a follow-up if we want to change the architecture of the tracker API.

src/core/xrio/cpp/controllertracker.cpp
aristarkhovNV: OpenXR would let the app know if some events were dropped so the app could react accordingly. Silently dropping may create some debugging headaches.

src/core/xrio/cpp/controllertracker.cpp
nvddr: You can't do the binding twice, they need to be attached in one shot.

src/core/oxr/cpp/oxr_session.cpp
nvddr: We won't get controller inputs without begin session unfortunately. Removed the specific stereo line and it will use default initialized. Doesn't seem to matter.

src/core/xrio/cpp/controllertracker.cpp
nvddr: Done

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
nvddr: Done

src/core/xrio/python/xrio_bindings.cpp
nvddr: I updated it to get rid of the whole queue thing. We can add it later if we need any queuing. Currently matches the openxr API - on sync actions, it takes a snapshot. We return this snapshot when queried.

examples/oxr/python/test_controller_tracker.py
nvddr: These events could be very old, on startup start looking at events starting from latest. but we removed the queue mechanism, So this is outdated.

examples/oxr/python/test_controller_tracker.py
nvddr: Snapshot now contains everything

src/core/oxr/cpp/oxr_session.cpp
nvddr: Done

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
nvddr: We need to cache in the impl because they are updated when the xr session update is called. Any subsequent queries just returns the cached value.

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
nvddr: We don't get inputs on query, we get the inputs from openxr when the xrio session is updated. This syncrhronizes everything.

src/core/xrio/cpp/inc/xrio/controllertracker.hpp
nvddr: Currently matches other files, we can do a followup.

src/core/xrio/cpp/controllertracker.cpp
nvddr: I don't understand this comment.. could you elaborate?

examples/oxr/python/test_controller_tracker.py
nvddr: We can't do this, because the bindings happen in the tracker impl initialization, and we cannot do it twice.

src/core/xrio/cpp/controllertracker.cpp
aristarkhovNV: Store `left_grip_space`, `right_grip_space`, etc in a `std::unique_ptr` with a custom deleter. It will ensure the resources are released when leaving the scope. And you can `std::move` them to transfer ownership explicitly. This makes code exception-safe without having to manually write any try-catch blocks.

nvddr: Done

src/core/oxr/cpp/oxr_session.cpp
jiwenc-nv: @nv-jakob looked it up and not setting it seems to be invalid:
`XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO`

jiwenc-nv: ideally we need a new config type: `XR_VIEW_CONFIGURATION_TYPE_NO_DISPLAY`

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +89,140 @@ struct OpenXRCoreFunctions
...
+    }
+};
+
+struct XrSpaceDeleter
+{
+    PFN_xrDestroySpace destroy_func;
+
+    XrSpaceDeleter() : destroy_func(nullptr)
+    {
+    }
```
aristarkhovNV: See if we can get rid of these default deleters. If a unique_ptr with such empty deleter gets assigned a new resource it won't be able to properly release that resource.

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +89,140 @@ struct OpenXRCoreFunctions
...
+    explicit XrSpaceDeleter(PFN_xrDestroySpace func) : destroy_func(func)
+    {
```
aristarkhovNV, line 180: assert the func is not null

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +89,140 @@ struct OpenXRCoreFunctions
...
+        if (actionSet && *actionSet != XR_NULL_HANDLE && destroy_func)
+        {
+            destroy_func(*actionSet);
+        }
+        delete actionSet;
+    }
+};
+
+struct XrSpaceDeleter
```
aristarkhovNV: could be one template deleter for all types

src/core/oxr/cpp/oxr_session.cpp
nvddr: Addressed

nv-jakob: Call `xrEnumerateViewConfigurations` and pick the first entry in the list and put it in here.

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +89,140 @@ struct OpenXRCoreFunctions
...
+
+// Create an action set with automatic cleanup
+inline XrActionSetPtr createActionSet(const OpenXRCoreFunctions& funcs,
+                                      XrInstance instance,
+                                      const XrActionSetCreateInfo* createInfo)
+{
+    if (!funcs.xrCreateActionSet)
+    {
+        return XrActionSetPtr(nullptr, XrActionSetDeleter(funcs.xrDestroyActionSet));
+    }
```
aristarkhovNV: this seems like an invalid situation

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +89,140 @@ struct OpenXRCoreFunctions
...
+
+    XrActionSet* actionSet = new XrActionSet(XR_NULL_HANDLE);
+    XrResult result = funcs.xrCreateActionSet(instance, createInfo, actionSet);
+
+    if (XR_FAILED(result))
+    {
+        delete actionSet;
+        return XrActionSetPtr(nullptr, XrActionSetDeleter(funcs.xrDestroyActionSet));
```
aristarkhovNV: throw here so you don't have to construct an empty ptr

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +89,140 @@ struct OpenXRCoreFunctions
...
+inline XrSpacePtr createActionSpace(const OpenXRCoreFunctions& funcs,
+                                    XrSession session,
+                                    const XrActionSpaceCreateInfo* createInfo)
+{
+    if (!funcs.xrCreateActionSpace)
+    {
+        return XrSpacePtr(nullptr, XrSpaceDeleter(funcs.xrDestroySpace));
+    }
+
+    XrSpace* space = new XrSpace(XR_NULL_HANDLE);
```
aristarkhovNV: do not allocate from heap, unique_ptr should be able to store XrSpace directly

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +89,140 @@ struct OpenXRCoreFunctions
...
+    XrActionSetDeleter() : destroy_func(nullptr)
+    {
+    }
+    explicit XrActionSetDeleter(PFN_xrDestroyActionSet func) : destroy_func(func)
+    {
+    }
+
+    void operator()(XrActionSet* actionSet) const
+    {
+        if (actionSet && *actionSet != XR_NULL_HANDLE && destroy_func)
```
aristarkhovNV: never supposed to have no destroy_func

src/core/oxr/cpp/oxr_session.cpp
```diff
@@ -173,4 +180,19 @@ bool OpenXRSession::create_reference_space()
...
 
+bool OpenXRSession::begin()
+{
+    XrSessionBeginInfo begin_info{ XR_TYPE_SESSION_BEGIN_INFO };
+    // primaryViewConfigurationType not set for headless tracking sessions
+
+    XrResult result = xrBeginSession(session_, &begin_info);
+    if (XR_FAILED(result))
+    {
+        std::cerr << "Failed to begin OpenXR session: " << result << std::endl;
```
aristarkhovNV: Throw right here so you don't have to check the result later and this will let you also attach the error code

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +92,186 @@ struct OpenXRCoreFunctions
...
+// Custom deleter for OpenXR handles
+template <typename HandleType>
+struct OpenXRHandleDeleter
+{
+    std::function<void(HandleType)> destroy_func;
+
+    void operator()(HandleType handle) const
+    {
+        if (handle != XR_NULL_HANDLE && destroy_func)
+        {
```
aristarkhovNV, line 126: I'd be bad if we've got a handle to destroy and not the destroy func
```suggestion
        if (handle != XR_NULL_HANDLE)
        {
            assert(destroy_func);
```

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +92,186 @@ struct OpenXRCoreFunctions
...
+    XrResult result = funcs.xrCreateActionSet(instance, createInfo, &actionSet);
+
+    if (XR_FAILED(result))
+    {
+        throw std::runtime_error("Failed to create action set: " + std::to_string(result));
+    }
+
+    auto deleter = [destroyFunc = funcs.xrDestroyActionSet](XrActionSet handle)
+    {
+        if (destroyFunc)
```
aristarkhovNV: should always have the destroy func

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_funcs.hpp
```diff
@@ -63,8 +92,186 @@ struct OpenXRCoreFunctions
...
+{
+    XrActionSet actionSet = XR_NULL_HANDLE;
+    XrResult result = funcs.xrCreateActionSet(instance, createInfo, &actionSet);
+
+    if (XR_FAILED(result))
+    {
+        throw std::runtime_error("Failed to create action set: " + std::to_string(result));
+    }
+
+    auto deleter = [destroyFunc = funcs.xrDestroyActionSet](XrActionSet handle)
```
aristarkhovNV, line 220:
```suggestion
    assert(funcs.xrDestroyActionSet);
    auto deleter = [destroyFunc = funcs.xrDestroyActionSet](XrActionSet handle)
```

PR #9, opened 2025-12-08
Devdeepr/retarget samples

nvddr:


src/core/retargeting_engine/python/retargeters/isaac_lab/README.md
rwiltz, line 1: I don't think we want the retargeters to be specific to an integration IE Isaac Lab or Isaac OS for instance. I expect that we want to get as much re-use from retargeters across platforms as possible. Can we move this directory to `src/core/retargeting_engine/python/retargeters/<grippers/dex_hands/locomotion/"etc">`

examples/teleop/python/config/dex_retargeting/hand_right_config.yml
```diff
@@ -0,0 +1,24 @@
...
+  scaling_factor: 1.2
+  target_joint_names:
+  - R_thumb_proximal_yaw_joint
+  - R_thumb_proximal_pitch_joint
+  - R_index_proximal_joint
+  - R_middle_proximal_joint
+  - R_ring_proximal_joint
+  - R_pinky_proximal_joint
+  type: DexPilot
+  urdf_path: /tmp/retarget_inspire_white_right_hand.urdf
```
rwiltz, line 23: Replace with robot hand?

PR #10, opened 2025-12-09
Add flatbuffer data schema

jiwenc-nv:


src/core/messages/test_tensor.cpp
```diff
@@ -0,0 +1,113 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Unit tests for the generated Tensor FlatBuffer message.
+
+#include <catch2/catch_test_macros.hpp>
+
+// Include Holoscan tensor header (from native_include in tensor.fbs).
+#include "holoscan/core/domain/tensor.hpp"
```
aristarkhovNV: use angle brackets <> for non-local includes
```suggestion
#include <holoscan/core/domain/tensor.hpp>
```

cmake/GenerateFlatBuffers.cmake
```diff
@@ -0,0 +1,74 @@
...
+  # Ensure flatc is available from FetchContent.
+  if(NOT TARGET flatc)
+    message(FATAL_ERROR "flatc target not found. Make sure FlatBuffers is fetched via FetchContent before calling this function.")
+  endif()
+
+  # Use generator expression to get flatc executable path at build time.
+  set(FLATC_EXECUTABLE $<TARGET_FILE:flatc>)
+
+  # Find all .fbs files in input directory.
+  file(GLOB FBS_FILES RELATIVE ${INPUT_DIR} "${INPUT_DIR}/*.fbs")
```
aristarkhovNV:
```suggestion
  file(GLOB FBS_FILES RELATIVE ${INPUT_DIR} CONFIGURE_DEPENDS "${INPUT_DIR}/*.fbs")
```

cmake/GenerateFlatBuffers.cmake
```diff
@@ -0,0 +1,74 @@
...
+  if(NOT FBS_FILES)
+    message(WARNING "No .fbs files found in ${INPUT_DIR}")
```
aristarkhovNV: should be fatal error probably?

cmake/GenerateFlatBuffers.cmake
```diff
@@ -0,0 +1,74 @@
...
+    set(${OUT_VAR} "" PARENT_SCOPE)
+    return()
+  endif()
+
+  # Create output directory if it doesn't exist.
+  if(NOT EXISTS ${OUTPUT_DIR})
+    file(MAKE_DIRECTORY ${OUTPUT_DIR})
+  endif()
```
aristarkhovNV: redundant, it gets created as part of custom command below

src/core/messages/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+
+add_library(teleopcore_messages INTERFACE)
+add_library(teleopcore::messages ALIAS teleopcore_messages)
+
+add_dependencies(teleopcore_messages messages_header_generation)
+
+# Include directories for generated headers and FlatBuffers runtime.
+target_include_directories(teleopcore_messages INTERFACE
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/generated>
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/schemas>
```
aristarkhovNV, line 38: schemas does not seem to contain any includes, does it need to be in the list?

src/core/messages/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+# ==============================================================================
+# Generate C++ headers from FlatBuffer schema files.
+
+include(${CMAKE_SOURCE_DIR}/cmake/GenerateFlatBuffers.cmake)
+
+# Generate headers from schemas.
+generate_flatbuffer_headers(
+  MESSAGES_GENERATED_HEADERS
+  ${CMAKE_CURRENT_SOURCE_DIR}/schemas
+  ${CMAKE_CURRENT_BINARY_DIR}/generated
```
aristarkhovNV: force users to prepend the lib name in their #include
```suggestion
  ${CMAKE_CURRENT_BINARY_DIR}/generated/teleopcore_messages
```

src/core/messages/test_tensor.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+// Include generated FlatBuffer header.
+#include "tensor_generated.h"
```
aristarkhovNV:
```suggestion
#include <teleopcore_messages/tensor_generated.h>
```

src/core/messages/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20)
+project(teleopcore_messages VERSION 1.0.0 LANGUAGES CXX)
```
aristarkhovNV: `project` statement is optional for nested cmakelists

src/core/messages/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+# ==============================================================================
+# Tensor Message with Holoscan SDK
+# ==============================================================================
+# Extended library that includes Holoscan tensor support.
+
+# Find Holoscan SDK (holoscan_DIR is set by deps/third_party after FetchContent).
+find_package(holoscan REQUIRED CONFIG)
+
+add_library(tensor_message INTERFACE)
+add_library(teleopcore::tensor_message ALIAS tensor_message)
```
aristarkhovNV: What is the difference from `teleopcore::messages` target? Seems a bit confusing that a single CMakeLists defines two consumable targets - can we make it one or split in two separate cmakelists if two targets are really necessary

src/core/messages/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+)
+
+message(STATUS "tensor_message target enabled with Holoscan SDK")
+
+# ==============================================================================
+# Tensor Unit Tests
+# ==============================================================================
+# Unit tests for FlatBuffer message types using Catch2.
+
+if(BUILD_TESTING)
```
aristarkhovNV, line 53: can we factor these out into a separate tensor_tests directory with its own cmakelists file

jiwenc-nv: Maybe just a general tests folder? I do want to add more messages and tests as well.

src/core/messages/test_tensor.cpp
jiwenc-nv: Turns out we don't need this anyway.

cmake/GenerateFlatBuffers.cmake
jiwenc-nv: done.

cmake/GenerateFlatBuffers.cmake
jiwenc-nv: done.

cmake/GenerateFlatBuffers.cmake
jiwenc-nv: done.

src/core/messages/CMakeLists.txt
jiwenc-nv: good suggestion to add a bit prepend for the `lib` name here. I updated this to `teleopcore/messages` instead.

src/core/messages/CMakeLists.txt
jiwenc-nv: removed

src/core/messages/test_tensor.cpp
jiwenc-nv: not sure we want to use the angle brackets <> for this? It's all from the teleop lib anyway.

src/core/messages/CMakeLists.txt
jiwenc-nv: removed.

src/core/messages/CMakeLists.txt
jiwenc-nv: We discussed this offline. The `tensor_message` library (also the alias) is not needed at the moment as we don't need to support holoscan linkage at the moment.

CMakeLists.txt
```diff
@@ -35,12 +35,13 @@ endif()
...
 option(BUILD_PYTHON_BINDINGS "Build Python bindings" ON)
 option(BUILD_PLUGINS "Build plugins" ON)
+option(BUILD_TESTING "Build unit tests" ON)
 option(ENABLE_EXPERIMENTAL_WINDOWS_BUILD "Enable experimental Windows build support (source code only)" OFF)
 
 # Windows support warning
 if(WIN32)
     if(NOT ENABLE_EXPERIMENTAL_WINDOWS_BUILD)
-        message(FATAL_ERROR 
+        message(FATAL_ERROR
```
nvddr, line 44: Why are we not building on windows?

src/core/messages/schemas/tensor.fbs
```diff
@@ -0,0 +1,55 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// FlatBuffers schema demonstrating native_type approach with holoscan::Tensor.
```
nvddr, line 4: Stale comment. Also add a comment link to the source fbs schema with a note that these should be in sync.

CMakeLists.txt
aristarkhovNV: we allow building but you have to acknowledge it is experimental and unsupported

src/core/messages/CMakeLists.txt
aristarkhovNV: still here

src/core/messages/tests/test_tensor.cpp
```diff
@@ -0,0 +1,110 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Unit tests for the generated Tensor FlatBuffer message.
+
+#include <catch2/catch_test_macros.hpp>
+
+// Include generated FlatBuffer header.
+#include "teleopcore/messages/tensor_generated.h"
```
aristarkhovNV, line 9:
```suggestion
#include <teleopcore/messages/tensor_generated.h>
```

src/core/messages/test_tensor.cpp
aristarkhovNV: the reasoning is to distinguish between local headers picked up locally by a relative path vs. found via include paths coming from other cmake targets -- https://docs.google.com/document/d/1zqp4TOaQYqzC2EAsxQc0UTsM4skwHmMF5zCxDWVf9xU/edit?tab=t.0

PR #13, opened 2025-12-11
Add Python binding for the schemas

jiwenc-nv:


.github/workflows/build-ubuntu.yml
```diff
@@ -36,5 +36,12 @@ jobs:
     - name: Build
       run: cmake --build build --parallel
 
-    - name: Run Tests
+    - name: Run Native Tests
       run: ctest --test-dir build --output-on-failure --parallel
+
+    - name: Run Python Tests
+      run: |
+        uv run --python 3.11 \
```
aristarkhovNV: since we target specific python version with pybind, that version should be specified in pyproject toml, no need to pass to uv here

src/core/python/pyproject.toml
aristarkhovNV, line 13: probably need to be exact

src/core/schema/python/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+# Ensure generated headers are available.
+add_dependencies(schema_py schema_header_generation)
+
+set_target_properties(schema_py PROPERTIES
+    OUTPUT_NAME "_schema"
+    LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/python_package/teleopcore/schema"
+)
+
+message(STATUS "TeleopCore schema Python bindings configured")
```
aristarkhovNV: lets skip the extra status logging to reduce the noise

src/core/schema/CMakeLists.txt
```diff
@@ -42,3 +42,12 @@ target_link_libraries(teleopcore_schema INTERFACE
...
 
 message(STATUS "TeleopCore schema library configured with FlatBuffer")
+
+# ==============================================================================
+# Python Bindings
+# ==============================================================================
+# Python bindings for schema types (optional, controlled by parent).
+
+if(BUILD_PYTHON_BINDINGS)
+    add_subdirectory(python)
```
aristarkhovNV, line 15: all the cpp related stuff above (together with headers generation) should also go into its own cmakelists under cpp/CMakeLists.txt, this is what other modules do

src/core/python/pyproject.toml
jiwenc-nv: actually that's accurate.

We do support python versions starting 3.11. I tested manually before with 3.12, 3.13, etc.

jiwenc-nv: We can configure it here:
https://github.com/NVIDIA/IsaacTeleop/blob/main/CMakeLists.txt#L17

Ideally, we should add tests for all Python versions. with something like:

```
cmake -B build -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DTELEOPCORE_PYTHON_VERSION=${{ matrix.python_version }}
```

aristarkhovNV: I'm being told that a pybind11 extension built against one CPython version is not guaranteed to work on future CPython versions. You normally need to rebuild wheels for each supported Python minor version.

jiwenc-nv: Ah. I think you are right. Then we should make the `pyproject.toml` file a template and fill in Python version with the CMake pipeline.

jiwenc-nv: I will adress this in a separate PR.

src/core/schema/python/CMakeLists.txt
jiwenc-nv: done.

src/core/schema/CMakeLists.txt
jiwenc-nv: done.

src/core/schema/python/schema_module.cpp
```diff
@@ -0,0 +1,23 @@
...
+namespace py = pybind11;
+
+PYBIND11_MODULE(_schema, m) {
+    m.doc() = "TeleopCore Schema - FlatBuffer message types for teleoperation";
+
+    // Bind tensor types (enums, structs, TensorT).
+    core::bind_tensor(m);
+
+    // Bind pose types (PoseT).
+    core::bind_pose(m);
```
aristarkhovNV, line 21: I wonder if fbs could compiled directly to python so we don't have to manually provide bindings for generated cpp types

jiwenc-nv: I think we still need the Python binding. So the the serialization always happen in the native land, also passing the data object from C++ to Python doesn't require serialization.

PR #14, opened 2025-12-11
Create manus plugin

rwiltz:
The manus plugin uses the Manus SDK to query glove data, and polls the OpenXR for controller pose. It then uses an offset aim pose as the wrist joint pose, and uses that as the root for all the subsequent joint poses from Manus. Then it pushes this back to the CloudXR runtime.

### DIFF
--- a/deps/cloudxr/openxr_extensions/CMakeLists.txt
+++ b/deps/cloudxr/openxr_extensions/CMakeLists.txt
@@ -11,7 +11,3 @@ target_include_directories(teleop_openxr_extensions INTERFACE
 install(FILES XR_NVX1_device_interface.h
     DESTINATION include
 )
-
-install(TARGETS teleop_openxr_extensions
-    EXPORT TeleopTargets
-)
--- a/src/plugins/manus/core/manus_hand_tracking_plugin.cpp
+++ b/src/plugins/manus/core/manus_hand_tracking_plugin.cpp
@@ -133,9 +133,9 @@ void ManusTracker::initialize(const std::string& app_name) noexcept(false)
         // Require the push device extension
         config.extensions = { XR_NVX1_DEVICE_INTERFACE_BASE_EXTENSION_NAME, XR_MND_HEADLESS_EXTENSION_NAME };
 #if defined(_WIN32)
-        config.extensions.push_back("XR_KHR_win32_convert_performance_counter_time");
+        config.extensions.push_back(XR_KHR_WIN32_CONVERT_PERFORMANCE_COUNTER_TIME_EXTENSION_NAME);
 #else
-        config.extensions.push_back("XR_KHR_convert_timespec_time");
+        config.extensions.push_back(XR_KHR_CONVERT_TIMESPEC_TIME_EXTENSION_NAME);
 #endif
         // Overlay mode required for headless
         config.use_overlay_mode = true;
--- a/src/plugins/manus/CMakePresets.json
+++ b/src/plugins/manus/CMakePresets.json
@@ -1,0 +1,36 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": { "major": 3, "minor": 22, "patch": 0 },
+  "configurePresets": [
+    {
+      "name": "linux-manus-default",
+      "displayName": "Linux Manus Default (RelWithDebInfo)",
+      "generator": "Unix Makefiles",
+      "binaryDir": "${sourceDir}/build-manus-default",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "MANUS_SDK_ROOT": "${sourceDir}/ManusSDK"
+      }
+    },
+    {
+      "name": "linux-manus-debug",
+      "displayName": "Linux Manus Debug",
+      "inherits": "linux-manus-default",
+      "binaryDir": "${sourceDir}/build-manus-debug",
+      "cacheVariables": { "CMAKE_BUILD_TYPE": "Debug" }
+    },
+    {
+      "name": "linux-manus-release",
+      "displayName": "Linux Manus Release",
+      "inherits": "linux-manus-default",
+      "binaryDir": "${sourceDir}/build-manus-release",
+      "cacheVariables": { "CMAKE_BUILD_TYPE": "Release" }
+    }
+  ],
+  "buildPresets": [
+    { "name": "manus-default", "configurePreset": "linux-manus-default" },
+    { "name": "manus-debug",   "configurePreset": "linux-manus-debug" },
+    { "name": "manus-release", "configurePreset": "linux-manus-release" }
+  ],
+  "testPresets": []
+}


src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -245,6 +368,132 @@ void ManusTracker::OnSkeletonStream(const SkeletonStreamInfo* skeleton_stream_in
...
+
+            {
+                std::lock_guard<std::mutex> lock(s_instance->m_controller_mutex);
+                if (is_left_glove)
+                {
+                    if (s_instance->m_latest_left.aim_valid)
+                    {
+                        XrPosef raw_pose = s_instance->m_latest_left.aim_pose;
+                        // Mirror Right Hand Offset
+                        XrPosef offset_pose = { { -0.70710678f, -0.5f, 0.0f, 0.5f }, { -0.1f, 0.02f, -0.02f } };
```
rwiltz: Follow up MR will make these offsets configurable

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -16,6 +18,42 @@
 #include <unordered_map>
 #include <vector>
 
+namespace
+{
+constexpr XrPosef multiply_poses(const XrPosef& a, const XrPosef& b)
+{
+    XrPosef result{};
```
nv-jakob: Follow up, some small helper C++ library for these type of things?

src/plugins/manus/XR_NVX1_device_interface.h
```diff
@@ -0,0 +1,124 @@
+// Copyright 2025, NVIDIA CORPORATION.
+// SPDX-License-Identifier: Apache-2.0 or MIT
+/*!
+ * @file
+ * @brief Header for OpenXR device interface.
+ * @ingroup external_openxr
+ */
```
nv-jakob: Oh his header isn't available elsewhere in the repo? Can it be refactored to a shared library?

nv-jakob: Also one that isn't effected by clang format. :D

rwiltz: Yeah looks like it was originally added under synthetic hands example. I'll move it (here or in a follow up depending on how much we want to get this functionality in)

nv-jakob: Follow up is fine.

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
rwiltz: Yes!

src/plugins/manus/XR_NVX1_device_interface.h
nvddr: We can have some utilities for plugin writers, but the plugin ecosystem is intended to be fully standalone binaries, so it's up to the plugin authors to do what they want. We should provide the device extension in a folder (almost like documentation), I think there's some benefit to not have the plugin build be complicated by our repo structure?

src/plugins/manus/CMakeLists.txt
```diff
@@ -135,14 +154,45 @@ set_target_properties(ManusHandTrackerPrinter PROPERTIES
...
-  # Allow running from build dir; resolve libs next to the exe
   SKIP_BUILD_RPATH NO
+  BUILD_RPATH "${_ISAAC_MANUS_EFFECTIVE_BUILD_RPATH_STR}"
+  RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
 )
 
-# Apply the same effective BUILD_RPATH to the CLI for consistency
-set_target_properties(ManusHandTrackerPrinter PROPERTIES
+# Main plugin executable
+add_executable(manus_hand_plugin
```
aristarkhovNV: This cmakelists defines several targets: couple libraries and the executable. Consider splitting it up into separate componet directories per https://docs.google.com/document/d/1zqp4TOaQYqzC2EAsxQc0UTsM4skwHmMF5zCxDWVf9xU/edit?tab=t.0

src/plugins/manus/controllers.cpp
```diff
@@ -0,0 +1,271 @@
...
+    }
+    return controllers;
+}
+
+Controllers::~Controllers()
+{
+    cleanup();
+}
+
+bool Controllers::initialize(XrInstance instance, XrSession session, XrSpace reference_space)
```
aristarkhovNV: consider adopting RAII - all initialization is to performed by the constructor which throws in case of an error. That way if you successfully created the object it would be guaranteed to be valid, will no additional checks.

src/plugins/manus/controllers.cpp
```diff
@@ -0,0 +1,271 @@
...
+    if (XR_FAILED(result))
+    {
+        std::cerr << "Failed to attach action sets: " << result << std::endl;
+        return false;
+    }
+
+    return true;
+}
+
+bool Controllers::setup_action_spaces(XrSession session)
```
aristarkhovNV: Lets adopt exceptions instead of return codes, especially the bool

src/plugins/manus/controllers.cpp
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    is_valid = (location.locationFlags & XR_SPACE_LOCATION_POSITION_VALID_BIT) &&
+               (location.locationFlags & XR_SPACE_LOCATION_ORIENTATION_VALID_BIT);
+    pose = location.pose;
+
+    return true;
+}
+
+void Controllers::cleanup()
```
aristarkhovNV, line 215: should be nothrow(), also do we really need separate cleanup() versus doing everything in dtor

src/plugins/manus/XR_NVX1_device_interface.h
nv-jakob: Yeah, hmm I think specifically for the OpenXR headers we should put them in like a "openxr_includes" directory and all collected together. Then if you want to use them outside of this repo, you can just copy them.

src/plugins/manus/hand_injector.cpp
```diff
@@ -0,0 +1,113 @@
...
+    data.jointCount = XR_HAND_JOINT_COUNT_EXT;
+    data.jointLocations = joints;
+
+    return XR_SUCCEEDED(pfn_push_(left_device_, &data));
+}
+
+bool HandInjector::push_right(const XrHandJointLocationEXT* joints, XrTime timestamp)
+{
+    if (right_device_ == XR_NULL_HANDLE)
+        return false;
```
aristarkhovNV: is this an error? Consider throwing if this is a runtime error or assert() if this is supposed to be a logic error

src/plugins/manus/app/main.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <thread>
+
+using namespace isaacteleop::plugins::manus;
+
+// Use atomic<bool> with relaxed ordering for signal safety
+std::atomic<bool> g_stop_requested{ false };
+
+void signal_handler(int signal)
+{
+    if (signal == SIGINT)
```
aristarkhovNV: are we doing any particular cleanup? May just keep the default kernel action of terminating the process immediately, no need for a custom signal handler. Otherwise should be handling SIGTERM as well

src/plugins/manus/app/main.cpp
```diff
@@ -0,0 +1,63 @@
...
+    {
+        auto frame_start = std::chrono::steady_clock::now();
+
+        tracker.update();
+
+        std::this_thread::sleep_until(frame_start + target_frame_duration);
+    }
+
+    std::cout << "Stopping..." << std::endl;
+    tracker.cleanup();
```
aristarkhovNV: destructor should do it, no need for an explicit cleanup

src/plugins/manus/core/ManusHandTrackingPlugin.cpp
```diff
@@ -5,6 +5,8 @@
 
 #include "ManusSDK.h"
 #include "ManusSDKTypeInitializers.h"
+#include "hand_injector.hpp"
+#include "session.hpp"
```
aristarkhovNV: would be good to place these headers in a subdirectory named after the library (or cmake target) they belong to. Unless they are actually local - in which case they should be at the top, above the manus sdk includes.

src/plugins/manus/ManusHandTrackingPlugin.cpp
```diff
@@ -5,6 +5,8 @@
 
 #include "ManusSDK.h"
```
aristarkhovNV: 3rd party headers found via include paths should be in angle brackets<>

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -33,6 +71,91 @@ ManusTracker::~ManusTracker()
...
+    try
+    {
+        SessionConfig config;
+        config.app_name = app_name;
+        // Require the push device extension
+        config.extensions = { XR_NVX1_DEVICE_INTERFACE_BASE_EXTENSION_NAME, XR_MND_HEADLESS_EXTENSION_NAME };
+#if defined(_WIN32)
+        config.extensions.push_back("XR_KHR_win32_convert_performance_counter_time");
+#else
+        config.extensions.push_back("XR_KHR_convert_timespec_time");
```
aristarkhovNV: use a macro for ext names

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -33,6 +71,91 @@ ManusTracker::~ManusTracker()
...
+        if (!m_controllers)
+        {
+            std::cerr << "Failed to create controllers" << std::endl;
+            return false;
+        }
+
+        std::cout << "OpenXR session, HandInjector and Controllers initialized" << std::endl;
+        return true;
+    }
+    catch (const std::exception& e)
```
aristarkhovNV, line 170: let it propagate and be handled by the caller

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -245,6 +368,127 @@ void ManusTracker::OnSkeletonStream(const SkeletonStreamInfo* skeleton_stream_in
             s_instance->output_map[orient_key][j * 4 + 2] = orientation.y;
             s_instance->output_map[orient_key][j * 4 + 3] = orientation.z;
         }
+
+        // OpenXR Injection
+        if (s_instance->m_injector)
```
aristarkhovNV: what if it is null? And is it expected to change at run time

src/plugins/manus/ManusHandTrackingPlugin.cpp
aristarkhovNV: singletons are considered harmful, can we get rid of it?

aristarkhovNV: This seems wrong, is initialization really supposed to be multithreaded?

src/plugins/manus/session.cpp
```diff
@@ -0,0 +1,175 @@
...
+}
+
+bool Session::initialize(const SessionConfig& config)
+{
+    config_ = config;
+
+    if (!create_instance(config) || !get_system() || !create_session() ||
+        !create_reference_space(config.reference_space_type))
+    {
+        cleanup();
```
aristarkhovNV, line 34: failed object will be deleted eventually, should not need an explicit cleanup here

src/plugins/manus/session.cpp
```diff
@@ -0,0 +1,175 @@
...
+void Session::cleanup()
+{
+    if (handles_.instance == XR_NULL_HANDLE)
+    {
+        return;
+    }
+
+    if (handles_.reference_space != XR_NULL_HANDLE)
+    {
+        xrDestroySpace(handles_.reference_space);
```
aristarkhovNV, line 139: should `handles_` own its objects and destroy them in ~SessionHandles? Otherwise what's the benefit of this encapsulation

src/plugins/manus/XR_NVX1_device_interface.h
```diff
@@ -0,0 +1,124 @@
+// Copyright 2025, NVIDIA CORPORATION.
```
aristarkhovNV: this file should probably be submoduled from an official location

src/plugins/manus/CMakeLists.txt
rwiltz: Thanks will do

src/plugins/manus/ManusHandTrackingPlugin.cpp
rwiltz: ACK

src/plugins/manus/core/ManusHandTrackingPlugin.cpp
rwiltz: Yes, I stole these from the synthetic hands example with the plan to clean them up and move them to a common directory in a follow up MR

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
rwiltz: ACK

src/plugins/manus/controllers.cpp
rwiltz: Yes, I will re-write these a bit and move them to a common location

src/plugins/manus/controllers.cpp
rwiltz: Prob not, I agree this class should use RAII

src/plugins/manus/app/main.cpp
rwiltz: ACK

src/plugins/manus/app/main.cpp
rwiltz: ACK

src/plugins/manus/session.cpp
rwiltz: Yes, I will re-write these to be more "modern" C++ with RAII, etc

src/plugins/manus/ManusHandTrackingPlugin.cpp
rwiltz: The non RAII pattern of this class comes from the fact that it was copied from a holoscan operator, that can be improved now. However the singleton is unfortunately a reality of the ManusSDK. It has internal state that is not returned to the user in the form of a context or anything so it's essentially per process SDK state.

rwiltz: Since this was ported from a holoscan operator, it's to catch the situation where a single app has multiple Manus operators

src/plugins/manus/XR_NVX1_device_interface.h
```diff
@@ -0,0 +1,124 @@
...
+// SPDX-License-Identifier: Apache-2.0 or MIT
```
jiwenc-nv: is it intentional to have dual license here? @nvddr ?

src/plugins/manus/XR_NVX1_device_interface.h
jiwenc-nv: Assuming this is 100% the same file from the synthetic hands example, then shall we consider this header part of the core?

src/plugins/manus/ManusHandTrackingPlugin.cpp
jiwenc-nv: I copied these two files from @rwiltz's standalone Manus repo a while ago. But we should rename the header to comply with the other naming conventions.

aristarkhovNV: If a singleton is unavoidable lets at least convert it to Meyers Singleton:
```
Singleton& instance() { static Singleton s; return s; }
```
In other words keep the class itself agnostic to how it's being instantiated.

src/plugins/controller_synthetic_hands/CMakeLists.txt
```diff
@@ -31,7 +28,7 @@ find_package(Threads REQUIRED)
 # Link against OpenXR (fetched via CMake FetchContent - statically linked)
 # When statically linking, we need to include system dependencies
 target_link_libraries(controller_synthetic_hands PRIVATE
-    OpenXR::openxr_loader
+    Teleop::plugin_common # Transitive dependency on OpenXR::openxr_loader and Teleop::openxr_extensions
```
aristarkhovNV: note: comments like this will get stale with no chance of detecting. Also transitive deps are the implementation detail that shouldn't be relied on - if `controller_synthetic_hands` _directly_ uses `OpenXR::openxr_loader` or `Teleop::openxr_extensions` then link with all nececcary deps explicitly.

src/plugins/common/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Static library for common plugin code
+add_library(teleop_plugin_common STATIC
+    controllers.cpp
+    hand_injector.cpp
+    session.cpp
```
aristarkhovNV: list headers too so they can appear in IDE

src/plugins/plugin_utils/inc/plugin_utils/controllers.hpp
```diff
@@ -18,7 +18,6 @@ struct ControllerPose
 class Controllers
```
aristarkhovNV, line 24: Put externally consumable headers into a subdirectory so they would be consumed like `#include <common/controllers.hpp>`.
Also "common" as a library name hardly tells what this lib is responsible for. Maybe name it "plugin_common" or something.

src/plugins/common/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+target_link_libraries(teleop_plugin_common
+    PUBLIC
+        Teleop::openxr_extensions
+        OpenXR::openxr_loader
+)
+
+# Include directories
+target_include_directories(teleop_plugin_common
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
```
aristarkhovNV: do not expose the source dir - it give access to your private ncludes and also cpp. Put all consumable headers under 'inc/common/...', keep only private headers in CMAKE_CURRENT_SOURCE_DIR. Expose only 'inc' as INTERFACE

src/plugins/controller_synthetic_hands/CMakeLists.txt
rwiltz: Good call, I did a lot of vibe coding here given the short timeline, I'll fix it.

src/plugins/plugin_utils/inc/plugin_utils/controllers.hpp
rwiltz: I'll do plugin_common or plugin_utils

CMakeLists.txt
```diff
@@ -70,11 +69,8 @@ endif()
 # Build dependencies (OpenXR SDK, etc.)
 add_subdirectory(deps)
 
-# Enable CTest at top level so tests from subdirectories are discoverable
-if(BUILD_TESTING)
-    # Make sure to call this after `deps` is added so that Catch2 is available
-    enable_testing()
-endif()
+# Build interfaces (headers)
+add_subdirectory(include)
```
jiwenc-nv: I'm not sure where we want the new OpenXR headers to be. But this location seems to be a bit different from https://docs.google.com/document/d/1zqp4TOaQYqzC2EAsxQc0UTsM4skwHmMF5zCxDWVf9xU/edit?tab=t.0#heading=h.obp3sb305kk7, which is the folder layout we are following here.

@rwiltz @nvddr @aristarkhovNV @nv-jakob

src/plugins/manus/core/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+add_library(isaac_manus_plugin SHARED
+  manus_hand_tracking_plugin.cpp
+  manus_hand_tracking_plugin.hpp
+)
+
+target_include_directories(isaac_manus_plugin
+  PRIVATE
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
```
aristarkhovNV: don't expose whole source dir, put consumable headers in 'inc/lib_name' subdirectory. Also same dir is listed twice -- needs to be listed just once as INTERFACE for external consumption. Internally consume headers by relative paths.

deps/cloudxr/openxr_extensions/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# Interface library for OpenXR extensions
+add_library(teleop_openxr_extensions INTERFACE)
+add_library(Teleop::openxr_extensions ALIAS teleop_openxr_extensions)
+
+# Associate the include directory with the target
+target_include_directories(teleop_openxr_extensions INTERFACE
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+    $<INSTALL_INTERFACE:include>
```
aristarkhovNV: do we actually use the install interface?

src/plugins/controller_synthetic_hands/README.md
```diff
@@ -114,7 +114,7 @@ while (running) {
 #### Session Initialization
 
 ```cpp
-#include "session.hpp"
+#include "plugin_utils/session.hpp"
```
aristarkhovNV:
```suggestion
#include <plugin_utils/session.hpp>
```

src/plugins/controller_synthetic_hands/CMakeLists.txt
```diff
@@ -3,45 +3,38 @@
...
-    controllers.cpp
     hand_generator.cpp
-    hand_injector.cpp
+    synthetic_hands_plugin.hpp
+    hand_generator.hpp
 )
 
-# Include directories
 target_include_directories(controller_synthetic_hands PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}
```
aristarkhovNV: do not add current source dir, local headers will be found by relative paths

src/plugins/manus/core/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+    $<BUILD_INTERFACE:${MANUS_SDK_INCLUDE_DIR}>
```
aristarkhovNV: this should be picked automatically by linking with ManusSDK::ManusSDK

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -0,0 +1,504 @@
...
+#include <plugin_utils/hand_injector.hpp>
+#include <plugin_utils/math.hpp>
+#include <plugin_utils/session.hpp>
+
+#include <ManusSDK.h>
+#include <ManusSDKTypeInitializers.h>
+#include <algorithm>
+#include <chrono>
+#include <iostream>
+#include <manus_hand_tracking_plugin.hpp>
```
aristarkhovNV: own headers should be included by a relative path which would look something like this
```suggestion
#include "inc/core/manus_hand_tracking_plugin.hpp"
```

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -0,0 +1,504 @@
...
+    static ManusTracker s(app_name);
+    return s;
+}
+
+void ManusTracker::update()
+{
+    if (!m_controllers)
+    {
+        return;
+    }
```
aristarkhovNV: is this a normal situation, should this be an error?

src/plugins/manus/tools/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+add_executable(ManusHandTrackerPrinter
+  manus_hand_tracker_printer.cpp
+)
+
+target_include_directories(ManusHandTrackerPrinter
+  PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}
```
aristarkhovNV: Do not list current source dir, include by relative path (with "") instead

src/plugins/manus/tools/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    ${MANUS_SDK_INCLUDE_DIR}
```
aristarkhovNV: Manus SDK includes should be picked up by linking with the appropriate cmake target, don't explicitly add 3rd party include dirs

src/plugins/manus/tools/manus_hand_tracker_printer.cpp
```diff
@@ -0,0 +1,99 @@
...
+    g_should_exit = 1;
+}
+
+int main(int argc, char** argv)
+{
+    (void)argc;
+    (void)argv;
+
+    std::signal(SIGINT, handle_signal);
+    std::signal(SIGTERM, handle_signal);
```
aristarkhovNV: Do we really need custom? Might just rely on default signal handling

src/plugins/manus/tools/manus_hand_tracker_printer.cpp
```diff
@@ -0,0 +1,99 @@
...
+    plugins::manus::ManusTracker* tracker_ptr = nullptr;
+    try
+    {
+        tracker_ptr = &plugins::manus::ManusTracker::instance("ManusHandPrinter");
+    }
+    catch (const std::exception& e)
+    {
+        std::cerr << "Failed to initialize Manus tracker: " << e.what() << std::endl;
+        return 1;
+    }
```
aristarkhovNV: Wrap the entire main() so we could capture all errors, not just from the tracker init:
```
void main() try
{
    // all stuff
    // ...

    return 0;
}
catch (const std::exception& e)
{
    print(e.what());
    return 1;
}
```

src/plugins/plugin_utils/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+# Dependencies
+target_link_libraries(teleop_plugin_utils
+    PUBLIC
+    Teleop::openxr_extensions
+    OpenXR::openxr_loader
+)
+
+# Include directories
+target_include_directories(teleop_plugin_utils
+    PUBLIC
```
aristarkhovNV, line 24:
```suggestion
    INTERFACE
```

deps/cloudxr/openxr_extensions/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+)
+
+# Install the headers
+install(DIRECTORY openxr
+    DESTINATION include
```
jiwenc-nv, line 12: Let's put it in `deps/cloudxr` as placeholder location for now. then we can move it to cloudxr SDK then it's public.

aristarkhovNV: +1 to `deps/openxr`, also we probably don't rely on the `install` at this point?

PR #16, opened 2025-12-17
Adding unit tests for plugins

rwiltz:


src/plugins/manus/tests/mock_headers/openxr/openxr.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
```
aristarkhovNV, line 1: why not use actual headers instead?

PR #17, opened 2025-12-18
Add schema for head & hand pose

jiwenc-nv:


src/core/schema/fbs/hand.fbs
```diff
@@ -0,0 +1,36 @@
+include "pose.fbs";
+
+namespace core;
+
+// Describe a HandJoint pose.
+struct HandJointPose {
+    // The concrete pose data
+    pose: Pose;
+
+    // Whether the head pose data is valid
```
aristarkhovNV:
```suggestion
    // Whether the hand joint pose data is valid
```

src/core/schema/fbs/hand.fbs
```diff
@@ -0,0 +1,36 @@
...
+    poses: [HandJointPose:26];
+}
+
+// Describe a Head pose.
+table HandPose {
+    // Vector of HandJointPose.
+    // For OpenXR, this should be 26 according to XR_HAND_JOINT_COUNT_EXT
+    joints: HandJoints (id: 0);
+
+    // Whether the head pose data is valid
```
aristarkhovNV: hand

src/core/schema/fbs/pose.fbs
```diff
@@ -1,14 +1,13 @@
...
-table Pose {
+struct Pose {
   // Position of the pose (x, y, z) in meters.
-  position: Tensor (id: 0);
+  position: Point (id: 0);
 
   // Orientation of the pose (w, x, y, z).
-  orientation: Tensor (id: 1);
+  orientation: Quaternion (id: 1);
 }
```
aristarkhovNV, line 13: I wonder if we could somehow import OpenXR types such as XrPose directly

src/core/schema/python/hand_bindings.h
```diff
@@ -0,0 +1,105 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aristarkhovNV, line 1: we need to settle on h vs. hpp header file extension, other places seem to prefer hpp?

src/core/schema/python/hand_bindings.h
```diff
@@ -0,0 +1,105 @@
...
+
+#pragma once
+
+#include <memory>
+#include <array>
+#include <vector>
+#include <cstring>
+
+#include "pybind11/pybind11.h"
+#include "pybind11/stl.h"
```
aristarkhovNV:
```suggestion
#include <pybind11/pybind11.h>
#include <pybind11/stl.h>
```

src/core/schema/python/head_bindings.h
```diff
@@ -0,0 +1,53 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aristarkhovNV, line 1: add new headers to cmakelists for easier access/search from IDE

src/core/schema/fbs/hand.fbs
jiwenc-nv: fixed.

src/core/schema/python/hand_bindings.h
jiwenc-nv: Good question. I do agree we should pick one and consolidate to one convention. I don't have strong opinion which one.

jiwenc-nv: We can do a quick vote tomorrow (effectively it's a coin toss)

src/core/schema/python/hand_bindings.h
jiwenc-nv: Fixed and good catch. Would be nice to have pre-submit to catch this kind of error automatcailly.

src/core/schema/fbs/pose.fbs
jiwenc-nv: actually, isn't the `struct Pose` here a one-to-one mapping to XrPose?

https://github.com/KhronosGroup/OpenXR-SDK/blob/main/include/openxr/openxr.h#L1181

Even the quaternion order is the same: xyzw.

jiwenc-nv: Or @aristarkhovNV you are suggest something else? like somehow make `Pose` struct generated from fbs and `XrPose` from OpenXR interchangeable at C++ level?

src/core/schema/python/head_bindings.h
jiwenc-nv: do you mean list the header file explicitly in the CMakeLists.txt?

src/core/schema/python/hand_bindings.h
jiwenc-nv: let's make another PR to change this.

src/core/schema/python/head_bindings.h
aristarkhovNV: yes, along with cpp

src/core/schema/fbs/pose.fbs
aristarkhovNV: yes ideally have fbs somehow produce an alias to the existing XR type that would be directly interchangeable at the c++ level

PR #31, opened 2025-12-23
feat: add mcap_recorder and integrate it to deviceio_session.

shaosu-nvidia:


deps/third_party/CMakeLists.txt
```diff
@@ -111,3 +111,31 @@ if(BUILD_TESTING)
...
+target_include_directories(mcap_interface INTERFACE
+    ${mcap_SOURCE_DIR}/cpp/mcap/include
+)
+# Disable compression to avoid Zstd/LZ4 dependencies
+target_compile_definitions(mcap_interface INTERFACE
+    MCAP_COMPRESSION_NO_ZSTD
+    MCAP_COMPRESSION_NO_LZ4
+)
+add_library(mcap::mcap ALIAS mcap_interface)
+message(STATUS "MCAP interface library configured")
```
aristarkhovNV, line 141: `mcap_interface` seems unused

deps/third_party/CMakeLists.txt
```diff
@@ -111,3 +111,31 @@ if(BUILD_TESTING)
...
+message(STATUS "Fetching MCAP from GitHub...")
+FetchContent_Declare(
+    mcap
+    GIT_REPOSITORY https://github.com/foxglove/mcap.git
+    GIT_TAG        releases/cpp/v1.4.1
+    GIT_SHALLOW    TRUE
+)
+
+FetchContent_MakeAvailable(mcap)
+message(STATUS "MCAP fetched")
```
aristarkhovNV, line 128: no need for extra logging, FetchContent_* already does necessary status logging

src/core/deviceio/cpp/inc/deviceio/deviceio_session.hpp
```diff
@@ -59,6 +62,13 @@ class DeviceIOSession
 #elif defined(XR_USE_TIMESPEC)
     PFN_xrConvertTimespecTimeToTimeKHR pfn_convert_timespec_{};
 #endif
+
+    // MCAP recording (controlled via MCAP_RECORDING env var)
```
aristarkhovNV: Let library users decide how they want to control recording, do not empose the env var at the library level

src/core/deviceio/cpp/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ add_library(deviceio_core STATIC
 )
 
 target_include_directories(deviceio_core
-    INTERFACE
+    PUBLIC
```
aristarkhovNV: keep the `inc` under INTERFACE to enforce own includes use "" and relative paths

src/core/deviceio/cpp/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ add_library(deviceio_core STATIC
...
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
         $<INSTALL_INTERFACE:include>
```
aristarkhovNV: install interface is unused and can be removed

src/core/mcap/cpp/inc/mcap_recorder.hpp
```diff
@@ -0,0 +1,93 @@
...
+    McapRecorder(const McapRecorder&) = delete;
+    McapRecorder& operator=(const McapRecorder&) = delete;
+
+    /**
+     * @brief Open an MCAP file for writing.
+     *
+     * @param filename Path to the output MCAP file.
+     * @return true if file was opened successfully, false otherwise.
+     */
+    bool open(const std::string& filename);
```
aristarkhovNV: lets use RAII instead - pass the file name and trackers to constructor so there is less mutable state to maintain throughout the life time of the object

src/core/deviceio/cpp/inc/deviceio/handtracker.hpp
```diff
@@ -17,15 +19,26 @@ namespace core
...
     // Public API - what external users see
     std::vector<std::string> get_required_extensions() const override;
     std::string get_name() const override
     {
-        return "HandTracker";
+        return TRACKER_NAME;
+    }
+    std::string get_schema_name() const override
+    {
+        return "core.HandPose";
```
jiwenc-nv: nit: `SCHEMA_NAME`? similar to TRACKER_NAME, so that we don't need to hard code `core.HandPose` again.

src/core/mcap/cpp/inc/mcap_recorder.hpp
```diff
@@ -0,0 +1,93 @@
...
+     */
+    void add_tracker(const std::shared_ptr<ITrackerImpl> tracker_impl);
+
+    /**
+     * @brief Get the channel ID assigned to a tracker impl.
+     *
+     * @param tracker_impl Pointer to the tracker implementation instance.
+     * @return The channel ID, or 0 if tracker not found.
+     */
+    mcap::ChannelId get_channel_id(const ITrackerImpl* tracker_impl) const;
```
aristarkhovNV: can this be made private?

src/core/mcap/cpp/inc/mcap/recorder.hpp
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * @brief Get the path to the currently open file.
+     *
+     * @return The filename, or empty string if no file is open.
+     */
+    std::string get_filename() const;
+
+private:
+    class Impl;
+    std::unique_ptr<Impl> impl_;
```
aristarkhovNV, line 73: Are there any benefits of using pimpl here? I know it's used in other places but we don't have to follow this pattern. Consider using a plain implementation instead.

src/core/mcap/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+    mcap_recorder.cpp
+    inc/mcap_recorder.hpp
+)
+
+target_include_directories(mcap_core
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
+        $<INSTALL_INTERFACE:include>
+        # Include deviceio headers for ITrackerImpl interface (header-only dependency)
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../deviceio/cpp/inc>
```
aristarkhovNV: remove, this should be picked up via `target_link_libraries`

src/core/mcap/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+# MCAP Recorder library (static for distribution)
+add_library(mcap_core STATIC
+    mcap_recorder.cpp
+    inc/mcap_recorder.hpp
+)
+
+target_include_directories(mcap_core
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
+        $<INSTALL_INTERFACE:include>
```
aristarkhovNV: install interface is unused

src/core/mcap/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+cmake_minimum_required(VERSION 3.20)
+
+# MCAP Recorder library (static for distribution)
+add_library(mcap_core STATIC
+    mcap_recorder.cpp
+    inc/mcap_recorder.hpp
+)
+
+target_include_directories(mcap_core
+    PUBLIC
```
aristarkhovNV: `inc` should be exposed as interface so there is a clear distinction between consuming own headers vs others
```suggestion
    INTERFACE
```

src/core/mcap_tests/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aristarkhovNV, line 1: move tests to src/core/mcap_tests so all targets sit at the same hierarchy level

deps/third_party/CMakeLists.txt
```diff
@@ -111,3 +111,31 @@ if(BUILD_TESTING)
...
+# ==============================================================================
+# MCAP (Recording Format Library)
+# ==============================================================================
+# MCAP is a modular container format for heterogeneous timestamped data.
+# It is a header-only library.
+message(STATUS "Fetching MCAP from GitHub...")
+FetchContent_Declare(
+    mcap
+    GIT_REPOSITORY https://github.com/foxglove/mcap.git
+    GIT_TAG        releases/cpp/v1.4.1
```
jiwenc-nv: any reason to use `v1.4.1`? 

seems there is `v2.1.1` already?
https://github.com/foxglove/mcap/releases/tag/releases%2Fcpp%2Fv2.1.1

jiwenc-nv: if I read it correctly: there is a year of diff?
https://github.com/foxglove/mcap/compare/releases/cpp/v1.4.1...releases/cpp/v2.1.1

deps/third_party/CMakeLists.txt
shaosu-nvidia: I checked the log and I don't actually see FetchContent_* print any status logging.

src/core/mcap/cpp/CMakeLists.txt
shaosu-nvidia: I have separated it to  2 target_include_directories and the INTERFACE is used for /inc

src/core/mcap/cpp/CMakeLists.txt
shaosu-nvidia: I think it's resolved but let me know if any further comment

src/core/mcap/cpp/inc/mcap/recorder.hpp
shaosu-nvidia: Thanks for pointing that out. I realized I have somehow exposed mcap::ChannelId get_channel_id from McapRecorder but it should be only used internally as a helper function. So after removing it from public interface, it seems the PIMPL here has some benefit. It basically makes mcap::mcap header becomes PRIVATE to the IMPL

deps/third_party/CMakeLists.txt
shaosu-nvidia: I think this is used because mcap repo is not CMake so need to explicitly define the interface that will be used by McapRecorder later.

deps/third_party/CMakeLists.txt
aristarkhovNV: Yea it seems to be quiet unless you specify `-DFETCHCONTENT_QUIET=NO`, so one line status should be ok then.

deps/third_party/CMakeLists.txt
aristarkhovNV: oh I see it's being picked up via mcap::mcap alias

examples/oxr/python/modular_example_with_mcap.py
```diff
@@ -0,0 +1,109 @@
...
+    trackers = [hand_tracker, head_tracker]
+    required_extensions = deviceio.DeviceIOSession.get_required_extensions(trackers)
+    print(f"✓ Required extensions: {required_extensions}")
+    
+    # Create OpenXR session
+    print("\nCreating OpenXR session...")
+    oxr_session = oxr.OpenXRSession.create("ModularExampleWithMCAP", required_extensions)
+    if oxr_session is None:
+        print("✗ Failed to create OpenXR session")
+        return 1
```
aristarkhovNV: we try to move away from manual error handling to keep the example and test code cleaner. OpenXRSession.create should raise the exception which would terminate the program with appropriate error message

examples/oxr/python/modular_example_with_mcap.py
```diff
@@ -0,0 +1,109 @@
...
+    
+    # Use context managers for proper RAII cleanup
+    with oxr_session:
```
aristarkhovNV:
```suggestion
    with oxr.OpenXRSession.create("ModularExampleWithMCAP", required_extensions) as oxr_session:
```

examples/oxr/python/modular_example_with_mcap.py
```diff
@@ -0,0 +1,109 @@
...
+                    if frame_count % 60 == 0:
+                        elapsed = time.time() - start_time
+                        print(f"[{elapsed:4.1f}s] Frame {frame_count} (recording...)")
+                        print()
+                    
+                    frame_count += 1
+                    time.sleep(0.016)  # ~60 FPS
+            
+            except KeyboardInterrupt:
+                print("\nInterrupted by user")
```
aristarkhovNV: no need to handle manually, the KeyboardInterrupt will properly terminate the `with session:` block freeing the session resources as necessary. Also the interrupt may technically happen at any other point in the program

src/core/deviceio/cpp/inc/deviceio/deviceio_session.hpp
```diff
@@ -59,6 +64,12 @@ class DeviceIOSession
 #elif defined(XR_USE_TIMESPEC)
     PFN_xrConvertTimespecTimeToTimeKHR pfn_convert_timespec_{};
 #endif
+
+    // MCAP recording
+    std::unique_ptr<McapRecorder> recorder_;
+    bool start_recording(const std::string& filename);
+    void stop_recording();
+    bool is_recording() const;
```
aristarkhovNV: if recording starts upon construction and continues till the end then we may not need lifecycle control functions - they create the impression that recording may be started and stopped at any point

src/core/mcap/cpp/inc/mcap/recorder.hpp
```diff
@@ -0,0 +1,84 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aristarkhovNV, line 1: put this header under inc/mcap/ directory so when consumed we'd have to prepend the module name as `#include <mcap/mcap_recorder.hpp>`

may then also omit the `mcap_` prefix so it becomes just `<mcap/recorder.hpp>`

src/core/mcap/cpp/inc/mcap_recorder.hpp
```diff
@@ -0,0 +1,84 @@
...
+    McapRecorder();
+
+    /**
+     * @brief Destructor - closes the file if open.
+     */
+    ~McapRecorder();
+
+    // Non-copyable
+    McapRecorder(const McapRecorder&) = delete;
+    McapRecorder& operator=(const McapRecorder&) = delete;
```
aristarkhovNV: this is unnecessary, the unique_ptr member makes it non-copyable already

src/core/mcap/cpp/inc/mcap_recorder.hpp
```diff
@@ -0,0 +1,84 @@
...
+     * @return true if file is open, false otherwise.
+     */
+    bool is_open() const;
+
+    /**
+     * @brief Add a tracker implementation to the recorder.
+     *
+     * @param tracker_impl The tracker implementation to register.
+     */
+    void add_tracker(const std::shared_ptr<ITrackerImpl> tracker_impl);
```
aristarkhovNV: lets try to pass all trackers upfront to the ctor so the recorder stays immutable for its life time. This makes the testing and support easier - fiwer corner cases to consider.

src/core/mcap/cpp/mcap_recorder.cpp
```diff
@@ -0,0 +1,203 @@
...
+        close();
+    }
+
+    bool open(const std::string& filename)
+    {
+        if (is_open_)
+        {
+            std::cerr << "McapRecorder: File already open, close it first" << std::endl;
+            return false;
+        }
```
aristarkhovNV: Open just once in the constructor so we don't have to keep track of the is_open state. The file would be assumed open of the tracker exists. Failed to open - constructor throws and no unopened impl ever gets created.

src/core/mcap/cpp/inc/mcap_recorder.hpp
```diff
@@ -0,0 +1,84 @@
...
+     * @return true if recording succeeded, false otherwise.
+     */
+    bool record(const std::shared_ptr<ITrackerImpl> tracker_impl);
+
+    /**
+     * @brief Get the path to the currently open file.
+     *
+     * @return The filename, or empty string if no file is open.
+     */
+    std::string get_filename() const;
```
aristarkhovNV: should not be needed, consider removing for now - can always add later if it becomes really necessary

src/core/mcap/cpp/CMakeLists.txt
aristarkhovNV: Don't explicitly add `target_include_directories` to access headers from "deviceio/cpp/inc". Instead add `deviceio_core` to your `target_link_libraries` and it will bring its headers and other necessary stuff.

src/core/mcap/cpp/mcap_recorder.cpp
```diff
@@ -0,0 +1,203 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#define MCAP_IMPLEMENTATION
```
aristarkhovNV, line 4: define seems unused?

src/core/mcap/cpp/CMakeLists.txt
shaosu-nvidia: I think this should be resolved with the most recent change.

src/core/mcap/cpp/mcap_recorder.cpp
shaosu-nvidia: from https://github.com/foxglove/mcap/tree/releases/cpp/v2.1.1/cpp, we will need to define once in cpp file. and from cursor explanation:
The MCAP library is a header-only library. It uses this pattern:
Without #define MCAP_IMPLEMENTATION: Headers only provide declarations (interfaces)
With #define MCAP_IMPLEMENTATION: Headers also include the actual function implementations

src/core/deviceio/cpp/inc/deviceio/handtracker.hpp
```diff
@@ -21,7 +23,17 @@ class HandTracker : public ITracker
...
-        return "HandTracker";
+        return TRACKER_NAME;
+    }
+
+    std::string get_schema_name() const override
+    {
+        return SCHEMA_NAME;
+    }
+
+    std::string get_schema_text() const override
```
aristarkhovNV: return string_view to avoid heap allocation

src/core/deviceio/cpp/inc/deviceio/handtracker.hpp
```diff
@@ -21,7 +23,17 @@ class HandTracker : public ITracker
     std::vector<std::string> get_required_extensions() const override;
     std::string get_name() const override
     {
-        return "HandTracker";
+        return TRACKER_NAME;
+    }
+
+    std::string get_schema_name() const override
```
aristarkhovNV: return string_view to avoid heap allocation

src/core/deviceio/cpp/inc/deviceio/handtracker.hpp
```diff
@@ -21,7 +23,17 @@ class HandTracker : public ITracker
     std::vector<std::string> get_required_extensions() const override;
     std::string get_name() const override
```
aristarkhovNV: change to string_view to avoid heap allocation on every call

src/core/deviceio/cpp/controllertracker.cpp
```diff
@@ -354,6 +354,29 @@ const ControllerDataT& ControllerTracker::Impl::get_controller_data() const
...
+void ControllerTracker::Impl::serialize(flatbuffers::FlatBufferBuilder& builder, int64_t* out_timestamp) const
+{
+    if (out_timestamp)
+    {
+        // Use left controller timestamp (or right if left is inactive)
+        if (controller_data_.left_controller && controller_data_.left_controller->is_active())
+        {
+            *out_timestamp = controller_data_.left_controller->timestamp().device_time();
+        }
+        else if (controller_data_.right_controller)
```
aristarkhovNV: check is_active?

src/core/deviceio/cpp/handtracker.cpp
```diff
@@ -92,6 +94,24 @@ HandTracker::Impl::Impl(const OpenXRSessionHandles& handles)
     std::cout << "HandTracker initialized (left + right)" << std::endl;
 }
 
+void HandTracker::Impl::serialize(flatbuffers::FlatBufferBuilder& builder, int64_t* out_timestamp) const
```
aristarkhovNV: can we explicitly indicate that output timestamp is the device time?
```suggestion
void HandTracker::Impl::serialize(flatbuffers::FlatBufferBuilder& builder, int64_t* out_device_time) const
```

Or alternatively return the whole timestamp structure so the user can pick what they need

src/core/deviceio/python/deviceio_bindings.cpp
```diff
@@ -104,11 +56,19 @@ PYBIND11_MODULE(_deviceio, m)
             py::arg("session"), py::return_value_policy::reference_internal,
             "Get complete controller data for both left and right controllers");
 
+    // Register the native DeviceIOSession type so it can be shared across modules
+    // This opaque binding allows other modules (like mcap) to accept DeviceIOSession references
+    py::class_<core::DeviceIOSession>(m, "_DeviceIOSessionNative");
```
aristarkhovNV: We don't pass the `core::DeviceIOSession` object because it will be destroyed when PyDeviceIOSession is closed after exiting the `with:` block and you'll end up with a dead pointer. Instead pass the `PyDeviceIOSession` wrapper around and use `.native()` call to retrieve the underlying c++ session.

src/core/deviceio/python/deviceio_bindings.cpp
```diff
@@ -104,11 +56,19 @@ PYBIND11_MODULE(_deviceio, m)
...
+
     // DeviceIOSession class (bound via wrapper for context management)
     py::class_<PyDeviceIOSession>(m, "DeviceIOSession")
         .def("update", &PyDeviceIOSession::update, "Update session and all trackers")
         .def("__enter__", &PyDeviceIOSession::enter)
         .def("__exit__", &PyDeviceIOSession::exit)
+        .def(
+            "_native", [](PyDeviceIOSession& self) -> core::DeviceIOSession& { return self.native(); },
+            py::return_value_policy::reference_internal,
+            "Internal: Get native C++ session reference for cross-module use")
```
aristarkhovNV: don't do this, native session is not supposed to be visible to python

src/core/mcap/cpp/mcap_recorder.cpp
```diff
@@ -0,0 +1,211 @@
...
+    std::unordered_map<const ITracker*, mcap::ChannelId> tracker_channel_ids_;
+};
+
+// McapRecorder public interface implementation
+
+std::unique_ptr<McapRecorder> McapRecorder::start_recording(const std::string& filename,
+                                                            const std::vector<TrackerChannelPair>& trackers)
+{
+    auto recorder = std::unique_ptr<McapRecorder>(new McapRecorder(filename, trackers));
+    if (!recorder->impl_->start_recording())
```
aristarkhovNV: Lets start recording in the McapRecorder constructor, so you never forget to start. Also let ctor throw on failure so we never have to explicitly check for nullptr.

src/core/mcap/python/mcap_bindings.cpp
```diff
@@ -0,0 +1,95 @@
...
+public:
+    PyMcapRecorder(std::unique_ptr<core::McapRecorder> recorder) : recorder_(std::move(recorder))
+    {
+    }
+
+    void stop_recording()
+    {
+        if (recorder_)
+        {
+            recorder_->stop_recording();
```
aristarkhovNV:
```suggestion
            recorder_->stop_recording();
            recorder_.reset();
```

src/core/mcap/python/mcap_bindings.cpp
```diff
@@ -0,0 +1,95 @@
...
+    // McapRecorder class
+    py::class_<PyMcapRecorder>(m, "McapRecorder")
+        .def_static(
+            "start_recording",
+            [](const std::string& filename, const std::vector<core::McapRecorder::TrackerChannelPair>& trackers)
+            {
+                auto recorder = core::McapRecorder::start_recording(filename, trackers);
+                if (!recorder)
+                {
+                    throw std::runtime_error("Failed to start recording to " + filename);
```
aristarkhovNV: let the `core::McapRecorder::start_recording` throw instead

src/core/mcap/cpp/inc/mcap/recorder.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * @brief Destructor - closes the file if open.
+     */
+    ~McapRecorder();
+
+    /**
+     * @brief Stop recording and close the MCAP file.
+     */
+    void stop_recording();
```
aristarkhovNV: native `McapRecorder` does not need `stop_recording`, it can use class destructor for that.

src/core/mcap/cpp/mcap_recorder.cpp
```diff
@@ -0,0 +1,211 @@
...
+        stop_recording();
+    }
+
+    bool start_recording()
+    {
+        if (is_recording_)
+        {
+            std::cerr << "McapRecorder: Recording already in progress" << std::endl;
+            return false;
+        }
```
aristarkhovNV: this check and the whole `is_recording_ ` state would not be needed if we implicitly start recording in ctor and stop in dtor

src/core/mcap/cpp/mcap_recorder.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+    return recorder;
+}
+
+McapRecorder::McapRecorder(const std::string& filename, const std::vector<TrackerChannelPair>& trackers)
+    : impl_(std::make_unique<Impl>(filename, trackers))
+{
+}
+
+McapRecorder::~McapRecorder() = default;
```
aristarkhovNV, line 160: destructor should stop recording

src/core/deviceio/cpp/CMakeLists.txt
aristarkhovNV: in other words - revert this line

src/core/deviceio/python/inc/py_deviceio/session.hpp
```diff
@@ -0,0 +1,61 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aristarkhovNV, line 1: create one more directory level such as inc/py_deviceio/ so this file would be consumed as `#include <py_deviceio/session.hpp>`

src/core/mcap/python/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    PRIVATE
+        mcap::mcap_core
+        deviceio::deviceio_core
+)
+
+# Include deviceio python headers for PyDeviceIOSession wrapper
+target_include_directories(mcap_py
+    PRIVATE
+        ${CMAKE_SOURCE_DIR}/src/core/deviceio/python/inc
+)
```
aristarkhovNV, line 32: Do `target_link_libraries(mcap_py PRIVATE deviceio_py)` instead, don't explicitly add other modules include dirs. Also `deviceio_py` target will need to expose its own inc directory

src/core/deviceio/python/CMakeLists.txt
```diff
@@ -24,9 +24,14 @@ target_link_libraries(deviceio_py
         deviceio_core
 )
 
+# Include directory for py_deviceio headers (session.hpp wrapper)
+target_include_directories(deviceio_py
+    PRIVATE
```
aristarkhovNV, line 29: This needs to be visible to consuming modules
```suggestion
    INTERFACE
```

aristarkhovNV: actually never mind, I just realized it will not work with pybind targets

src/core/mcap/python/CMakeLists.txt
aristarkhovNV: Leave as is for now, this needs a nore involved change

PR #32, opened 2025-12-23
Add timestamps for device, use flatbuffer for controller schema.

nvddr:
…ffer schema

### DIFF
--- a/examples/oxr/python/test_controller_tracker.py
+++ b/examples/oxr/python/test_controller_tracker.py
@@ -6,7 +6,7 @@
 Test script for ControllerTracker with simplified API.
 
 Demonstrates:
-- Snapshot mode for continuous state queries (includes poses and inputs)
+- Getting complete controller data for both left and right controllers
 """
 
 import sys
@@ -59,20 +59,21 @@ with oxr.OpenXRSession.create("ControllerTrackerTest", required_extensions) as o
         print("✓ Update successful")
         print()
         
-        # Test 5: Check initial controller state (snapshot mode)
-        print("[Test 5] Checking controller state (snapshot mode)...")
-        left_snap = controller_tracker.get_snapshot(deviceio.Hand.Left)
-        right_snap = controller_tracker.get_snapshot(deviceio.Hand.Right)
-        print(f"  Left controller: {'ACTIVE' if left_snap.is_active else 'INACTIVE'}")
-        print(f"  Right controller: {'ACTIVE' if right_snap.is_active else 'INACTIVE'}")
+        # Test 5: Check initial controller state
+        print("[Test 5] Checking controller state...")
+        controller_data = controller_tracker.get_controller_data()
+        left_snap = controller_data.left_controller
+        right_snap = controller_data.right_controller
+        print(f"  Left controller: {'ACTIVE' if left_snap and left_snap.is_active else 'INACTIVE'}")
+        print(f"  Right controller: {'ACTIVE' if right_snap and right_snap.is_active else 'INACTIVE'}")
         
-        if left_snap.is_active and left_snap.grip_pose.is_valid:
-            pos = left_snap.grip_pose.position
-            print(f"  Left grip position: [{pos[0]:.3f}, {pos[1]:.3f}, {pos[2]:.3f}]")
+        if left_snap and left_snap.is_active and left_snap.grip_pose.is_valid:
+            pos = left_snap.grip_pose.pose.position
+            print(f"  Left grip position: [{pos.x:.3f}, {pos.y:.3f}, {pos.z:.3f}]")
         
-        if right_snap.is_active and right_snap.grip_pose.is_valid:
-            pos = right_snap.grip_pose.position
-            print(f"  Right grip position: [{pos[0]:.3f}, {pos[1]:.3f}, {pos[2]:.3f}]")
+        if right_snap and right_snap.is_active and right_snap.grip_pose.is_valid:
+            pos = right_snap.grip_pose.pose.position
+            print(f"  Right grip position: [{pos.x:.3f}, {pos.y:.3f}, {pos.z:.3f}]")
         print()
         
         # Test 6: Available inputs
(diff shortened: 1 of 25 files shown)


examples/oxr/python/test_controller_tracker.py
```diff
@@ -66,13 +66,13 @@
         print(f"  Left controller: {'ACTIVE' if left_snap.is_active else 'INACTIVE'}")
         print(f"  Right controller: {'ACTIVE' if right_snap.is_active else 'INACTIVE'}")
         
-        if left_snap.is_active and left_snap.grip_pose.is_valid:
-            pos = left_snap.grip_pose.position
-            print(f"  Left grip position: [{pos[0]:.3f}, {pos[1]:.3f}, {pos[2]:.3f}]")
+        if left_snap.is_active and left_snap.grip_pose and left_snap.grip_pose.is_valid:
```
aristarkhovNV: having to both check for Null and .is_valid seems excessive. Can we ensure only one check is needed? Like to guarantee being valid if present so user only has to check for Null. Or to guarantee being present, so user would only check .is_valid

examples/oxr/python/test_controller_tracker.py
```diff
@@ -103,20 +103,20 @@
...
                     right_snap = controller_tracker.get_snapshot(deviceio.Hand.Right)
                     
-                    # Show current state using field access
-                    left_trigger = left_snap.inputs.trigger_value
-                    left_squeeze = left_snap.inputs.squeeze_value
-                    left_stick_x = left_snap.inputs.thumbstick_x
-                    left_stick_y = left_snap.inputs.thumbstick_y
-                    left_primary = left_snap.inputs.primary_click
-                    left_secondary = left_snap.inputs.secondary_click
+                    # Show current state using field access (check if inputs exist)
```
aristarkhovNV: clarify in what case it may not exist

examples/oxr/python/test_controller_tracker.py
```diff
@@ -161,21 +161,22 @@ def print_controller_summary(hand_name, hand):
                 grip = snapshot.grip_pose
                 aim = snapshot.aim_pose
                 
-                if grip.is_valid:
-                    pos = grip.position
-                    print(f"    Grip position: [{pos[0]:+.3f}, {pos[1]:+.3f}, {pos[2]:+.3f}]")
+                if grip and grip.is_valid and grip.pose:
```
aristarkhovNV: if grip is valid then the pose thould be filled in, right?

src/core/deviceio/cpp/controllertracker.cpp
```diff
@@ -278,60 +278,86 @@ bool ControllerTracker::Impl::update(XrTime time)
...
-                snapshot.grip_pose.orientation[0] = grip_location.pose.orientation.x;
-                snapshot.grip_pose.orientation[1] = grip_location.pose.orientation.y;
-                snapshot.grip_pose.orientation[2] = grip_location.pose.orientation.z;
-                snapshot.grip_pose.orientation[3] = grip_location.pose.orientation.w;
+                // Create pose struct with position and orientation
+                Point position(
+                    grip_location.pose.position.x, grip_location.pose.position.y, grip_location.pose.position.z);
+                Quaternion orientation(grip_location.pose.orientation.x, grip_location.pose.orientation.y,
+                                       grip_location.pose.orientation.z, grip_location.pose.orientation.w);
+                snapshot.grip_pose->pose = std::make_unique<Pose>(position, orientation);
```
aristarkhovNV: There should be a way to for schema types to use native types directly so no conversion is necessary. See if we can apply this https://docs.google.com/document/d/1JTBeJMJ8L75Am_hMCeiHHFohKgzZtL3l3-T4yYSR724/edit?disco=AAABxqzGdR4

src/core/deviceio/cpp/controllertracker.cpp
```diff
@@ -278,60 +278,86 @@ bool ControllerTracker::Impl::update(XrTime time)
...
-        snapshot.inputs.primary_click = get_boolean_action_state(session_, core_funcs_, primary_click_action_, hand_path);
-        snapshot.inputs.secondary_click =
+        snapshot.inputs->primary_click =
+            get_boolean_action_state(session_, core_funcs_, primary_click_action_, hand_path);
+        snapshot.inputs->secondary_click =
             get_boolean_action_state(session_, core_funcs_, secondary_click_action_, hand_path);
 
-        get_vector2_action_state(session_, core_funcs_, thumbstick_action_, hand_path, snapshot.inputs.thumbstick_x,
-                                 snapshot.inputs.thumbstick_y);
+        float thumbstick_x = 0.0f, thumbstick_y = 0.0f;
```
aristarkhovNV, line 329: can the temp vars be skipped by retrieving the action state directly into  snapshot.inputs->thumbstick_*

src/core/schema/fbs/controller.fbs
```diff
@@ -0,0 +1,66 @@
...
+    grip_pose: ControllerPose (id: 0);
+
+    // Aim pose - represents where the controller is pointing
+    aim_pose: ControllerPose (id: 1);
+
+    // Input state - all buttons and axes
+    inputs: ControllerInputState (id: 2);
+
+    // Whether the controller is active (connected and tracked)
+    is_active: bool (id: 3);
```
aristarkhovNV: would this duplicate the ControllerPose.is_valid

src/core/schema/python/hand_bindings.h
```diff
@@ -99,6 +101,15 @@ inline void bind_hand(py::module& m)
                 }
             },
             py::arg("joint_poses"), "Set the joints from a list of 26 HandJointPose entries.")
+        .def(
+            "set_timestamp",
+            [](HandPoseT& self, int64_t device_time, int64_t common_time)
```
aristarkhovNV: these seems to be an inconsistency between the getter and the setter. Get returns the whole Timestamp object, Set takes two separate components

examples/oxr/python/test_controller_tracker.py
nvddr: There are scenarios in which we do get a fake pose that may not be valid.. or a scenario in which we only have aim pose and no grip pose? If the schema allows it we should be checking for it.

nvddr: Updated to make the snapshot struct

src/core/deviceio/cpp/controllertracker.cpp
nvddr: Changing to struct means we need temp vars because the snapshot needs to be constructed in one shot.

src/core/schema/fbs/controller.fbs
nvddr: Controller active is different from individual pose validity

src/core/schema/python/hand_bindings.h
nvddr: I think this is okay? The setter becomes a simple function call.

src/core/deviceio/cpp/controllertracker.cpp
nvddr: Deferred

examples/oxr/python/test_controller_tracker.py
nvddr: Fixed

examples/oxr/python/test_controller_tracker.py
nvddr: Fixed

src/core/schema/python/hand_bindings.h
aristarkhovNV: So we have a read-only property and a setter at the same time. Read-only implies immutability, but if we allow write access then might as well do it uniformly via the prop so the api contract is more clean. Like we can assign is_active directly but not the timestamp for some reason? Although I've just noticed we handle joints the same way. I'll leave it up to you, functionality-wise seems fine either way.

nvddr: Actually the python bindings should not have setters at all..?

nvddr: Removed setters from python

PR #33, opened 2026-01-05
Remove is_initialized() method

aristarkhovNV:


src/core/deviceio/cpp/inc/deviceio/tracker.hpp
```diff
@@ -39,7 +39,6 @@ class ITracker
     // Public API - visible to all users
     virtual std::vector<std::string> get_required_extensions() const = 0;
     virtual std::string get_name() const = 0;
-    virtual bool is_initialized() const = 0;
```
jiwenc-nv, line 42: @aristarkhovNV @nvddr maybe you folks already chatted offline. But would be helpful to document in the PR message on why we think this is no longer needed.

aristarkhovNV: discussed in slack - is_initialized was only used for logging and sanity checks.

PR #37, opened 2026-01-07
Initial version of mano to G1 retargeting

lduan-intern:
The robot hand world trajectory is not right yet.


jiwenc-nv:
closing off old PR.

PR #38, opened 2026-01-07
Add Tensor and TensorGroup classes with comprehensive tests

nvddr:


src/core/retargeting_engine_tests/python/test_tensor.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def test_tensor_repr_with_value(self):
+        """Test repr with set value."""
+        int_type = IntType("counter")
+        tensor = Tensor(int_type, 42)
+        
+        repr_str = repr(tensor)
+        assert "Tensor" in repr_str
+        assert "counter" in repr_str
+        assert "42" in repr_str
```
aristarkhovNV: may do a single assert with exact expected string representation

src/core/retargeting_engine_tests/python/test_tensor_group.py
```diff
@@ -0,0 +1,441 @@
...
+            tensors=[FloatType("x")]
+        )
+        tensor_group = TensorGroup(group_type)
+        tensor_group[0] = 1.0
+        
+        with pytest.raises(IndexError):
+            _ = tensor_group[1]
+        
+        with pytest.raises(IndexError):
+            _ = tensor_group[-2]
```
aristarkhovNV, line 155: is access at -1 valid and does it do the normal python thing? may add such test case

PR #39, opened 2026-01-09
Add retargeter interface

nvddr:
Adds the retargeter interface with tests

### DIFF
--- a/src/core/retargeting_engine_tests/python/CMakeLists.txt
+++ b/src/core/retargeting_engine_tests/python/CMakeLists.txt
@@ -32,3 +32,12 @@ foreach(test_file ${TEST_FILES})
     )
 endforeach()
 
+# Add mypy type checking test
+add_test(
+    NAME "retargeting_mypy_type_check"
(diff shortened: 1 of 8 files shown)


src/core/retargeting_engine_tests/python/CMakeLists.txt
```diff
@@ -32,3 +32,12 @@ foreach(test_file ${TEST_FILES})
...
 
+# Add mypy type checking test
+add_test(
+    NAME "retargeting_mypy_type_check"
+    COMMAND uv run --python ${TELEOPCORE_PYTHON_VERSION} --extra dev 
+        mypy --config-file pyproject.toml
+        "${CMAKE_SOURCE_DIR}/src/core/retargeting_engine/python"
+    WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+)
+
```
aristarkhovNV, line 43: so this runs against the retargeting_engine but the working directory is retargeting_engine_tests? Where does the actual test live?

nvddr: The test is the command itself, it just does a mypy typecheck on the source.

PR #42, opened 2026-01-09
Add source nodes for the retargeting engine

nvddr:


examples/retargeting/python/sources_example.py
```diff
@@ -0,0 +1,211 @@
...
+                        if right_ctrl_active:
+                            print(f"      Trigger: {right_trigger:4.2f}")
+                        
+                        print()
+                    
+                    frame_count += 1
+                    time.sleep(0.016)  # ~60 FPS
+            
+            except KeyboardInterrupt:
+                print("\nInterrupted by user")
```
aristarkhovNV: should not be needed, with all the RAII stuff we should be able to rely on the default interrupt handling

src/core/retargeting_engine/python/__init__.py
```diff
@@ -3,11 +3,16 @@
 
 """Retargeting Engine for TeleopCore."""
 
-from . import interface
-from . import tensor_types
+# Submodules are available but not eagerly imported
+# Users should import them explicitly:
+#   from teleopcore.retargeting_engine import interface
+#   from teleopcore.retargeting_engine import tensor_types
+#   from teleopcore.retargeting_engine import sources
```
aristarkhovNV: what is this comment about?

examples/retargeting/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20)
+
+# Python examples don't need CMake - they use uv
+# Just install them to the appropriate location
+install(DIRECTORY python/
```
aristarkhovNV, line 8: do we actually rely on the install flow at the moment and is the install output being tested?

src/core/retargeting_engine/python/__init__.py
nvddr: There was some issue that cursor tried to fix, resolved.

examples/retargeting/CMakeLists.txt
nvddr: Yes, for the examples, we need to install, and run uv from the installed directory. It tests the whole library wheel package.

nvddr: I will do a follow-up to change this behavior.

examples/retargeting/python/sources_example.py
```diff
@@ -0,0 +1,202 @@
...
+    
+    # ========================================================================
+    # Step 3: Create OpenXR session
+    # ========================================================================
+    print("\n[Step 3] Creating OpenXR session...")
+    oxr_session = oxr.OpenXRSession.create("RetargetingSourcesExample", required_extensions)
+    if oxr_session is None:
+        print("  ✗ Failed to create OpenXR session")
+        print("\nNote: This example requires an OpenXR runtime (e.g., SteamVR, Oculus)")
+        return 1
```
aristarkhovNV: there should be an exception here, no need to check for null

src/core/retargeting_engine/python/sources/controllers_source.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # Use in retargeting graph
+        gripper = GripperRetargeter(name="gripper")
+        connected = gripper.connect({
+            "controller_left": controllers.output("controller_left"),
+            "controller_right": controllers.output("controller_right")
+        })
+    """
+
+    def __init__(self, controller_tracker: Any, session: Any, name: str) -> None:
```
aristarkhovNV: can args be strongly typed?

src/core/retargeting_engine_tests/python/test_sources.py
```diff
@@ -0,0 +1,510 @@
...
+        
+        assert source._name == "controllers"
+        assert source._controller_tracker is tracker
+        assert source._session is session
+    
+    def test_controllers_source_has_no_inputs(self):
+        """Test that ControllersSource has no inputs."""
+        tracker = MockControllerTracker()
+        session = MockSession()
+        source = ControllersSource(tracker, session, name="controllers")
```
aristarkhovNV, line 183: how is the `ControllersSource` happy with the fake session?

nvddr: Type checking only happens in the mypy test target, there's no runtime checking. So it works, it's really ugly but hard to mock properly because we don't have well-defined protocols and the trackers come from python bindings and custom flat-buffer-like types in python.

PR #44, opened 2026-01-10
Add retargeter tuning UI

nvddr:


examples/retargeting/python/multi_retargeter_tuning_demo.py
```diff
@@ -0,0 +1,136 @@
...
+    stage2 = GainOffsetRetargeter("Stage 2", config_file="stage2.json")
+    stage3 = GainOffsetRetargeter("Stage 3", config_file="stage3.json")
+    
+    ui = MultiRetargeterTuningUI(
+        [stage1, stage2, stage3],
+        title="Chained Retargeter Tuning",
+        layout_mode=layout_mode
+    )
+    
+    input_tensor = TensorGroup(stage1._input_types["value"])
```
aristarkhovNV: are you supposed to access private members like this vs. going through the retargeter interface?

examples/retargeting/python/multi_retargeter_tuning_demo_imgui.py
```diff
@@ -0,0 +1,187 @@
...
+)
+from teleopcore.retargeting_engine.tensor_types import FloatType
+
+try:
+    from teleopcore.retargeting_engine.interface import (
+        MultiRetargeterTuningUIImGui,
+        LayoutModeImGui,
+    )
+except ImportError:
+    print("ERROR: ImGui UI not available. Install with: pip install imgui[glfw]")
```
aristarkhovNV: How would you know it is necessarily "ImGui UI not available", could be other stuff. Better let it propagate through, the default handler should print what really happened.

examples/retargeting/python/multi_retargeter_tuning_demo_imgui.py
```diff
@@ -0,0 +1,187 @@
...
+    sys.exit(1)
+
+
+class GainOffsetRetargeter(BaseRetargeter):
```
aristarkhovNV: is it possible to reuse this sample retargfeter between samples?

examples/retargeting/python/multi_retargeter_tuning_demo_imgui.py
```diff
@@ -0,0 +1,187 @@
...
+                      f"S1: {value1:+6.3f} → "
+                      f"S2: {value2:+6.3f} → "
+                      f"S3: {value3:+6.3f}", end="")
+            
+            # Small sleep to avoid spinning
+            time.sleep(0.016)  # ~60 FPS
+    
+    except KeyboardInterrupt:
+        print("\n")
+        print("✓ Stopped by user")
```
aristarkhovNV: should not be necessary, default behavior should terminate the app just fine

examples/retargeting/python/multi_retargeter_tuning_demo.py
```diff
@@ -0,0 +1,136 @@
...
+    
+    frame_count = 0
+    start_time = time.time()
+    
+    while True:
```
aristarkhovNV: might do something like `while ui.visible:`

src/core/retargeting_engine/python/interface/__init__.py
```diff
@@ -32,5 +67,25 @@
...
+        "LayoutMode",
+    ])
+
+if _HAS_IMGUI_TUNING_UI:
+    __all__.extend([
+        "AsyncParameterTuningUIImGui",
+        "ParameterTuningUIImGui",
+        "MultiRetargeterTuningUIImGui",
+        "LayoutModeImGui",
+    ])
```
aristarkhovNV: Retargeting engine should not bring specific UI dependencies. Any chance these could be moved into a separate module `retargeting_engine_ui`? Apps could choose not to import our UI if they implement their own one.

examples/retargeting/python/simple_tunable_demo_imgui.py
```diff
@@ -0,0 +1,153 @@
...
+                      f"Input: {input_tensor[0]:+6.3f} | "
+                      f"Output: {output_value:+6.3f} | "
+                      f"gain={retargeter.gain:.2f} offset={retargeter.offset:.2f}", end="")
+            
+            # Small sleep to avoid spinning
+            time.sleep(0.016)  # ~60 FPS
+    
+    except KeyboardInterrupt:
+        print("\n")
+        print("✓ Stopped by user")
```
aristarkhovNV: remove manual interrupt handling?

src/core/retargeting_engine/python/interface/multi_retargeter_tuning_ui_imgui.py
```diff
@@ -0,0 +1,423 @@
...
+            # Open the window
+            self._open()
+            
+            # Event loop
+            while self._running and not self._stop_event.is_set():
+                if not self._update():
+                    break
+                
+                # Small sleep to avoid spinning
+                time.sleep(0.001)
```
aristarkhovNV: that's still pretty fast spinning, maybe cap at ~120hz, also if configured properly the present should be v-synced

src/core/retargeting_engine/python/interface/parameter_state.py
```diff
@@ -0,0 +1,199 @@
...
+            for param_name, sync_fn in self._sync_functions.items():
+                if param_name in self._parameters:
+                    sync_fn(self._parameters[param_name].get_value())
+    
+    def get(self, name: str) -> Any:
+        """Get parameter value (thread-safe)."""
+        with self._lock:
+            if name not in self._parameters:
+                raise KeyError(f"Parameter '{name}' not found")
+            return self._parameters[name].get_value()
```
aristarkhovNV: individual locked getters are succeptible to a race condition when whole parameter set changed between locks, so two gets separated by time could return inconsistent parameters from different sets.

src/core/retargeting_engine_tests/python/pyproject.toml
```diff
@@ -46,3 +46,8 @@ exclude = [
     "test_.*",
 ]
 
+# Ignore missing imports for optional UI dependencies
+[[tool.mypy.overrides]]
+module = ["FreeSimpleGUI"]
+ignore_missing_imports = true
```
aristarkhovNV: should not have missing imports in tests

examples/retargeting/python/multi_retargeter_tuning_demo_imgui.py
nvddr: If we want UI deps to be optiona, we will need something like this. I moved it to the __init__.py, PTAL and lmk if you have any suggestions.

nvddr: Actually moved the two uis into two separate modules, that solves htis issue

src/core/retargeting_engine/python/interface/parameter_state.py
```diff
@@ -0,0 +1,256 @@
...
+        Register a parameter (internal, called from __init__).
+        
+        The sync function from parameter.sync_fn is called:
+        - Immediately with the initial value (or loaded config value)
+        - On every sync_all() call to update member variables
+        
+        Args:
+            parameter: Parameter specification with optional sync_fn
+        """
+        with self._lock:
```
aristarkhovNV: may not need the lock if we do we expect parameters to be added at run time

src/core/retargeting_engine/python/interface/parameter_state.py
```diff
@@ -0,0 +1,256 @@
...
+                    raise KeyError(f"Parameter '{name}' not found")
+                param_spec = self._parameters[name]
+                if not param_spec.validate(value):
+                    raise ValueError(f"Invalid value for parameter '{name}': {value}")
+                self._values[name] = value
+    
+    def get_all_parameter_specs(self) -> Dict[str, ParameterSpec]:
+        """Get all parameter specifications (immutable, no copy needed)."""
+        with self._lock:
+            return self._parameters
```
aristarkhovNV: if return does not make a deep copy then lock is useless

src/core/retargeting_engine/python/interface/parameter_state.py
```diff
@@ -0,0 +1,256 @@
...
+            print(f"[ParameterState:{self._name}] Failed to load: {e}")
+            return False
+    
+    def reset_to_defaults(self) -> None:
+        """Reset all parameters to their default values."""
+        with self._lock:
+            for param_name, param_spec in self._parameters.items():
+                self._values[param_name] = param_spec.get_default_value()
+    
+    def reset_to_saved(self) -> bool:
```
aristarkhovNV: seems redundant if you can just call `load_from_file()`

src/core/python/requirements.txt
```diff
@@ -1 +1,3 @@
 numpy>=1.19.0
+FreeSimpleGUI>=5.0.0
+imgui[glfw]>=2.0.0
```
aristarkhovNV: these specific UI libs should not be in the core package requirements

nvddr: Moved to extras.

PR #48, opened 2026-01-15
Refactor plugin_utils

jiwenc-nv:
# Consolidate plugin infrastructure: reuse core session and controller tracking

## Summary

Refactors plugins to reuse `core::OpenXRSession` and `core::DeviceIOSession`, eliminating ~570 lines of duplicate code from `plugin_utils`.

## Changes

- **Plugin refactoring**: `controller_synthetic_hands` and `manus` now use `core::OpenXRSession` + `core::ControllerTracker` instead of `plugin_utils::Session` and `plugin_utils::Controllers`
- **New `oxr_utils` module**: Moved `multiply_poses()` to `oxr_utils/math.hpp`; added `pose_conversions.hpp` for FlatBuffers-to-OpenXR conversions
- **Rate-limited warnings**: `DeviceIOSession` now logs tracker failures on first occurrence + every 1000th to reduce log spam
- **CI**: Extended test script to run native C++ tests (`mcap_tests`, `schema_tests`) alongside Python tests

## Removed
- `plugin_utils/session.hpp`, `session.cpp`
- `plugin_utils/controllers.hpp`, `controllers.cpp`

### DIFF
--- a/src/plugins/manus/core/manus_hand_tracking_plugin.cpp
+++ b/src/plugins/manus/core/manus_hand_tracking_plugin.cpp
@@ -112,14 +112,15 @@ void ManusTracker::initialize(const std::string& app_name) noexcept(false)
 
     try
     {
-        // Create session with required extensions - Create() automatically begins the session
-        std::vector<std::string> extensions = { XR_NVX1_DEVICE_INTERFACE_BASE_EXTENSION_NAME };
-#if defined(_WIN32)
-        extensions.push_back(XR_KHR_WIN32_CONVERT_PERFORMANCE_COUNTER_TIME_EXTENSION_NAME);
-#else
-        extensions.push_back(XR_KHR_CONVERT_TIMESPEC_TIME_EXTENSION_NAME);
-#endif
+        // Create ControllerTracker and DeviceIOSession
+        m_controller_tracker = std::make_shared<core::ControllerTracker>();
+        std::vector<std::shared_ptr<core::ITracker>> trackers = { m_controller_tracker };
 
+        // Get required extensions from trackers
+        auto extensions = core::DeviceIOSession::get_required_extensions(trackers);
+        extensions.push_back(XR_NVX1_DEVICE_INTERFACE_BASE_EXTENSION_NAME);
+
+        // Create session with required extensions - Create() automatically begins the session
         m_session = core::OpenXRSession::Create(app_name, extensions);
         if (!m_session)
         {
@@ -130,9 +131,6 @@ void ManusTracker::initialize(const std::string& app_name) noexcept(false)
         // Initialize hand injector
         m_injector.emplace(m_handles.instance, m_handles.session, m_handles.space);
 
-        // Create ControllerTracker and DeviceIOSession
-        m_controller_tracker = std::make_shared<core::ControllerTracker>();
-        std::vector<std::shared_ptr<core::ITracker>> trackers = { m_controller_tracker };
         m_deviceio_session = core::DeviceIOSession::run(trackers, m_handles);
 
         std::cout << "OpenXR session, HandInjector and DeviceIOSession initialized" << std::endl;


src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -128,32 +127,36 @@ void ManusTracker::initialize(const std::string& app_name) noexcept(false)
...
-        config.extensions.push_back(XR_KHR_CONVERT_TIMESPEC_TIME_EXTENSION_NAME);
+        extensions.push_back(XR_KHR_CONVERT_TIMESPEC_TIME_EXTENSION_NAME);
 #endif
-        // Overlay mode required for headless
-        config.use_overlay_mode = true;
 
-        m_session.emplace(config);
-        m_session->begin();
+        m_session = core::OpenXRSession::Create(app_name, extensions);
+        if (!m_session)
```
rwiltz, line 125: We should be consistent across the repo. If we use exceptions, it's strange that we rely on a null check here. Create should throw in the event of a failure.

jiwenc-nv: That's actually a good catch. We should actually still avoid using exceptions...

Did a bit reading here:
https://docs.google.com/document/d/1NDNS0-lBXqsCN9_wt3ZI4am6Zff7Tv3Uo4y3Sg-0ocI/edit?tab=t.0#bookmark=id.3vm7lyxhdaya

Teleop is something should be considered as "safety-critical embedded systems". I'm 99% sure. Let's discuss more tomorrow.

Most importantly: need to tell AI to avoid using exception 😆

jiwenc-nv: OK. After looking at the code base, we're using quite a strong pattern of exceptions in the same base already. Guess that's a much bigger discussion out of scope of this MR.

I will keep the "null check and throw" here as-is (still saves us from segfault and crash). Will start another thread for the exceptions discussion.

deps/cloudxr/docker-compose.test.yaml
```diff
@@ -45,31 +46,63 @@ services:
...
-                echo "✅ PASSED: $$test_script"
-                PASSED=$((PASSED + 1))
-            else
-                echo "❌ FAILED: $$test_script"
-                FAILED=$((FAILED + 1))
-                FAILED_TESTS="$$FAILED_TESTS $$test_script"
-            fi
-            echo ""
-        done
+        # Run native C++ tests
```
aristarkhovNV, line 82: I assume these are specifically GPU tests? lets make that very explicit. Otherwise we already run CPU-only unit tests via ctest immediately after build on that same buiuld machine. If we are pulling all build artifacts here anyway we might consider consolidating all tests under ctest.

.github/workflows/build-ubuntu.yml
```diff
@@ -59,12 +59,14 @@ jobs:
...
+    - name: Upload Release artifacts
       if: matrix.build_type == 'Release'
       uses: actions/upload-artifact@v4
       with:
-        name: teleopcore-install-release
-        path: install/
+        name: teleopcore-release
+        path: |
+          install/
+          build/
```
aristarkhovNV: including the whole build/ will probably upload a ton of intermediate output files? This might consume too much of the storage quota, GH storage is expensive

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
```diff
@@ -9,42 +10,33 @@ namespace plugins
...
-        m_injector.emplace(
-            handles.instance, handles.session, m_controllers->left_aim_space(), m_controllers->right_aim_space());
-    }
-    else
+    // Initialize session - Create() automatically begins the session
+    m_session = core::OpenXRSession::Create("ControllerSyntheticHands", extensions);
+    if (!m_session)
     {
-        m_injector.emplace(handles.instance, handles.session, handles.reference_space);
+        throw std::runtime_error("Failed to create OpenXR session");
```
aristarkhovNV, line 30: let OpenXRSession::Create throw instead, it may also include some additional info about the failure

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -128,44 +112,31 @@ void ManusTracker::initialize(const std::string& app_name) noexcept(false)
...
 
-        m_session.emplace(config);
-        m_session->begin();
-
-#if defined(_WIN32)
-        if (m_session->get_extension_function("xrConvertWin32PerformanceCounterToTimeKHR", &m_convertWin32Time))
-#else
-        if (m_session->get_extension_function("xrConvertTimespecTimeToTimeKHR", &m_convertTimespecTime))
-#endif
+        m_session = core::OpenXRSession::Create(app_name, extensions);
```
aristarkhovNV, line 124: trackers may require additional extensions

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -128,44 +112,31 @@ void ManusTracker::initialize(const std::string& app_name) noexcept(false)
...
-#if defined(_WIN32)
-            error_msg = "Failed to get xrConvertWin32PerformanceCounterToTimeKHR function";
-#else
-            error_msg = "Failed to get xrConvertTimespecTimeToTimeKHR function";
-#endif
-        }
+        // Create ControllerTracker and DeviceIOSession
+        m_controller_tracker = std::make_shared<core::ControllerTracker>();
+        std::vector<std::shared_ptr<core::ITracker>> trackers = { m_controller_tracker };
+        m_deviceio_session = core::DeviceIOSession::run(trackers, m_handles);
```
aristarkhovNV, line 134: we are supposed to gather required extensions from trackers to ensure `m_handles` are compatible

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -128,44 +112,31 @@ void ManusTracker::initialize(const std::string& app_name) noexcept(false)
...
-        // Require the push device extension
-        config.extensions = { XR_NVX1_DEVICE_INTERFACE_BASE_EXTENSION_NAME, XR_MND_HEADLESS_EXTENSION_NAME };
+        // Create session with required extensions - Create() automatically begins the session
+        std::vector<std::string> extensions = { XR_NVX1_DEVICE_INTERFACE_BASE_EXTENSION_NAME };
 #if defined(_WIN32)
-        config.extensions.push_back(XR_KHR_WIN32_CONVERT_PERFORMANCE_COUNTER_TIME_EXTENSION_NAME);
+        extensions.push_back(XR_KHR_WIN32_CONVERT_PERFORMANCE_COUNTER_TIME_EXTENSION_NAME);
 #else
-        config.extensions.push_back(XR_KHR_CONVERT_TIMESPEC_TIME_EXTENSION_NAME);
+        extensions.push_back(XR_KHR_CONVERT_TIMESPEC_TIME_EXTENSION_NAME);
```
aristarkhovNV: manual time conversion seems no longer required? what is responsible for it now? -that entity should request necessary extensions

.github/workflows/build-ubuntu.yml
jiwenc-nv: The retention days are only 1, tho, so the cost should be manageable?

That being said, I don't mind switch to test only `install/` stuff.

deps/cloudxr/docker-compose.test.yaml
jiwenc-nv: Those will be tests that will depend on GPUs, but may or may not be directly during GPU. More likely, they are just native tests that depends on having a runtime (which in term requires the GPU).

src/plugins/controller_synthetic_hands/synthetic_hands_plugin.cpp
jiwenc-nv: I agree per the other way's discussion. But let's update the behavior of `OpenXRSession::Create` in a follow up PR, so that we don't draft this one's scope too long.

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
jiwenc-nv: good catch. This should just be consistent with what we are doing in the controller_synthesis_hands.

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
jiwenc-nv: right. using `core::DeviceIOSession::get_required_extensions` instead. that's the benefits of using the deviceio impl.

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
jiwenc-nv: ditto: `core::DeviceIOSession::get_required_extensions(trackers)` is the way

PR #50, opened 2026-01-16
Teleop utils for easy building of teleop apps.

nvddr:


examples/retargeting/python/sources_example.py
```diff
@@ -32,56 +32,61 @@ def main():
...
-            head_source = HeadSource(head_tracker, session, name="head")
-            controllers_source = ControllersSource(controller_tracker, session, name="controllers")
-            print("  ✓ Created HandsSource")
-            print("  ✓ Created HeadSource")
-            print("  ✓ Created ControllersSource")
+            print("\n[Step 6] Injecting DeviceIO session into sources...")
+            hands_source.set_session(session)
+            head_source.set_session(session)
+            controllers_source.set_session(session)
+            print("  ✓ Session injected into all sources")
```
aristarkhovNV: mid-life session injection seems like a step back in the design, introduces mutable state and implicit dependency that users may forget add

examples/teleop_utils/python/gripper_retargeting_simple.py
```diff
@@ -0,0 +1,232 @@
...
+    # Import retargeter components for inline definition
+    from teleopcore.retargeting_engine.interface import BaseRetargeter, FloatParameter, ParameterState
+    from teleopcore.retargeting_engine.tensor_types import FloatType, ControllerInput
+    from teleopcore.retargeting_engine.interface.tensor_group_type import TensorGroupType
+    from teleopcore.retargeting_engine.interface.tensor_group import TensorGroup
+    from typing import Dict
+except ImportError as e:
+    print(f"Error: {e}")
+    print("Make sure TeleopCore and all modules are built and installed")
+    sys.exit(1)
```
aristarkhovNV: can we rely on default import failures behavior

examples/teleop_utils/python/gripper_retargeting_simple.py
```diff
@@ -0,0 +1,232 @@
...
+                
+                # Print status every 0.5 seconds
+                if session.frame_count % 30 == 0:
+                    elapsed = session.get_elapsed_time()
+                    print(f"[{elapsed:5.1f}s] Left: {left:.2f}  Right: {right:.2f}")
+                
+                time.sleep(0.016)  # ~60 FPS
+        
+        except KeyboardInterrupt:
+            print("\n\nInterrupted by user")
```
aristarkhovNV: lets rely on default interrupt handling for cleaner sample look

examples/teleop_utils/python/multi_source_example.py
```diff
@@ -0,0 +1,228 @@
...
+    # Import retargeter components for inline definition
+    from teleopcore.retargeting_engine.interface import BaseRetargeter, FloatParameter, ParameterState
+    from teleopcore.retargeting_engine.tensor_types import FloatType, ControllerInput, HandInput
+    from teleopcore.retargeting_engine.interface.tensor_group_type import TensorGroupType
+    from teleopcore.retargeting_engine.interface.tensor_group import TensorGroup
+    from typing import Dict
+except ImportError as e:
+    print(f"Error: {e}")
+    print("Make sure TeleopCore and all modules are built and installed")
+    sys.exit(1)
```
aristarkhovNV: let it through

examples/teleop_utils/python/multi_source_example.py
```diff
@@ -0,0 +1,228 @@
...
+                if now - last_print >= 1.0:
+                    elapsed = session.get_elapsed_time()
+                    fps = session.frame_count / elapsed if elapsed > 0 else 0
+                    print(f"[{elapsed:5.1f}s] Left: {left:.2f}  Right: {right:.2f}  FPS: {fps:.1f}")
+                    last_print = now
+                
+                time.sleep(0.016)  # ~60 FPS
+        
+        except KeyboardInterrupt:
+            print("\n\nInterrupted by user")
```
aristarkhovNV: would be nice to remove

src/core/retargeting_engine/python/sources/head_source.py
```diff
@@ -26,39 +26,53 @@ class HeadSource(BaseRetargeter):
...
+    
+    def set_session(self, session: "DeviceIOSession") -> None:
+        """Set the DeviceIO session after it's created.
+        
+        Called by TeleopSession during initialization.
+        
+        Args:
+            session: The DeviceIOSession instance to use
+        """
+        self._session = session
```
aristarkhovNV: is it possible to have the session from the beginning of the object life cycle

src/core/teleop_utils/python/helpers.py
```diff
@@ -0,0 +1,52 @@
...
+"""
+Helper utilities for creating input nodes and retargeters.
+"""
+
+try:
+    from teleopcore.retargeting_engine.sources import HandsSource, ControllersSource, HeadSource
+except ImportError:
+    HandsSource = None
+    ControllersSource = None
+    HeadSource = None
```
aristarkhovNV: is there any point in continuing without the imports?

examples/teleop_utils/python/gripper_retargeting_simple.py
```diff
@@ -0,0 +1,232 @@
...
+            layout_mode="horizontal",
+        )
+    
+    # ==================================================================
+    # Create and run TeleopSession (trackers auto-discovered!)
+    # ==================================================================
+    
+    config = TeleopSessionConfig(
+        app_name="GripperRetargetingSimple",
+        pipeline=pipeline,  # Trackers auto-discovered from pipeline!
```
aristarkhovNV: If you want truly declarative then pass an abstract pipeline description (in other words an instruction how to make the pipeline), not the pipeline object itself. Then real object construction can be delayed, and you won't need to deal with session injection mid-life. Another benefit is the pipeline description could be saved and authored offline, hot reloaded, visualized, etc.

The description may look something like this, does not have to be json, just an example:
```
{
    "nodes": {
        "gripper": {"class": "SimpleGripperRetargeter"},
        "controllers": {"class": "ControllersSource"}
    },
    "edges": [
        {"target" : "gripper.controller_left", "source": "controllers.controller_left"},
        {"target" : "gripper.controller_right", "source": "controllers.controller_right"}
    ]
}
```

When real pipeline instance is constructed, all necessary dependencies would be supplied at run time after parsing and validating the meta description.

examples/retargeting/python/sources_example.py
nvddr: Changed the design, PTAL

examples/teleop_utils/python/gripper_retargeting_simple.py
nvddr: Done. For some reason cursor keeps prefering this.

examples/teleop_utils/python/multi_source_example.py
nvddr: Done

src/core/retargeting_engine/python/sources/head_source.py
nvddr: Changed the API so that we need to explicitly poll device io if not using teleop utils.

examples/teleop_utils/python/gripper_retargeting_simple.py
nvddr: Retargeters need various configurations for initialization. See the alternate design that keeps the retargeting pipeline API the native API that we have built, and avoids the lazy session init.

examples/teleop_utils/python/gripper_retargeting_simple.py
```diff
@@ -0,0 +1,223 @@
...
+        
+        # Get controller inputs
+        left_controller = inputs["controller_left"]
+        right_controller = inputs["controller_right"]
+        left_gripper = outputs["left_gripper"]
+        right_gripper = outputs["right_gripper"]
+        
+        # Index 10 is trigger_value, index 11 is is_active
+        left_active = left_controller[11]
+        left_trigger = left_controller[10] if left_active else 0.0
```
aristarkhovNV: would be nice to define these indices as named constants somewhere

examples/teleop_utils/python/gripper_retargeting_simple.py
```diff
@@ -0,0 +1,223 @@
...
+            
+            # Check exit conditions
+            if elapsed >= args.duration:
+                break
+            if args.ui and not session.is_ui_running():
+                print("\n\nUI closed by user")
+                break
+            
+            # Run one iteration (updates trackers + executes pipeline)
+            result = session.run()
```
aristarkhovNV: `run` usually implies something long, for one iteration try to come up with a more descriptive name such as `step`, `tick`, `one_iteration` or similar

src/core/teleop_utils/python/teleop_session.py
```diff
@@ -0,0 +1,345 @@
...
+        self._sources = [node for node in leaf_nodes if isinstance(node, IDeviceIOSource)]
+        
+        # Create tracker-to-source mapping for efficient lookup
+        self._tracker_to_source: Dict[Any, Any] = {}
+        for source in self._sources:
+            tracker = source.get_tracker()
+            self._tracker_to_source[id(tracker)] = source
+    
+    
+    def _initialize_ui(self) -> bool:
```
aristarkhovNV: Depending on whether `teleop_utils` is intended to be used by consuming products we may not want to put any UI deps in here

examples/teleop_utils/python/gripper_retargeting_simple.py
nvddr: I made the example simpler for now.

examples/retargeting/python/sources_example.py
```diff
@@ -21,63 +24,58 @@
 import time
 import teleopcore.deviceio as deviceio
 import teleopcore.oxr as oxr
-from teleopcore.retargeting_engine.sources import HandsSource, HeadSource, ControllersSource
+from teleopcore.retargeting_engine.deviceio_source_nodes import HandsSource, HeadSource, ControllersSource
```
jiwenc-nv, line 27: not 100% relevant to this PR, but what's our policy for column width limit? this line is more than 100

@aristarkhovNV @nvddr any thoughts?

nvddr: Let's add a python formatter to CI similar to how we have clang-format for cpp. Followup MR.

PR #51, opened 2026-01-16
Add body tracking testbed for Pico WebXR body tracking extension 

gareth-morgan-nv:
Adds a simple testbed app demonstrating WebXR body tracking.

### DIFF
--- a/examples/cxrjs/bodytracking/README.md
+++ b/examples/cxrjs/bodytracking/README.md
@@ -5,10 +5,11 @@ A simple testbed application demonstrating the WebXR Body Tracking API for rende
 ## Overview
 
 This application demonstrates body tracking capabilities by:
-1. Requesting a WebXR session with body-tracking features
+1. Requesting a WebXR session with body-tracking feature
 2. Reading body joint poses each frame from the XR device
-3. Rendering joints as cubes and bones as cylinders
-4. Displaying a mirrored skeleton facing the user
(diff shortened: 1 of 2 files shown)


examples/cxrjs/bodytracking/w3c_proposal/index.bs
```diff
@@ -0,0 +1,301 @@
...
+  "left-elbow",
+  "right-elbow",
+  "left-wrist",
+  "right-wrist",
+  "left-hand",
+  "right-hand"
+};
+
+[Exposed=Window]
+interface XRBodyPico {
```
gareth-morgan-nv: I'm a little dubious of these duplicated class names with the Pico suffix.  Its not something other WebXR extensions do (unlike OXR).  But the alternative would a be more involved version that selects a skeleton format based on whats available.  Any thoughts @nv-jakob ?

jiwenc-nv: Maybe I got a bit out of date on this, but I thought we are going to use https://immersive-web.github.io/body-tracking?

gareth-morgan-nv: Yeah this is our original proposal. Pico have said they can't do this but can give use the Meta WebXR skeleton in a form that will make the translation trivial  (i.e. just the bones that map directly to Pico bones).  @yanziz-nvidia do we have details from Pico of exactly what that will look like?  I'll update this code to match that when we do.

PR #55, opened 2026-01-21
Add Docker-based test fixture

jiwenc-nv:
This is mainly used for running TeleopCore tests against CloudXR runtime, both locally and in CI:

- Volume isolation via Docker Compose project name (teleopcore-test)
- Automatic cleanup on exit via trap handler
- Build artifact validation (checks for single teleopcore wheel)
- GPU runtime detection with warnings
- CI-friendly output with GitHub Actions log groups
- Configurable test list in script header
- Remove pull-cloudxr.yml workflow (now included in test-cloudxr.yml)

### DIFF
--- a/deps/cloudxr/docker-compose.test.yaml
+++ b/deps/cloudxr/docker-compose.test.yaml
@@ -1,0 +1,85 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Test docker-compose file for running TeleopCore tests with CloudXR
+# Extends the base CloudXR compose and adds test runner service
+#
+# Note: Run with a different project name to isolate volumes from dev environment:
+#   docker compose -p teleopcore-test -f docker-compose.yaml -f docker-compose.test.yaml up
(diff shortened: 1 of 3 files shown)


deps/cloudxr/docker-compose.test.yaml
```diff
@@ -0,0 +1,84 @@
...
+      cloudxr-runtime:
+        condition: service_healthy
+    environment:
+      - XR_RUNTIME_JSON=/openxr/share/openxr/1/openxr_cloudxr.json
+      - TEST_SCRIPTS=${TEST_SCRIPTS:-test_extensions.py}
+    volumes:
+      - openxr-volume:/openxr/:ro
+    working_dir: /app/tests
+    entrypoint: ["/bin/bash", "-c"]
+    command:
```
aristarkhovNV, line 39: should this block be a script that could be run outside of CI?

jiwenc-nv: This block is indeed executable outside of CI, via docker compose.

So, effectively: `run_tests_with_cloudxr.sh` -> docker compose -> this snippet.

This way, all the tests will be executed in a container for local executions as well. This helps us to get more consistent results between CI and local enrivonment.

PR #59, opened 2026-01-23
Force runtime to use Quest3 as device profile during testing

jiwenc-nv:


deps/cloudxr/docker-compose.yaml
```diff
@@ -16,6 +16,7 @@ services:
     environment:
       - ACCEPT_EULA=${ACCEPT_CLOUDXR_EULA:-Y}
       - NV_CXR_ENABLE_PUSH_DEVICES=${NV_CXR_ENABLE_PUSH_DEVICES}
+      - NV_DEVICE_PROFILE=Quest3
```
yanziz-nvidia: Should not do this - it should be auto-webrtc. Otherwise it won't respect PICO configs sent from client.

jiwenc-nv: Ah, good to know...

Is `auto-webrtc` the default in 6.0.1 EA container for `xxx-webrtc`? During a discussion earlier with @nv-jakob, we determined that using `auto-webrtc` causing runtime to have for a headset before it can let other HEADLESS clients to connect.

Maybe I only need to hard code this in the composer file for testing.

yanziz-nvidia: based on Slack discussion it's probably better to have a dedicated device profile for HEADLESS 

And in the script, pass a param and then set an env arg to launch docker-compose.yaml to take in the device profile setting is probably the best

jiwenc-nv: Moved to testing

jiwenc-nv: Also added the ARG.

PR #60, opened 2026-01-23
feat: Create a generic Camera Device Plugin interface and add support oak-d

shaosu-nvidia:


src/plugins/camera/README.md
```diff
@@ -0,0 +1,35 @@
...
+## Shared Components (core/)
+
+- **CameraPlugin**: Plugin lifecycle management, threading, OpenXR integration
+- **Mp4Writer**: FFmpeg-based MP4 muxing for H.264 streams
+- **ICamera**: Interface for camera implementations
+- **CameraConfig/RecordConfig**: Configuration structures
+
+## Adding a New Camera Plugin
+
+1. Create a new subdirectory (e.g., `realsense/`)
```
jiwenc-nv: do we want different camera modules to live in different plugin folders?

src/core/camera/inc/camera_plugin_core/camera_config.hpp
```diff
@@ -0,0 +1,25 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+namespace plugins
```
jiwenc-nv: maybe it makes sense to have the `camera` interface to be a module in the `src/core`, then different cameras will have independent plugins.

src/plugins/camera/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
...
+    endif()
+endif()
+
+# Build FFmpeg from source if not available
+if(NOT FFMPEG_PREBUILT AND NOT FFMPEG_USE_SYSTEM)
+    ExternalProject_Add(ffmpeg_external
+        URL https://ffmpeg.org/releases/ffmpeg-6.1.tar.xz
+        URL_HASH SHA256=488c76e57dd9b3bee901f71d5c95eaf1db4a5a31fe46a28654e837144207c270
+        CONFIGURE_COMMAND <SOURCE_DIR>/configure
+            --prefix=${FFMPEG_INSTALL_DIR}
```
jiwenc-nv: I think you will need `--disable-gpl` and `--disable-nonfree`. Please double check: https://www.ffmpeg.org/legal.html

cc: @farbod-nv for viz, who went through the OSRB reviews before.

jiwenc-nv: Ideally, we want to avoid ffmpeg if possible (or be super careful on the license configs)

src/plugins/camera/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
...
+            --disable-doc
+            --disable-network
+            --disable-x86asm
+            --disable-everything
+            --enable-muxer=mp4
+            --enable-protocol=file
+            --enable-static
+            --disable-shared
+            --enable-pic
+        BUILD_COMMAND make -j${CMAKE_BUILD_PARALLEL_LEVEL}
```
jiwenc-nv: how long does it add to the build time?

src/plugins/camera/CMakeLists.txt
shaosu-nvidia: got it removed thanks!

src/core/camera/inc/camera_plugin_core/camera_config.hpp
shaosu-nvidia: yeah I have thought of that. but the shared part such as camera_plugin.hpp/cpp the rawdata_writer (I am thinking of to organize it into a /sink folder so in the future it can be a stream) are plugin related so moving them to core sounds a little bit weird. wdyt.

src/plugins/camera/CMakeLists.txt
shaosu-nvidia: I timed it and depthai takes ~3min to build.

src/plugins/camera/oakd/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+    add_executable(camera_plugin_oakd
+        main.cpp
+        oakd_camera.cpp
+    )
+
+    target_include_directories(camera_plugin_oakd
+        PRIVATE
+            ${CMAKE_CURRENT_SOURCE_DIR}/inc
+            ${CAMERA_PLUGIN_INC_DIR}
+    )
```
aristarkhovNV: all includes should be pulled in via `target_link_libraries`, don't use `target_include_directories`

src/plugins/camera/README.md
shaosu-nvidia: I m not sure what would be the best way to organize the folders.

src/plugins/oakd/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+
+    target_link_libraries(camera_plugin_oakd
+        PRIVATE
+            depthai::core
+            camera_plugin_core
+    )
+
+    set_target_properties(camera_plugin_oakd PROPERTIES
+        OUTPUT_NAME camera_plugin
+    )
```
aristarkhovNV: can we keep `camera_plugin_oakd` output name? Or rename the target itself

src/plugins/camera/core/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+add_library(camera_plugin_core SHARED
```
aristarkhovNV: does it need to be shared? static lib would be easier to manage

src/core/camera/inc/camera_core/camera_config.hpp
```diff
@@ -0,0 +1,25 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aristarkhovNV, line 1: put `inc/core` under `src/plugins/camera/core` so the whole core library is contained within one directory

src/plugins/camera/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+# Shared headers location
+# ==============================================================================
+set(CAMERA_PLUGIN_INC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/inc")
+
+# ==============================================================================
+# Build shared core library
+# ==============================================================================
+message(STATUS "Building camera plugin core library")
+add_subdirectory(core)
+add_library(isaac::camera_plugin ALIAS camera_plugin_core)
```
aristarkhovNV: This top level file should only include two subdirs and do nothing else. All the additional setup should be contained within nested cmakelists

src/core/camera/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+find_package(Threads REQUIRED)
+
+add_library(camera_plugin_core STATIC
+    camera_plugin.cpp
+    rawdata_writer.cpp
+)
```
aristarkhovNV, line 9:
```suggestion
target_include_directories(camera_plugin_core INTERFACE inc)
)
```

src/plugins/camera/oakd/main.cpp
```diff
@@ -0,0 +1,150 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "camera_plugin.hpp"
```
aristarkhovNV: use <> and prepend the moduke name
```suggestion
#include <camera_plugin_core/camera_plugin.hpp>
```

src/core/camera/inc/camera_core/camera_config.hpp
aristarkhovNV, line 1: inside inc/ make another directory level like camera_plugin_core or similar. This will force all consuming includes to explicitly spell out that module name in front of the header file name

src/plugins/camera/README.md
aristarkhovNV: Sounds reasonable to have a directory at least per vendor. Some plugins may support multiple devices if they all share the same implementation.

src/core/camera/CMakeLists.txt
shaosu-nvidia: it seems I need to make it PUBLIC since it's internally consumed?

aristarkhovNV: internally consume by relative path: `#include "inc/camera_plugin_core/header.hpp"`

src/plugins/oakd/oakd_camera.hpp
```diff
@@ -0,0 +1,60 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aristarkhovNV, line 1: private headers never exposed to other modules may be placed next to cpp files

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+    int retry_count = 0;
+    while (m_running && !m_stop_requested)
+    {
+        if (init_camera())
+        {
+            break;
+        }
+
+        retry_count++;
+        std::cerr << "Camera init failed (attempt " << retry_count << "). Retrying in " << m_retry_interval << "s..."
```
aristarkhovNV: why do we expect the init to fail?

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+    m_writer = std::make_unique<RawDataWriter>(m_record_config);
+    m_writer->open();
+}
+
+bool CameraPlugin::init_camera()
+{
+    try
+    {
+        m_camera = m_camera_factory(m_camera_config);
+        m_camera->start();
```
aristarkhovNV: can we require the factory to create the camera in the started state, so no separate start() would be required

src/plugins/oakd/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+    while (!g_stop_requested.load(std::memory_order_relaxed) && !plugin->stop_requested())
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    }
+
+    // Request plugin stop
+    plugin->request_stop();
+
+    // Plugin destructor handles cleanup
+    plugin.reset();
```
aristarkhovNV: plugin dtor should implicitly stop the camera, so no explicit `request_stop()` should be required

src/plugins/oakd/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+    catch (const std::exception& e)
+    {
+        std::cerr << "Failed to create plugin: " << e.what() << std::endl;
+        return 1;
+    }
+
+    std::cout << "Plugin running. Press Ctrl+C to stop." << std::endl;
+
+    // Wait for stop signal
+    while (!g_stop_requested.load(std::memory_order_relaxed) && !plugin->stop_requested())
```
aristarkhovNV: `plugin->stop_requested()` will likely never happen because we only ever call `plugin->request_stop()` below

src/plugins/oakd/oakd_camera.cpp
```diff
@@ -0,0 +1,139 @@
...
+    m_h264_queue.reset();
+    m_device.reset();
+    m_pipeline.reset();
+
+    std::cout << "OAK-D camera pipeline stopped" << std::endl;
+}
+
+bool OakDCamera::is_running() const
+{
+    return m_running && m_device && !m_device->isClosed();
```
aristarkhovNV: additional `m_running` flag is redundant, we can rely on just the existence of `m_device`

src/core/camera/CMakeLists.txt
shaosu-nvidia: Since the whole camera_plugin_core folder is moved under core/camera. I think it makes sense to not use relative path?

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+    std::cout << "Camera Plugin initialized and running" << std::endl;
+}
+
+CameraPlugin::~CameraPlugin()
+{
+    std::cout << "------------------------------------------------------------" << std::endl;
+    std::cout << "Shutting down Camera Plugin..." << std::endl;
+
+    m_running = false;
+    m_stop_requested = true;
```
aristarkhovNV: m_running seems redundant, `m_stop_requested` should be enough to handle stopping the worker thread

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    if (m_thread.joinable())
```
aristarkhovNV: no need to check, must always be joinable

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+    m_session.emplace(config);
+    m_session->begin();
+
+    std::cout << "OpenXR session initialized" << std::endl;
+}
+
+void CameraPlugin::init_writer()
+{
+    m_writer = std::make_unique<RawDataWriter>(m_record_config);
+    m_writer->open();
```
aristarkhovNV: create opened, don't require a separate `open()`

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+        retry_count++;
+        std::cerr << "Camera init failed (attempt " << retry_count << "). Retrying in " << m_retry_interval << "s..."
+                  << std::endl;
+        std::this_thread::sleep_for(std::chrono::seconds(m_retry_interval));
+    }
+
+    if (!m_camera)
+    {
+        std::cerr << "Failed to initialize camera. Stopping." << std::endl;
+        m_stop_requested.store(true, std::memory_order_relaxed);
```
aristarkhovNV: should not touch `m_stop_requested` from within the worker thread

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+void CameraPlugin::worker_thread()
+{
+    std::cout << "------------------------------------------------------------" << std::endl;
+    std::cout << "Running capture loop..." << std::endl;
+
+    constexpr auto status_interval = std::chrono::seconds(10);
+    auto last_status_time = std::chrono::steady_clock::now();
+
+    // Camera initialization with retry
```
aristarkhovNV: initialize on the main thread

src/plugins/oakd/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
aristarkhovNV: busy waiting on main thread seems like a waste, what is the purpose of the worker thread then? Might just pull frames right here instead of doing nothing.

src/core/camera/CMakeLists.txt
aristarkhovNV: relocating the folder didn't change relative paths within that folder. Now that it is no longer under plugins might just rename `camera_plugin_core` -> `camera_core`. Otherwise the module should still consume its own headers by relative paths like "inc/camera_core/blah.hpp"

src/core/camera/inc/camera_plugin_core/camera_interface.hpp
```diff
@@ -0,0 +1,75 @@
...
+#pragma once
+
+#include <camera_plugin_core/camera_config.hpp>
+
+#include <chrono>
+#include <cstdint>
+#include <optional>
+#include <vector>
+
+namespace plugins
```
jiwenc-nv: turns out I said something wrong during the workshop...

this should be: `namespace core` and you do need to change the code...

src/core/camera/inc/camera_plugin_core/camera_interface.hpp
```diff
@@ -0,0 +1,75 @@
...
+{
+namespace camera
```
jiwenc-nv: wondering whether @aristarkhovNV will have opinions on nested namespace. I'm OK with it if no strong push back from others.

src/core/camera/inc/camera_plugin_core/camera_interface.hpp
aristarkhovNV: Top level namespace should be something more specific than `plugins` or `core`

src/core/camera/inc/camera_plugin_core/camera_interface.hpp
aristarkhovNV: I don't see a big need for deeply nested namespaces, but not opposed either

src/core/camera/inc/camera_plugin_core/camera_interface.hpp
aristarkhovNV: more specific == identifieble with nv, isaac and teleop

.github/workflows/build-ubuntu.yml
```diff
@@ -23,18 +26,39 @@ jobs:
         submodules: recursive
 
     - name: Install uv
-      uses: astral-sh/setup-uv@v7
+      run: pip install uv
```
aristarkhovNV: why replace the action

.github/workflows/build-ubuntu.yml
```diff
@@ -23,18 +26,39 @@ jobs:
...
+      run: |
+        if [ ! -d "${{ env.VCPKG_ROOT }}/.git" ]; then
+          git clone https://github.com/microsoft/vcpkg.git ${{ env.VCPKG_ROOT }}
+        fi
+        cd ${{ env.VCPKG_ROOT }}
+        git fetch origin
+        git checkout a9eee3b18df395dbb8be71a31bd78ea441056e42
+        if [ ! -f "${{ env.VCPKG_ROOT }}/vcpkg" ]; then
+          ./bootstrap-vcpkg.sh
+        fi
```
aristarkhovNV: consider using this community action instead of manual setup

```yml
- name: Install vcpkg dependencies with caching
  uses: lukka/run-vcpkg@v11
```

.github/workflows/build-windows.yml
```diff
@@ -31,26 +34,57 @@ jobs:
...
+        $vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
+        $vsPath = & $vswhere -latest -property installationPath
+        $vcvarsall = "$vsPath\VC\Auxiliary\Build\vcvars64.bat"
+        cmd /c """$vcvarsall"" amd64 && set" | ForEach-Object {
+          if ($_ -match "^([^=]+)=(.*)$") {
+            [System.Environment]::SetEnvironmentVariable($matches[1], $matches[2], "Process")
+            echo "$($matches[1])=$($matches[2])" >> $env:GITHUB_ENV
+          }
+        }
+      shell: pwsh
```
aristarkhovNV: what was wrong with `ilammy/msvc-dev-cmd@v1`

cmake/usb-1.0/usb-1.0Config.cmake
```diff
@@ -0,0 +1,54 @@
+# usb-1.0Config.cmake
```
aristarkhovNV: if we go with vcpkg lets use the default libusb port (don't customize) - yes it will require a couple more apt deps but that's fine

.github/workflows/build-ubuntu.yml
shaosu-nvidia: reverted

.github/workflows/build-windows.yml
shaosu-nvidia: reverted

cmake/usb-1.0/usb-1.0Config.cmake
shaosu-nvidia: Thanks for help, this has been deleted.

.github/workflows/build-ubuntu.yml
shaosu-nvidia: this is due to:
The action lukka/run-vcpkg@v11 is not allowed in NVIDIA/IsaacTeleop because all actions must be from a repository owned by your enterprise, created by GitHub, verified in the GitHub Marketplace, or match one of the patterns: BSFishy/pip-action@v1, DavidAnson/markdownlint-cli2-action@07035fd053f7be764496c0f8d8f9f41f98305101, EndBug/add-and-commit@*, EnricoMi/publish-unit-test-result-action@*, JamesIves/github-pages-deploy-action@*, Jimver/cuda-toolkit@*, Mattraks/delete-workflow-runs@5bf9a1dac5c4d041c029f0a8370ddf0c5cb5aeb7, NVIDIA/blossom-action@*, NVIDIA/gitlab-answer-app/*, NVIDIA/spark-rapids-common@*, SonarSource/sonarqube-scan-action@*, TreTuna/sync-branches@1.4.0, Wandalen/wretry.action/post@*, abbbi/github-actions-tune@*, actions-cool/pr-approve@v0.0.1-beat, actions-ecosystem/action-add-assignees@v1, actions-ecosystem/action-add-labels@*, actions-ecosystem/action-regex-match@*, actions-ecosystem/action-remove-labels@*, actions/cache@*, actions/checkout@*, actions/checkout@08c...

src/core/camera/inc/camera_plugin_core/camera_interface.hpp
shaosu-nvidia: I have made it to a single layer namespace core as other modules.

examples/oxr/python/test_oakd_camera.py
```diff
@@ -0,0 +1,145 @@
...
+
+            while time.time() - start_time < duration:
+                # Poll plugin health every ~1 second
+                elapsed = time.time() - start_time
+                if int(elapsed) > check_count:
+                    check_count = int(elapsed)
+                    try:
+                        plugin.check_health()
+                        print(f"  [{check_count:3d}s] Plugin healthy, recording...")
+                    except pm.PluginCrashException as e:
```
aristarkhovNV: let it propagate and terminate the program, no need for special handling

examples/oxr/python/test_oakd_camera.py
```diff
@@ -0,0 +1,145 @@
...
+
+            print()
+            print(f"  ✓ Recording completed ({duration:.1f} seconds)")
+
+    except pm.PluginCrashException as e:
+        print(f"  ✗ Plugin crashed during shutdown: {e}")
+        return False
+    except Exception as e:
+        print(f"  ✗ Error: {e}")
+        return False
```
aristarkhovNV: no need for special handling, the default behavior is already doing what we want - terminate, print the stack and error message

src/core/camera/inc/camera_plugin_core/camera_interface.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @brief Stop the camera capture pipeline
+     */
+    virtual void stop() = 0;
+
+    /**
+     * @brief Check if the camera pipeline is running
+     */
+    virtual bool is_running() const = 0;
```
aristarkhovNV: get rid of `start()`, `stop()`, `is_running()`. Instead rely on object life time for capture session, rely on `get_frame` to throw an exception if something unexpected happens to the camera at runtime.

src/core/camera/inc/camera_plugin_core/camera_plugin.hpp
```diff
@@ -0,0 +1,106 @@
...
+private:
+    void worker_thread();
+    bool init_camera();
+    void init_writer();
+    void init_openxr_session();
+
+    // Configuration
+    CameraFactory m_camera_factory;
+    CameraConfig m_camera_config;
+    RecordConfig m_record_config;
```
aristarkhovNV: `m_camera_config` and `m_record_config` are not needed - only used to pass data between functions once, should be passed as arguments instead.

src/core/camera/inc/camera_plugin_core/camera_plugin.hpp
```diff
@@ -0,0 +1,106 @@
...
+    std::string m_plugin_root_id;
```
aristarkhovNV: `m_record_config` assigned but not used other than logging, remove.

src/core/camera/inc/camera_plugin_core/frame_metadata_tensor.hpp
```diff
@@ -0,0 +1,318 @@
...
+    uint64_t m_push_success_count{ 0 };
+    uint64_t m_push_fail_count{ 0 };
+
+    // Function pointers
+    PFN_xrCreatePushTensorCollectionNV m_xrCreatePushTensorCollectionNV{ nullptr };
+    PFN_xrPushTensorCollectionDataNV m_xrPushTensorCollectionDataNV{ nullptr };
+    PFN_xrDestroyPushTensorCollectionNV m_xrDestroyPushTensorCollectionNV{ nullptr };
+#ifdef XR_USE_TIMESPEC
+    PFN_xrConvertTimespecTimeToTimeKHR m_xrConvertTimespecTimeToTimeKHR{ nullptr };
+#endif
```
aristarkhovNV: `m_xrConvertTimespecTimeToTimeKHR` seems unused

src/core/camera/inc/camera_core/rawdata_writer.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace core
+{
+
+/**
+ * @brief Raw H.264 file writer
+ *
+ * Writes H.264 NAL units directly to a file without container.
+ */
+class RawDataWriter
```
aristarkhovNV, line 22: this appears to be just a wrapper on top of `std::ofstream`, does not add functionality other than some extra logging that could be totally skipped. Consider using `std::ofstream` directly.

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,184 @@
...
+            break;
+        }
+
+        // Get and write H.264 frames
+        auto frame = m_camera->get_frame();
+        if (frame && m_writer)
+        {
+            if (m_writer->write(frame->data))
+            {
+                m_frame_count.fetch_add(1, std::memory_order_relaxed);
```
aristarkhovNV: terminate on write error

src/plugins/oakd/CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+            "Install on Ubuntu/Debian:\n"
+            "  sudo apt install libudev-dev autoconf automake libtool pkg-config\n"
+            "\n"
+            "Or disable OAK-D plugin:\n"
+            "  cmake -B build -DBUILD_PLUGINS=OFF\n"
+            "================================================================================\n"
+        )
+    endif()
+    message(STATUS "Found libudev (vcpkg will build libusb)")
+endif()
```
aristarkhovNV: this should be handled through vcpkg entirely, don't add special case for any transitive dependencies in our build code.

src/plugins/oakd/CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(LIBUSB REQUIRED IMPORTED_TARGET libusb-1.0)
+
+add_library(usb-1.0 ALIAS PkgConfig::LIBUSB)
+set(usb-1.0_FOUND TRUE)
+message(STATUS "usb-1.0: using vcpkg libusb via pkg-config")
+]=])
+
+set(usb-1.0_DIR "${USB10_CONFIG_DIR}" CACHE PATH "" FORCE)
+message(STATUS "Generated usb-1.0 config at ${USB10_CONFIG_DIR}")
```
aristarkhovNV: let vcpkg handle that

vcpkg.json
```diff
@@ -0,0 +1,31 @@
...
+        "lz4",
+        "liblzma",
+        "eigen3",
+        "yaml-cpp",
+        "fmt",
+        "magic-enum",
+        "neargye-semver",
+        "fp16",
+        "libusb",
+        "openssl",
```
aristarkhovNV: we should be able to use DepthAI provided port files from `depthai-core/cmake/ports` to not specify all its deps manually

src/core/camera/inc/camera_plugin_core/frame_metadata_tensor.hpp
shaosu-nvidia: sorry this file should not be checked in. My bad. have some local tensor api stuff. deleted

src/plugins/oakd/CMakeLists.txt
shaosu-nvidia: Resolve the comments related to vcpkg

src/plugins/oakd/CMakeLists.txt
shaosu-nvidia: Resolving vcpkg related comments

src/core/camera/inc/camera_core/rawdata_writer.hpp
shaosu-nvidia: I'm thinking to make this writer a example sink to the plugin, I think we may have more sinks in the future so would like to separate the data sink part from the plugin module. Wdyt?

src/core/camera/camera_plugin.cpp
shaosu-nvidia: remove the retry logic here and let it crash if any exception may happen.

src/plugins/oakd/main.cpp
shaosu-nvidia: I have switched to do the loop on the main thread. So need some special handle in the main.cpp to terminate the plugin as well as properly calling all the destructor. PTAL

cmake/Hunter/config.cmake
```diff
@@ -0,0 +1,166 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Hunter package configuration for DepthAI dependencies.
+# Derived from depthai-core v2.29.0 cmake/Hunter/config.cmake
+# https://github.com/luxonis/depthai-core
```
aristarkhovNV: why do we need to derive our own, can we take their existing config?

cmake/HunterGate.cmake
```diff
@@ -0,0 +1,543 @@
+# Copyright (c) 2013-2019, Ruslan Baratov
```
aristarkhovNV: can we fetch this file from github instead of checking in a copy?

src/core/camera/inc/camera_core/record_config.hpp
```diff
@@ -0,0 +1,26 @@
...
+{
+
+/**
+ * @brief Configuration for raw H.264 file recording
+ */
+struct RecordConfig
+{
+    std::string output_path; // Explicit path, or empty for auto-naming
+    std::string output_dir = "./recordings";
+    bool auto_name = true;
```
aristarkhovNV: comment above mentiones empty output_path for auto naming, do we still need the additional bool flag for that?

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,105 @@
...
+              << std::setprecision(1) << fps << " fps)" << std::endl;
+    std::cout << "Plugin stopped" << std::endl;
+    std::cout << "============================================================" << std::endl;
+}
+
+void CameraPlugin::init_openxr_session()
+{
+    std::vector<std::string> extensions = { XR_MND_HEADLESS_EXTENSION_NAME, XR_EXTX_OVERLAY_EXTENSION_NAME };
+
+    m_session = core::OpenXRSession::Create("Camera", extensions);
```
aristarkhovNV: looks like the session is initialized but never used? Do we need a session at all?

src/core/camera/camera_plugin.cpp
```diff
@@ -0,0 +1,105 @@
...
+    if (!m_session)
+    {
+        throw std::runtime_error("Failed to create OpenXR session");
+    }
+
+    std::cout << "OpenXR session initialized" << std::endl;
+}
+
+void CameraPlugin::start()
```
aristarkhovNV: if this never returns lets name this function "capture_loop" or similar

src/core/camera/inc/camera_core/rawdata_writer.hpp
aristarkhovNV: If there will be more writers then they all should conform to an interface, could be added as a follow up.

src/plugins/oakd/main.cpp
```diff
@@ -0,0 +1,128 @@
...
+        g_plugin = &plugin;
+
+        std::cout << "Plugin running. Press Ctrl+C to stop." << std::endl;
+        plugin.start();
+
+        g_plugin = nullptr;
+    }
+    catch (const std::exception& e)
+    {
+        std::cerr << "Failed to create plugin: " << e.what() << std::endl;
```
aristarkhovNV: error could occur during capture loop as well, not only at creation

src/plugins/oakd/oakd_camera.cpp
```diff
@@ -0,0 +1,107 @@
...
+    m_h264_queue = m_device->getOutputQueue("h264", 8, false);
+
+    std::cout << "OAK-D camera pipeline started" << std::endl;
+}
+
+OakDCamera::~OakDCamera()
+{
+    m_h264_queue.reset();
+    m_device.reset();
+    m_pipeline.reset();
```
aristarkhovNV: smart pointers will be autoreleased, no need for explicit `.reset()`. You may skip defining `~OakDCamera()` altogether and it should still work.

src/core/camera/camera_plugin.cpp
shaosu-nvidia: Sorry that was added to check if the openxr session can be retrieved, removed and will add when integrating with the tensor api.

cmake/Hunter/config.cmake
shaosu-nvidia: I think the idea behind this file is it would include all the dependencies that will be needed for the project and the huntergate will fetch them all before the top project. So if new huntergate dependency is added (though I doubt), we need to have all their dependencies added to this file as well?

src/plugins/oakd/main.cpp
shaosu-nvidia: I removed the try catch and let the exception propagate.

src/core/camera/inc/camera_core/rawdata_writer.hpp
shaosu-nvidia: added a TODO. 😃

src/plugins/oakd/main.cpp
aristarkhovNV: sorry I didn't mean to remove the try-catch, just to fix the error message. In c++ we have to catch at the top level, otherwise uncaught exception will crash the program

shaosu-nvidia: Sounds good. I see we still need to properly release all the component for the exception case. Added them back.

cmake/Hunter/config.cmake
shaosu-nvidia: I alternatively download the huntergate config.cmake for depthai from the repo. And the top level is just a aggregation so new huntergate project is added we can add the config file download similarly.

PR #61, opened 2026-01-24
oxr_session: Ensure OpenXR env variables are set

jiwenc-nv:
1. XR_RUNTIME_JSON for specifying the OpenXR runtime.
2. XDG_RUNTIME_DIR for IPC communication between the application and containerized OpenXR runtime.

### DIFF
--- a/src/core/oxr/cpp/oxr_session.cpp
+++ b/src/core/oxr/cpp/oxr_session.cpp
@@ -36,13 +36,20 @@ void ensure_env_set(const char* env_name, const std::string& default_value)
     std::cerr << "Warning: " << env_name << " environment variable is not set." << std::endl;
     std::cerr << "  Please set it before running, e.g.:" << std::endl;
     std::cerr << "    export " << env_name << "=" << default_value << std::endl;
+    std::cerr << "  or set TELEOPCORE_DISABLE_CXR_ENV_CHECKS to disable this check" << std::endl;
 
     throw std::runtime_error("Environment variable " + std::string(env_name) + " is not set");
 }
 
 // Ensure required environment variables are set before xrCreateInstance.
-void ensure_runtime_configured()
+void ensure_cloudxr_runtime_configured()
 {
+    // Check the TELEOPCORE_DISABLE_CXR_ENV_CHECKS
+    if (std::getenv("TELEOPCORE_DISABLE_CXR_ENV_CHECKS") != nullptr)
+    {
+        return;
+    }
+
     const std::string home = get_home_dir();
 
     // NV_CXR_RUNTIME_DIR - required by some OpenXR runtimes for IPC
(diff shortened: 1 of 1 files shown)


src/core/oxr/cpp/oxr_session.cpp
```diff
@@ -3,13 +3,63 @@
...
+void ensure_runtime_configured()
+{
+    const std::string home = get_home_dir();
+
+    // NV_CXR_RUNTIME_DIR - required by some OpenXR runtimes for IPC
+    ensure_env_set("NV_CXR_RUNTIME_DIR", home + "/.cloudxr/run");
+
+    // XR_RUNTIME_JSON - tells OpenXR loader which runtime to use
+    ensure_env_set("XR_RUNTIME_JSON", home + "/.cloudxr/share/openxr/1/openxr_cloudxr.json");
+}
```
nv-jakob, line 60: I'm not too enthused about doing this inside of the code, I would prefer a `set_cloudxr_env.sh` script. Which sets these up, especially these will only work for teleop-core application, not the rendering application.

jiwenc-nv: Having a `set_cloudxr_env.sh` that people can source is nice. Will add that and make things like `run_isaac_lab.sh` to source it.

jiwenc-nv: But we having the `oxr_session` to set default is still nice. At least people using TeleopCore utils get the benefits of not need to worry about the env var.

At the end of the day, env var is annoying it's still a global state...

jiwenc-nv: Discuss a bit internally. Let's do set the default value for now.

src/core/oxr/cpp/oxr_session.cpp
```diff
@@ -3,13 +3,57 @@
...
+    // Not set - warn and set default
+    std::cerr << "Warning: " << env_name << " environment variable is not set." << std::endl;
+    std::cerr << "  Please set it before running, e.g.:" << std::endl;
+    std::cerr << "    export " << env_name << "=" << default_value << std::endl;
+
+    throw std::runtime_error("Environment variable " + std::string(env_name) + " is not set");
+}
+
+// Ensure required environment variables are set before xrCreateInstance.
+void ensure_runtime_configured()
```
nv-jakob: Needs to be explicitly call out that it's checking that it's cloudxr/cxr that it's checking for.

src/core/oxr/cpp/oxr_session.cpp
```diff
@@ -3,13 +3,57 @@
...
+{
+    const std::string home = get_home_dir();
+
+    // NV_CXR_RUNTIME_DIR - required by some OpenXR runtimes for IPC
+    ensure_env_set("NV_CXR_RUNTIME_DIR", home + "/.cloudxr/run");
+
+    // XR_RUNTIME_JSON - tells OpenXR loader which runtime to use
+    ensure_env_set("XR_RUNTIME_JSON", home + "/.cloudxr/share/openxr/1/openxr_cloudxr.json");
```
nv-jakob, line 59: Add a way to disable these checks, that should be put in the error message.

`"To disable these checks set the env variable NV_DISABLE_ENV_CHECKS"`

src/core/oxr/cpp/oxr_session.cpp
jiwenc-nv: renaming to `ensure_cloudxr_runtime_configured`

src/core/oxr/cpp/oxr_session.cpp
jiwenc-nv: Added `TELEOPCORE_DISABLE_CXR_ENV_CHECKS`, renamed it to be consistent with the reset of code base.

PR #62, opened 2026-01-24
Add support for Python 3.10

jiwenc-nv:
Turns out https://github.com/nvlabs/GR00T-WholeBodyControl is still on Python 3.10, we'd better support that version as well.


README.md
```diff
@@ -9,7 +9,7 @@ SPDX-License-Identifier: Apache-2.0
 
 **The unified framework for high-fidelity ego-centric and robotics data collection.**
 
-[![Python](https://img.shields.io/badge/python-3.11-blue.svg)](https://www.python.org/downloads/)
+[![Python](https://img.shields.io/badge/python-3.10%2B-blue.svg)](https://www.python.org/downloads/)
```
aristarkhovNV, line 12: why is there %2B in the url?

src/core/schema_tests/python/pyproject.toml
```diff
@@ -5,7 +5,7 @@
 name = "teleopcore-schema-tests"
 version = "0.0.0"  # Internal tests - not versioned
 description = "Tests for TeleopCore schema types"
-requires-python = ">=3.11"
+requires-python = ">=3.10,<3.12"
```
aristarkhovNV, line 8: I wonder if there is way to centralize the python version in one place somewhere

README.md
jiwenc-nv: That's a "+" sign, so it reads Python 3.10+

src/core/schema_tests/python/pyproject.toml
jiwenc-nv: In Core, we have it set by CMake. But guess we are not doing it as well for other libs.

https://github.com/NVIDIA/IsaacTeleop/blob/main/src/core/python/pyproject.toml.in#L16

jiwenc-nv: Also templates are annoying...

PR #66, opened 2026-01-29
Add devcontainer configuration for Isaac Teleop

aristarkhovNV:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Isaac Teleop Dev Container",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
```
jiwenc-nv, line 3: why not using 24.04?

aristarkhovNV: I believe this is our baseline, same as for the runtime. Everything builds and works just fine with this version. Can bump at any moment if needed.

jiwenc-nv: I see. Sounds good. Let's discuss 24.04 in a separate thread down the road.

PR #67, opened 2026-01-29
Remove ctor definitions from OpenXRSessionHandles struct

aristarkhovNV:


src/core/oxr_utils/cpp/inc/oxr_utils/oxr_session_handles.hpp
jiwenc-nv, line 14: If I do 

```
OpenXRSessionHandles h;
```

C++ doesn't guarantee it will be default initialized? Did I remember this right? 

shall we do:

```
struct OpenXRSessionHandles {
    XrInstance instance = XR_NULL_HANDLE;
    XrSession session = XR_NULL_HANDLE;
    XrSpace space = XR_NULL_HANDLE;
    PFN_xrGetInstanceProcAddr xrGetInstanceProcAddr = nullptr;
};
```

instead?

aristarkhovNV: yes good point

PR #68, opened 2026-01-29
Add a Hello Tensor sample to demonstrate Generic Tensor Interface

jiwenc-nv:


examples/hello_tensor/hello_tensor_plugin.cpp
```diff
@@ -0,0 +1,227 @@
...
+{
+    const auto handles = m_session->get_handles();
+
+    // Set up DLPack tensor properties for a 4x4 float32 matrix
+    XrPushTensorDlpackCreateInfoNV dlpackInfo{};
+    dlpackInfo.type = XR_TYPE_PUSH_TENSOR_DLPACK_CREATE_INFO_NV;
+    dlpackInfo.next = nullptr;
+    dlpackInfo.data.versionMajor = 1;
+    dlpackInfo.data.versionMinor = 0;
+    dlpackInfo.data.dtype = DLPACK_DTYPE_FLOAT32;
```
nv-jakob, line 116: Hmm yeah I probably need to change this in the interface for next version, the dtype is the below:
https://github.com/dmlc/dlpack/blob/main/include/dlpack/dlpack.h#L202-L215

jiwenc-nv: `dtype` is a uint32

`dlpack` has unit8 uint8 uint16, so it kind of fits perfectly?

nv-jakob: Yes, but I think that I should have just done `unit8`, `uint8`, `uint16`. Looking at `DLPACK_DTYPE_FLOAT32` it doesn't set the lane field for instance, I think.

jiwenc-nv: yes, we are missing the lane field. here.

PR #69, opened 2026-01-29
Enable Tensor API from CloudXR runtime 6.1.0-pid4

jiwenc-nv:


deps/cloudxr/.env.default
```diff
@@ -21,3 +21,9 @@ CXR_HOST_VOLUME_PATH=$HOME/.cloudxr
 # If False: The default OpenXR hand tracking data from XR HMD will be visible
 #          via the xrio API.
 NV_CXR_ENABLE_PUSH_DEVICES=1
+
+# Whether to enable the Generic Tensor Collection interface for pushing tensor
+# data into the runtime.
+# If true: the tensor data will be visible through the xrio API.
+# If false: the tensor data will not be visible through the xrio API.
+NV_CXR_ENABLE_TENSOR_DATA=1
```
nv-jakob: Nit: Newline

jiwenc-nv: good catch... we should have automation for this.

PR #70, opened 2026-01-29
Add publish-wheel job to GitHub Actions for tagging releases

aristarkhovNV:
Python whell is uploaded to artifactory for every pushed git tag

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -132,10 +132,12 @@ jobs:
 
   publish-wheel:
     runs-on: [self-hosted, linux]
-    needs: test-cloudxr
+    needs:
+      - build-ubuntu
+      - test-cloudxr
     environment: dev
-    # Publish only for tag pushes, and only after CloudXR tests succeed.
-    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') && needs.test-cloudxr.result == 'success' }}
+    # Publish only for tag pushes, and only after build and CloudXR tests succeed.
+    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') && needs.build-ubuntu.result == 'success' && needs.test-cloudxr.result == 'success' }}
 
     steps:
     - name: Set up Python


.github/workflows/build-ubuntu.yml
```diff
@@ -127,3 +129,63 @@ jobs:
 
         # Run tests with CloudXR runtime
         ./scripts/run_tests_with_cloudxr.sh
+
+  publish-wheel:
+    runs-on: [self-hosted, linux]
+    needs: test-cloudxr
+    environment: dev
+    # Publish only for tag pushes, and only after CloudXR tests succeed.
+    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') && needs.test-cloudxr.result == 'success' }}
```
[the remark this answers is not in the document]
aristarkhovNV: we specifically want success

.github/workflows/build-ubuntu.yml
```diff
@@ -127,3 +129,65 @@ jobs:
...
         ./scripts/run_tests_with_cloudxr.sh
+
+  publish-wheel:
+    runs-on: [self-hosted, linux]
+    needs:
+      - build-ubuntu
+      - test-cloudxr
+    environment: dev
+    # Publish only for tag pushes, and only after build and CloudXR tests succeed.
+    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') && needs.build-ubuntu.result == 'success' && needs.test-cloudxr.result == 'success' }}
```
jiwenc-nv, line 140: Maybe we can do it in a follow up commit: but can we also publish for successful pull requests as well?

This can be very useful for internal NV (but non-teleopcore) developers to quickly adopt a version that's still under development.

.github/workflows/build-ubuntu.yml
```diff
@@ -127,3 +129,65 @@ jobs:
...
+
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.11'
```
jiwenc-nv, line 146: I think we need to support more python versions. 3.10 ~ 3.13

.github/workflows/build-ubuntu.yml
aristarkhovNV: Sure we can do that, but to avoid arbitrary overwrites we'll need an auto-version mechanism

.github/workflows/build-ubuntu.yml
aristarkhovNV: We'll need a multi-python build producing a wheel per python version. Lets figure out how that can be done in a follow up change.

PR #73, opened 2026-01-30
Implement SchemaPusher & Tracker on top of Generic Tensor Device

jiwenc-nv:


examples/schemaio/pedal_printer.cpp
```diff
@@ -0,0 +1,99 @@
...
+        return 1;
+    }
+
+    std::cout << "  OpenXR session created" << std::endl;
+
+    // Step 3: Create DeviceIOSession with the tracker
+    std::cout << "[Step 3] Creating DeviceIOSession..." << std::endl;
+
+    std::unique_ptr<core::DeviceIOSession> session;
+    try
```
aristarkhovNV: surround the entire main() with try-catch, in c++ anything can throw

examples/schemaio/pedal_printer.cpp
```diff
@@ -0,0 +1,99 @@
...
+    std::cout << "Sample " << sample_count;
+
+    std::cout << std::fixed << std::setprecision(3) << " [left=" << data.left_pedal << ", right=" << data.right_pedal
+              << ", rudder=" << data.rudder << "]";
+
+    std::cout << std::endl;
+}
+
+int main()
+{
```
aristarkhovNV, line 42:
```suggestion
try
{
```

examples/schemaio/pedal_printer.cpp
```diff
@@ -0,0 +1,99 @@
...
+            print_pedal_data(tracker->get_data(*session), new_count);
+            received_count = new_count;
+        }
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    }
+
+    std::cout << "\nTotal samples received: " << received_count << std::endl;
+    return 0;
+}
```
aristarkhovNV, line 98:
```suggestion
}
catch (const std::exception& e)
{
    std::cerr << "Error: " << e.what() << std::endl;
    return 1;
}
```

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,131 @@
...
+ *     }
+ * };
+ * @endcode
+ */
+class SchemaPusherBase
+{
+public:
+    //! Required OpenXR extensions for pushing tensor data
+    static constexpr const char* REQUIRED_EXTENSIONS[] = { "XR_NVX1_push_tensor", "XR_NVX1_tensor_data" };
+    static constexpr size_t REQUIRED_EXTENSIONS_COUNT = 2;
```
aristarkhovNV: Try to use `std::size(REQUIRED_EXTENSIONS)` instead

src/core/pusherio/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+
+# pusherio only depends on oxr_utils (header-only) and openxr_extensions (for tensor push/read APIs)
+# It does NOT link to the OpenXR loader - callers create the session and pass handles in
+if(NOT TARGET oxr::oxr_utils)
+    message(FATAL_ERROR "oxr_utils target not found. Make sure oxr_utils is built first.")
+endif()
+
+if(NOT TARGET Teleop::openxr_extensions)
+    message(FATAL_ERROR "openxr_extensions target not found. Make sure OpenXR extensions are available.")
+endif()
```
aristarkhovNV: no need to check, cmake will complain if you attempt to link with non-existing target alias

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,131 @@
...
+    bool push_buffer(const uint8_t* buffer, size_t size);
+
+    /*!
+     * @brief Access the configuration for subclass use.
+     */
+    const SchemaPusherConfig& config() const;
+
+private:
+    class Impl;
+    std::unique_ptr<Impl> m_impl;
```
aristarkhovNV: I don't see a big need for pimpl here

src/core/pusherio/cpp/schema_pusher.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+        // Create padded buffer to match declared tensor size
+        // The DLPack tensor is declared as uint8[max_flatbuffer_size], so we need to pad
+        std::vector<uint8_t> padded_buffer(m_config.max_flatbuffer_size, 0);
+        std::memcpy(padded_buffer.data(), buffer, size);
+
+        // Get current time for timestamps
+        auto now = std::chrono::steady_clock::now();
+        auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(now.time_since_epoch()).count();
+        XrTime xr_time = static_cast<XrTime>(ns);
```
aristarkhovNV: we must use OpenXR time conversion extensions to obtain XrTime from system time

src/core/pusherio/cpp/schema_pusher.cpp
```diff
@@ -0,0 +1,235 @@
...
+        // Create the tensor collection
+        create_tensor_collection();
+
+        std::cout << "SchemaPusherBase initialized for collection: " << m_config.collection_id << std::endl;
+    }
+
+    ~Impl()
+    {
+        // Destroy the push tensor collection
+        if (m_push_tensor != XR_NULL_HANDLE && m_destroy_fn != nullptr)
```
aristarkhovNV: must never have null destroy function, the object would fail to initialize. No need to check but may add an assert.

Same goes for `m_push_tensor` - if the creation fails the destructor will not be called.

src/core/pusherio/cpp/schema_pusher.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+        m_tensor_collection_id = createResult.tensorCollectionId;
+    }
+
+    OpenXRSessionHandles m_handles;
+    SchemaPusherConfig m_config;
+
+    // Push tensor collection handle
+    XrPushTensorCollectionNV m_push_tensor{ XR_NULL_HANDLE };
+    XrTensorCollectionIDNV m_tensor_collection_id{ 0 };
```
aristarkhovNV: m_tensor_collection_id is unused

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,131 @@
...
+ * creation and sample pushing logic. Subclasses provide typed `push()` methods that serialize
+ * their specific FlatBuffer types and call `push_buffer()`.
+ *
+ * The caller is responsible for creating the OpenXR session with the required extensions
+ * (XR_NVX1_PUSH_TENSOR_EXTENSION_NAME, XR_NVX1_TENSOR_DATA_EXTENSION_NAME) and passing
+ * the handles to this class.
+ *
+ * Example subclass:
+ * @code
+ * class HeadPosePusher : public SchemaPusherBase {
```
aristarkhovNV: Subclass oriented API is considered harmful these days. Consider using composition instead:

```
class HeadPosePusher
{
private:
    SchemaPusherBase m_pusher_base; // up to implementer, may also be externally provided
}
```

src/core/pusherio/cpp/schema_pusher.cpp
```diff
@@ -0,0 +1,235 @@
...
+            return false;
+        }
+
+        // Create padded buffer to match declared tensor size
+        // The DLPack tensor is declared as uint8[max_flatbuffer_size], so we need to pad
+        std::vector<uint8_t> padded_buffer(m_config.max_flatbuffer_size, 0);
+        std::memcpy(padded_buffer.data(), buffer, size);
+
+        // Get current time for timestamps
+        auto now = std::chrono::steady_clock::now();
```
aristarkhovNV: provide timestamp externally so this class could be more derministically unit tested.

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+// SchemaTracker Implementation
+// =============================================================================
+
+SchemaTracker::SchemaTracker(SchemaTrackerConfig config) : m_config(std::move(config))
+{
+}
+
+std::vector<std::string> SchemaTracker::get_required_extensions() const
+{
+    return { REQUIRED_EXTENSIONS[0] };
```
aristarkhovNV: don't hardcode for specific array size

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+        // Create tensor list
+        create_tensor_list();
+
+        std::cout << "SchemaTracker initialized, looking for collection: " << m_config.collection_id << std::endl;
+    }
+
+    ~ImplData()
+    {
+        // Destroy tensor list
+        if (m_tensor_list != XR_NULL_HANDLE && m_destroy_list_fn != nullptr)
```
aristarkhovNV: No need to null check. Neither must be null if the ctor succeeded, otherwise dtor won't be called.

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
```diff
@@ -0,0 +1,217 @@
...
+        Impl(const Impl&) = delete;
+        Impl& operator=(const Impl&) = delete;
+        Impl(Impl&&) = delete;
+        Impl& operator=(Impl&&) = delete;
+
+        /*!
+         * @brief Check if the target tensor collection has been discovered.
+         * @return true if connected to the target collection, false otherwise.
+         */
+        bool is_connected() const;
```
aristarkhovNV: seem unused, remove

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+            throw std::runtime_error("Failed to create tensor list, result=" + std::to_string(result));
+        }
+    }
+
+    void poll_for_updates()
+    {
+        // Check if tensor list needs update
+        uint64_t latest_generation = 0;
+        XrResult result = m_get_latest_gen_fn(m_handles.session, &latest_generation);
+        if (result == XR_SUCCESS && latest_generation != m_cached_generation)
```
aristarkhovNV: throw if not XR_SUCCESS

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+        {
+            result = m_update_list_fn(m_tensor_list);
+            if (result == XR_SUCCESS)
```
aristarkhovNV: throw if unsuccessful, don't swallow errors

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    int find_target_collection()
+    {
+        // Get list properties
+        XrSystemTensorListPropertiesNV listProps{ XR_TYPE_SYSTEM_TENSOR_LIST_PROPERTIES_NV };
+        listProps.next = nullptr;
+
+        XrResult result = m_get_list_props_fn(m_tensor_list, &listProps);
+        if (result != XR_SUCCESS)
```
aristarkhovNV, line 166: throw

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+                m_target_collection_index = find_target_collection();
+            }
+        }
+    }
+
+    int find_target_collection()
+    {
+        // Get list properties
+        XrSystemTensorListPropertiesNV listProps{ XR_TYPE_SYSTEM_TENSOR_LIST_PROPERTIES_NV };
+        listProps.next = nullptr;
```
aristarkhovNV: not needed, it is default initialized to null by {} above

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+        if (listProps.tensorCollectionCount == 0)
+        {
+            return -1; // No collections available yet
+        }
+
+        // Search for matching collection
+        for (uint32_t i = 0; i < listProps.tensorCollectionCount; ++i)
+        {
+            XrTensorCollectionPropertiesNV collProps{ XR_TYPE_TENSOR_COLLECTION_PROPERTIES_NV };
+            collProps.next = nullptr;
```
aristarkhovNV: not needed

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+            result = m_get_coll_props_fn(m_tensor_list, i, &collProps);
+            if (result != XR_SUCCESS)
```
aristarkhovNV: throw

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+    {
+        if (m_target_collection_index < 0)
+        {
+            return false;
+        }
+
+        // Prepare retrieval info
+        XrTensorDataRetrievalInfoNV retrievalInfo{ XR_TYPE_TENSOR_DATA_RETRIEVAL_INFO_NV };
+        retrievalInfo.next = nullptr;
+        retrievalInfo.tensorCollectionIndex = static_cast<uint32_t>(m_target_collection_index);
```
aristarkhovNV: may store m_target_collection_index as uint32_t to avoid casts, use uint_max when unknown

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+        tensorData.bufferCapacity = static_cast<uint32_t>(dataBuffer.size());
+        tensorData.writtenSampleCount = 0;
+
+        // Retrieve samples
+        XrResult result = m_get_data_fn(m_tensor_list, &retrievalInfo, &tensorData);
+        if (result != XR_SUCCESS)
+        {
+            if (result == XR_ERROR_TENSOR_LOST_NV)
+            {
+                std::cerr << "Tensor collection lost, will re-discover..." << std::endl;
```
aristarkhovNV: this is a normal situation, not an error. Don't spam the log and rediscover right away

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+        tensorData.metadataCapacity = 1;
+        tensorData.buffer = dataBuffer.data();
+        tensorData.bufferCapacity = static_cast<uint32_t>(dataBuffer.size());
+        tensorData.writtenSampleCount = 0;
+
+        // Retrieve samples
+        XrResult result = m_get_data_fn(m_tensor_list, &retrievalInfo, &tensorData);
+        if (result != XR_SUCCESS)
+        {
+            if (result == XR_ERROR_TENSOR_LOST_NV)
```
aristarkhovNV: else: throw

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+    // Target collection info
+    int m_target_collection_index{ -1 };
+    uint32_t m_sample_batch_stride{ 0 };
+    uint32_t m_sample_size{ 0 };
+
+    // Generation tracking
+    uint64_t m_cached_generation{ 0 };
+
+    // Sample tracking
+    int64_t m_last_sample_index{ -1 };
```
aristarkhovNV: btw -1 is a valid index

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    size_t get_read_count() const
+    {
+        return m_read_count;
+    }
+
+    bool read_buffer(std::vector<uint8_t>& buffer)
+    {
+        // Poll for tensor list updates
+        poll_for_updates();
```
aristarkhovNV, line 61: Poll only when collection index is unknown or on demand when the tensor collection is lost. No need to poll on every read

examples/schemaio/README.md
```diff
@@ -0,0 +1,69 @@
...
+- Takes externally-provided OpenXR session handles
+- Creates a tensor collection with the configured identifier
+- Provides `push_buffer()` for subclasses to push serialized FlatBuffer data
+- Subclasses implement typed `push()` methods for specific schema types
+
+**SchemaTracker** (`deviceio` library) - Base class for reading FlatBuffer schema data:
+- Integrates with the `ITracker` interface for use with `DeviceIOSession`
+- Discovers tensor collections by identifier
+- Provides `read_buffer()` for subclasses to read raw sample data
+- Subclasses implement `update()` to deserialize and store typed data
```
aristarkhovNV: Use composition instead of subclasses. This should also allow pure python trackers in the future

src/core/deviceio/cpp/generic_3axis_pedal_tracker.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include <vector>
+
+namespace core
+{
+
+// ============================================================================
+// Generic3AxisPedalTracker::Impl
+// ============================================================================
+
+class Generic3AxisPedalTracker::Impl : public SchemaTracker::Impl
```
aristarkhovNV: isn't SchemaTracker::Impl supposed to be private? And also this whole inheritance & pimpl hell feels wrong.

examples/schemaio/pedal_printer.cpp
jiwenc-nv: I'm not sure we really want to promote the "try - catch - all" behavior? esp in a sample like this.

examples/schemaio/pedal_printer.cpp
jiwenc-nv: let's discuss offline. I'm still a bit upset of the usage of exception logics without knowing what's throwing...

examples/schemaio/pedal_printer.cpp
jiwenc-nv: ditto.

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: done.

src/core/pusherio/cpp/CMakeLists.txt
jiwenc-nv: done.

src/core/pusherio/cpp/schema_pusher.cpp
jiwenc-nv: Done. Refactoring the XrTime conversion logic into a separate util: `oxr_time.h`

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: Grilled the AI a bit, it seems they want to hide the OpenXR headers. I don't have strong opinion on that. Removed.

src/core/schema/fbs/pedals.fbs
```diff
@@ -0,0 +1,21 @@
+include "point.fbs";
```
shaosu-nvidia, line 1: rebase main for this file change?

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,131 @@
...
+    std::string collection_id;
+
+    //! Maximum serialized FlatBuffer message size in bytes.
+    //! The tensor collection is created with this fixed buffer size.
+    //! Serialized messages larger than this will be rejected.
+    size_t max_flatbuffer_size;
+
+    //! Tensor name within the collection (e.g., "head_pose").
+    //! This identifies the specific tensor holding the serialized data.
+    std::string tensor_identifier;
```
shaosu-nvidia, line 38: just want to make sure I understand this correctly. A collection can have multiple tensor_identifier correct? this schema_pusher only has one tensor per collection that's why we have 2 ids here.

examples/schemaio/pedal_printer.cpp
jiwenc-nv: More discussions today: we aligned on the following statement:

```markdown
1. TeleopCore SDK (Python & C++) may throw exceptions, but when unrecoverable errors are happening.
2. When this happens, users of TeleopCore SDK should be responsible for catching it and respond as they see fit (like: exit program, print out log, bring up error dialog).
3. We should update all samples to handle those exceptions at top-level main to be consistent with this message.
```

examples/schemaio/pedal_printer.cpp
jiwenc-nv: updated and included the try / catch syntax. but we should update the rest of the code base to be consistent

src/core/schema/fbs/pedals.fbs
jiwenc-nv: yep.

src/core/pusherio/cpp/schema_pusher.cpp
jiwenc-nv: done

src/core/pusherio/cpp/schema_pusher.cpp
jiwenc-nv: done.

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: `REQUIRED_EXTENSIONS` seems to be completely redundant, I can just return the vector here.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
jiwenc-nv: done. also removed get_read_count

src/core/pusherio/cpp/schema_pusher.cpp
jiwenc-nv: moved to a util class `XrTimeConverter`, which can be mocked out if needed.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: uint_max seems too verbose. switch to use std::optional<uint32> instead.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: use std::optional instead

jiwenc-nv: done.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

examples/schemaio/README.md
jiwenc-nv: update the doc.

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: By "2 ids", do you mean `collection_id` and `tensor_identifier`?

So the "collection" only has one "tensor", then both the "collection" and the "tensor" in it has an ID. But since we are doing a 1:1 mapping, we could just simplify it and use the same id string.

src/core/deviceio/cpp/generic_3axis_pedal_tracker.cpp
jiwenc-nv: hmm, that's just the Impl factory thing used to separate OpenXR headers from creeping out. I don't have strong opinion on whether to remove it or not. But let's just keep it as-is, so it's consistent with the existing pattern.

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: consolidated them into one.

examples/schemaio/pedal_printer.cpp
aristarkhovNV: nested try-catch no longer needed

examples/schemaio/pedal_printer.cpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+        // Update session (this calls update on all trackers)
+        if (!session->update())
+        {
+            std::cerr << "Update failed" << std::endl;
+            break;
+        }
+
+        // Print current data if available
+        const auto& data = tracker->get_data(*session);
```
aristarkhovNV, line 84: get data in a loop while it is available, otherwise the data retrieval is rate limited and may not keep up with the pusher if clocks drift.

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
```diff
@@ -0,0 +1,106 @@
...
+        struct timespec ts;
+        clock_gettime(CLOCK_MONOTONIC, &ts);
+
+        if (pfn_convert_timespec_)
+        {
+            pfn_convert_timespec_(handles_.instance, &ts, &time);
+            return true;
+        }
+#endif
+        std::cerr << "Cannot get time - time conversion not available" << std::endl;
```
aristarkhovNV: make this compile time error

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
```diff
@@ -0,0 +1,106 @@
...
+     */
+    bool get_current_time(XrTime& time) const
+    {
+#if defined(XR_USE_PLATFORM_WIN32)
+        LARGE_INTEGER counter;
+        QueryPerformanceCounter(&counter);
+
+        if (pfn_convert_win32_)
+        {
+            pfn_convert_win32_(handles_.instance, &counter, &time);
```
aristarkhovNV: check result & throw

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
```diff
@@ -0,0 +1,106 @@
...
+            return true;
+        }
+#elif defined(XR_USE_TIMESPEC)
+        struct timespec ts;
+        clock_gettime(CLOCK_MONOTONIC, &ts);
+
+        if (pfn_convert_timespec_)
+        {
+            pfn_convert_timespec_(handles_.instance, &ts, &time);
```
aristarkhovNV: check result & throw

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
```diff
@@ -0,0 +1,106 @@
...
+     * @param[out] time The current XrTime value.
+     * @return true on success, false if time conversion failed.
+     */
+    bool get_current_time(XrTime& time) const
+    {
+#if defined(XR_USE_PLATFORM_WIN32)
+        LARGE_INTEGER counter;
+        QueryPerformanceCounter(&counter);
+
+        if (pfn_convert_win32_)
```
aristarkhovNV: redundant check, must always have the pointer, otherwise ctor would fail

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
```diff
@@ -0,0 +1,106 @@
...
+        {
+            pfn_convert_win32_(handles_.instance, &counter, &time);
+            return true;
+        }
+#elif defined(XR_USE_TIMESPEC)
+        struct timespec ts;
+        clock_gettime(CLOCK_MONOTONIC, &ts);
+
+        if (pfn_convert_timespec_)
```
aristarkhovNV: redundant check, must always have the pointer, otherwise ctor would fail

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
```diff
@@ -0,0 +1,106 @@
...
+        QueryPerformanceCounter(&counter);
+
+        if (pfn_convert_win32_)
+        {
+            pfn_convert_win32_(handles_.instance, &counter, &time);
+            return true;
+        }
+#elif defined(XR_USE_TIMESPEC)
+        struct timespec ts;
+        clock_gettime(CLOCK_MONOTONIC, &ts);
```
aristarkhovNV: check return value, throw if failed, get the reason from `errno`

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
```diff
@@ -0,0 +1,106 @@
...
+    /*!
+     * @brief Gets the current time as XrTime using platform-specific conversion.
+     * @param[out] time The current XrTime value.
+     * @return true on success, false if time conversion failed.
+     */
+    bool get_current_time(XrTime& time) const
+    {
+#if defined(XR_USE_PLATFORM_WIN32)
+        LARGE_INTEGER counter;
+        QueryPerformanceCounter(&counter);
```
aristarkhovNV: check result & throw

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,137 @@
...
+     * @param buffer Pointer to serialized FlatBuffer data.
+     * @param size Size of the serialized data in bytes.
+     * @return true on success, false on failure.
+     */
+    bool push_buffer(const uint8_t* buffer, size_t size);
+
+    /*!
+     * @brief Returns the number of samples successfully pushed.
+     */
+    size_t get_push_count() const;
```
aristarkhovNV: this is not essential functionality for the pusher. Ther user may count themselves if needed.

src/core/pusherio/cpp/schema_pusher.cpp
```diff
@@ -0,0 +1,184 @@
...
+    tensorData.timestamp = xr_time;
+    tensorData.rawDeviceTimestamp = static_cast<uint64_t>(xr_time);
+    tensorData.buffer = padded_buffer.data();
+    tensorData.bufferSize = static_cast<uint32_t>(m_config.max_flatbuffer_size);
+
+    // Push the data
+    XrResult result = m_push_fn(m_push_tensor, &tensorData);
+    if (result != XR_SUCCESS)
+    {
+        std::cerr << "Failed to push tensor data, result=" << result << std::endl;
```
aristarkhovNV: throw?

examples/schemaio/pedal_pusher.cpp
```diff
@@ -0,0 +1,136 @@
...
+        return 1;
+    }
+
+    std::cout << "  OpenXR session created" << std::endl;
+
+    // Step 2: Create the pusher with the session handles
+    std::cout << "[Step 2] Creating Generic3AxisPedalPusher..." << std::endl;
+
+    std::unique_ptr<Generic3AxisPedalPusher> pusher;
+    try
```
aristarkhovNV: nested try-catch no longer necessary

examples/schemaio/pedal_pusher.cpp
```diff
@@ -0,0 +1,136 @@
...
+        pedal_output.is_valid = true;
+        pedal_output.left_pedal = left_pedal;
+        pedal_output.right_pedal = right_pedal;
+        pedal_output.rudder = rudder;
+
+        auto now = std::chrono::steady_clock::now();
+        auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(now.time_since_epoch()).count();
+        pedal_output.timestamp = std::make_shared<core::Timestamp>(ns, ns);
+
+        if (pusher->push(pedal_output))
```
aristarkhovNV: failed push should throw, or otherwise we need a clearly defined failure semantics that would explicitly outline repeated push expectations.

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
aristarkhovNV: still here

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+    /*!
+     * @brief Get required OpenXR extensions (includes tensor data extension).
+     */
+    std::vector<std::string> get_required_extensions() const override;
+
+    // Subclasses must implement these from ITracker:
+    // virtual std::string_view get_name() const = 0;
+    // virtual std::string_view get_schema_name() const = 0;
+    // virtual std::string_view get_schema_text() const = 0;
```
aristarkhovNV: this will be compiler enforced, no need to comment

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
```diff
@@ -0,0 +1,184 @@
...
+        bool read_buffer(std::vector<uint8_t>& buffer);
+
+        /*!
+         * @brief Access the configuration for subclass use.
+         */
+        const SchemaConfig& config() const;
+
+    private:
+        class ImplData;
+        std::unique_ptr<ImplData> m_data;
```
aristarkhovNV: can we get rid of this PIMPL too?

src/core/deviceio/cpp/CMakeLists.txt
```diff
@@ -16,16 +16,25 @@ if(NOT TARGET oxr::oxr_utils)
     message(FATAL_ERROR "oxr_utils target not found. Make sure oxr_utils is built first.")
 endif()
 
+# Check for openxr_extensions (needed for SchemaTracker tensor data APIs)
+if(NOT TARGET Teleop::openxr_extensions)
+    message(FATAL_ERROR "openxr_extensions target not found. Make sure OpenXR extensions are available.")
+endif()
```
aristarkhovNV: we never build deviceio standalone and these checks are not needed

src/core/deviceio/cpp/deviceio_session.cpp
```diff
@@ -18,28 +17,10 @@ namespace core
 
 DeviceIOSession::DeviceIOSession(const std::vector<std::shared_ptr<ITracker>>& trackers,
                                  const OpenXRSessionHandles& handles)
-    : handles_(handles)
+    : handles_(handles), time_converter_(handles)
 {
-    // Get time conversion function using the provided xrGetInstanceProcAddr
     assert(handles_.xrGetInstanceProcAddr && "xrGetInstanceProcAddr cannot be null");
```
aristarkhovNV: this class no longer cares about missing xrGetInstanceProcAddr

src/core/deviceio/cpp/deviceio_session.cpp
aristarkhovNV: these should now be retrieved via time converter

src/core/pusherio/cpp/schema_pusher.cpp
```diff
@@ -0,0 +1,184 @@
...
+    // Create padded buffer to match declared tensor size
+    // The DLPack tensor is declared as uint8[max_flatbuffer_size], so we need to pad
+    std::vector<uint8_t> padded_buffer(m_config.max_flatbuffer_size, 0);
+    std::memcpy(padded_buffer.data(), buffer, size);
+
+    // Get current time for timestamps
+    XrTime xr_time;
+    if (!m_time_converter.get_current_time(xr_time))
+    {
+        std::cerr << "ERROR: Failed to get current XrTime. Skipping push." << std::endl;
```
aristarkhovNV: throw?

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * Includes platform-specific time conversion extension.
+     */
+    static std::vector<std::string> get_required_extensions()
+    {
+#if defined(XR_USE_PLATFORM_WIN32)
+        return { "XR_NVX1_push_tensor", "XR_NVX1_tensor_data", "XR_KHR_win32_convert_performance_counter_time" };
+#elif defined(XR_USE_TIMESPEC)
+        return { "XR_NVX1_push_tensor", "XR_NVX1_tensor_data", "XR_KHR_convert_timespec_time" };
+#else
```
aristarkhovNV: time extensions are now implementation detail of the time converter and should be retrieved from it

src/core/pusherio/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+        # OpenXR extensions for push tensor and tensor data APIs
+        Teleop::openxr_extensions
+)
+
+# Platform-specific definitions
+if(WIN32)
+    target_compile_definitions(pusherio PUBLIC XR_USE_PLATFORM_WIN32 NOMINMAX)
+else()
+    target_compile_definitions(pusherio PUBLIC XR_USE_TIMESPEC)
+endif()
```
aristarkhovNV: these should now come from oxr_utils

examples/schemaio/common_utils.hpp
```diff
@@ -0,0 +1,20 @@
...
+{
+
+//! Maximum size for serialized Generic3AxisPedalOutput FlatBuffer.
+static constexpr size_t MAX_FLATBUFFER_SIZE = 256;
+
+//! Maximum total samples to push/receive before exiting.
+static constexpr int MAX_SAMPLES = 100;
+
+//! Collection ID shared between pusher and reader.
+static constexpr const char* COLLECTION_ID = "schemaio_example";
```
shaosu-nvidia, line 18: is this example for general purpose or pedal specifically?

jiwenc-nv: it's a general purpose sample. pedal was randomly picked as an example.

examples/schemaio/pedal_printer.cpp
jiwenc-nv: done.

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
jiwenc-nv: done.

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
jiwenc-nv: done.

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
jiwenc-nv: done.

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
jiwenc-nv: done.

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: done. Yeah, there are just endless YAGNI from...

src/core/oxr_utils/cpp/inc/oxr_utils/oxr_time.hpp
jiwenc-nv: done.

src/core/pusherio/cpp/schema_pusher.cpp
jiwenc-nv: yeah, why not.

examples/schemaio/pedal_pusher.cpp
jiwenc-nv: done. catch all for main.

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
jiwenc-nv: done.

examples/schemaio/pedal_pusher.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
jiwenc-nv: removed:

so, SchemaTracker should not a subclass of ITracker. rather a simple util class.

Then Generic3AxisPedalTracker should be a subclass of ITracker directly.

src/core/deviceio/cpp/CMakeLists.txt
jiwenc-nv: removed...

src/core/deviceio/cpp/deviceio_session.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/deviceio_session.cpp
jiwenc-nv: right. move to XrTimeConverter::get_required_extensions

src/core/pusherio/cpp/schema_pusher.cpp
jiwenc-nv: yeah, changed `get_current_time` to return XrTime directly, but will throw.

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: done.

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
aristarkhovNV: yep sounds good!

examples/schemaio/pedal_pusher.cpp
```diff
@@ -0,0 +1,131 @@
...
+    void push(const core::Generic3AxisPedalOutputT& data)
+    {
+        flatbuffers::FlatBufferBuilder builder(m_pusher.config().max_flatbuffer_size);
+        auto offset = core::Generic3AxisPedalOutput::Pack(builder, &data);
+        builder.Finish(offset);
+
+        if (!m_pusher.push_buffer(builder.GetBufferPointer(), builder.GetSize()))
+        {
+            throw std::runtime_error("Generic3AxisPedalPusher: Failed to push data");
+        }
```
aristarkhovNV: maybe `m_pusher.push_buffer` itself could throw instead

src/core/pusherio/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20)
+
+# PusherIO library (static for distribution)
+# Note: SchemaReaderBase has been moved to deviceio as SchemaTracker
```
aristarkhovNV: AI note can be removed

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
```diff
@@ -0,0 +1,201 @@
...
+ *
+ *     private:
+ *         SchemaTracker::Impl m_schema_reader;
+ *         std::vector<uint8_t> buffer_;
+ *         LocomotionCommandT data_;
+ *     };
+ * };
+ * @endcode
+ */
+class SchemaTracker
```
aristarkhovNV, line 117: looks like `SchemaTracker` class itself is now unused except for the static `SchemaTracker::get_required_extensions`

src/core/pusherio/cpp/CMakeLists.txt
jiwenc-nv: done.

examples/schemaio/pedal_pusher.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
jiwenc-nv: clean up all nested classes in it.

examples/schemaio/pedal_printer.cpp
aristarkhovNV: leftover try-catch

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,156 @@
...
+    /*!
+     * @brief Access the configuration.
+     */
+    const SchemaPusherConfig& config() const;
+
+private:
+    void initialize_push_tensor_functions();
+    void create_tensor_collection();
+
+    OpenXRSessionHandles m_handles;
```
aristarkhovNV: handles only used during initialization, does not need to be stored in the class

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,156 @@
...
+    XrPushTensorCollectionNV m_push_tensor{ XR_NULL_HANDLE };
+    XrTensorCollectionIDNV m_tensor_collection_id{ 0 };
+
+    // Extension function pointers
+    PFN_xrCreatePushTensorCollectionNV m_create_fn{ nullptr };
+    PFN_xrPushTensorCollectionDataNV m_push_fn{ nullptr };
+    PFN_xrDestroyPushTensorCollectionNV m_destroy_fn{ nullptr };
+
+    // Statistics
+    size_t m_push_count{ 0 };
```
aristarkhovNV: remove

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
```diff
@@ -0,0 +1,156 @@
...
+    void initialize_push_tensor_functions();
+    void create_tensor_collection();
+
+    OpenXRSessionHandles m_handles;
+    SchemaPusherConfig m_config;
+    XrTimeConverter m_time_converter;
+
+    // Push tensor collection handle
+    XrPushTensorCollectionNV m_push_tensor{ XR_NULL_HANDLE };
+    XrTensorCollectionIDNV m_tensor_collection_id{ 0 };
```
aristarkhovNV: `m_tensor_collection_id` initialized but unused

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+private:
+    void initialize_tensor_data_functions();
+    void create_tensor_list();
+    void poll_for_updates();
+    std::optional<uint32_t> find_target_collection();
+    bool read_next_sample(std::vector<uint8_t>& buffer);
+
+    struct Data;
+    std::unique_ptr<Data> m_data;
```
aristarkhovNV: can we get rid of separately allocated `struct Data` and place all its members directly under `SchemaTracker::` similarly how it is done for the SchemaPusher

src/core/deviceio/cpp/generic_3axis_pedal_tracker.cpp
```diff
@@ -0,0 +1,109 @@
...
+// ============================================================================
+
+class Generic3AxisPedalTracker::Impl : public ITrackerImpl
+{
+public:
+    Impl(const OpenXRSessionHandles& handles, SchemaTrackerConfig config) : m_schema_reader(handles, std::move(config))
+    {
+    }
+
+    bool update(XrTime /* time */) override
```
aristarkhovNV, line 26: for a future PR: @shaosu-nvidia and I discussed today that the tracker should use the provided timestamp to lookup (or interpolate between) the closest sample(s) retrieved from tensor API.

src/core/deviceio/cpp/schema_tracker.cpp
```diff
@@ -0,0 +1,316 @@
...
+// SchemaTracker::Data - Internal data for tensor reading
+// =============================================================================
+
+/*!
+ * @brief Internal data for SchemaTracker.
+ *
+ * Uses externally-provided OpenXR session handles for tensor read operations.
+ * This struct is used to hide OpenXR types from the header.
+ */
+struct SchemaTracker::Data
```
aristarkhovNV: put these directly in SchemaTracker

src/core/deviceio/cpp/generic_3axis_pedal_tracker.cpp
jiwenc-nv: I was wondering which component should be doing that. The tracker or the runtime.

If we try to be closer to how `xrLocateSpace` works, runtime is better suited to do the lookup/interpolation.

But if we following the current tensor API design, StreamTracker or each individual tracker can do that.

shaosu-nvidia: - if it happens in the runtime, it's going to be hard since the tensor API is carrying arbitrary/opaque data from different collection, so there is no mutual way for interpolation.

- And it seems the pedal tracker at the moment will take the first available data point from the tensor queue. Just want to confirm that if pedal is at 60Hz and robot control loop is at 30Hz, then we are shifting by 1 data sample every time we call update to get the state?

examples/schemaio/pedal_printer.cpp
jiwenc-nv: fixed.

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: right. it's only used by the constructor.

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: done

src/core/pusherio/cpp/inc/pusherio/schema_pusher.hpp
jiwenc-nv: removed.

src/core/deviceio/cpp/inc/deviceio/schema_tracker.hpp
jiwenc-nv: yes. the only drawback is now that `schema_tracker.hpp` will need to expose the `XR_NVX1_tensor_data.h` header. I don't hold strong view against it.

jiwenc-nv: I meant I got rid of that.

src/core/deviceio/cpp/schema_tracker.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/generic_3axis_pedal_tracker.cpp
jiwenc-nv: yeah, we shall have more discussions around that.

PR #80, opened 2026-02-02
Streamline CloudXR Docker container setup

jiwenc-nv:
The `run_cloudxr.sh` script will automatically:
- Download the CloudXR Web SDK from NGC (if not already present)
- Build the necessary Docker containers (wss-proxy, web-app)
- Start all CloudXR services

### DIFF
--- a/deps/cloudxr/Dockerfile.web-app
+++ b/deps/cloudxr/Dockerfile.web-app
@@ -1,0 +1,147 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# CloudXR Web App - Multi-stage build
+# Stage 1: Build the web application
+# Stage 2: Serve with nginx (HTTP on 8080, HTTPS on 8443)
+
+#############################################
+# Stage 1: Build the web application
+#############################################
+FROM node:24-slim AS builder
+
+WORKDIR /app
+
+# Install webpack globally for convenience
+RUN npm install -g webpack webpack-cli
+
+# Path to the CloudXR Web SDK (relative to build context)
+ARG CXR_WEB_SDK_VERSION=${CXR_WEB_SDK_VERSION}
+ENV SDK_PATH=cloudxr-web-sdk-${CXR_WEB_SDK_VERSION}
+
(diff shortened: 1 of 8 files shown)


scripts/download_cloudxr_sdk.sh
```diff
@@ -0,0 +1,103 @@
...
+
+# Check if ngc CLI is available
+if ! command -v ngc &> /dev/null; then
+    echo -e "${RED}Error: NGC CLI not found. Please install it first.${NC}"
+    echo -e "Visit: https://ngc.nvidia.com/setup/installers/cli"
+    exit 1
+fi
+
+# Clean up any previous partial downloads
+rm -rf "$SDK_DOWNLOAD_DIR"
```
aristarkhovNV: this can be destructive without user knowing. Better fail if the directory exists

deps/cloudxr/Dockerfile.web-app
```diff
@@ -0,0 +1,147 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aristarkhovNV, line 1: Previously we were able to download existing containers, why now build ourselves? Also would be good if these docker files were served with CloudXR SDK, otherwise feels like it's a copy & paste of some existing docker files that we used to build the conteiners from before.

jiwenc-nv: No. Actually CloudXR didn't ship existing containers in the past. We were just building it under:
https://github.com/NVIDIA/IsaacTeleop/tree/main/examples/cxrjs

We need to keep that example alive for a couple more days (Pico is currently using it).

aristarkhovNV: do you plan to remove that example later? Can we temporarily take the docker files from the example to avoid duplicating here?

scripts/download_cloudxr_sdk.sh
jiwenc-nv: fixed :)

deps/cloudxr/Dockerfile.web-app
jiwenc-nv: Yes. The plan is to remove that example, but the docker files are indeed different (the new one builds Javascript stuff, while the older one doesn't).

PR #82, opened 2026-02-02
Allow Python v3.12

hougantc-nvda:


examples/teleop/python/pyproject.toml
```diff
@@ -5,7 +5,7 @@
 name = "teleop-examples"
 version = "0.1.0"
 description = "TeleopCore Retargeting Examples"
-requires-python = ">=3.11,<3.12"
+requires-python = ">=3.11,<3.13"
```
jiwenc-nv: nit: can you fix this one to be 3.10 as well?

`3.10`

Thanks.

PR #83, opened 2026-02-03
Use git commit count in versioning

aristarkhovNV:


cmake/IsaacTeleopVersion.cmake
```diff
@@ -0,0 +1,62 @@
...
+
+	string(REGEX MATCH "^([0-9]+)\\.([0-9]+)\\.([0-9]+)" _isaac_teleop_version_match "${_isaac_teleop_version_base}")
+	if(NOT _isaac_teleop_version_match)
+		message(FATAL_ERROR "Version must be in MAJOR.MINOR.PATCH format: '${_isaac_teleop_version_base}'")
+	endif()
+	set(_isaac_teleop_version_major "${CMAKE_MATCH_1}")
+	set(_isaac_teleop_version_minor "${CMAKE_MATCH_2}")
+	set(_isaac_teleop_version_patch "${CMAKE_MATCH_3}")
+	string(REGEX REPLACE "^([0-9]+)\\.([0-9]+)\\.([0-9]+)" "" _isaac_teleop_version_suffix "${_isaac_teleop_version_base}")
+	math(EXPR _isaac_teleop_version_patch "${_isaac_teleop_version_patch}+${_isaac_teleop_git_count}")
```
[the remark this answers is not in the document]
aristarkhovNV: math EXPR is not space sensitive

ISAAC_TELEOP_VERSION_BASE
jiwenc-nv, line 1: I suggest we need the file based as `VERSION` as that's more aligned with conventions.

cmake/IsaacTeleopVersion.cmake
```diff
@@ -0,0 +1,67 @@
+# Read base version from file and increment the patch by the number of git commits
+# since the last change to the version file.
```
jiwenc-nv, line 2: maybe the `VERSION` file can contain `MAJOY.MINOR`?

`PATCH` == $(git commits# since the last change to the version file)

then for release branch: it's simply `MAJOR.MINOR.PATCH`
then for main branch: it will be: `MAJOR.MINOR.dev{PATCH}`?

ISAAC_TELEOP_VERSION_BASE
aristarkhovNV: I wanted to emphasize this is a starting point for version counting, not the actual final version. We could go with:
1. VERSION_BASE
2. VERSION but contain "0.2.x" instead of a fully resolved one

aristarkhovNV: went with option 2

cmake/IsaacTeleopVersion.cmake
aristarkhovNV: will introduce `dev{PATCH}` for `main` in a follow up PR

VERSION
```diff
@@ -1 +1 @@
-0.2.0
+0.2.x
```
jiwenc-nv: good timing to bump into 0.3.x? up to you

PR #85, opened 2026-02-03
Boost up the build time with caching on .hunter

shaosu-nvidia:


.github/workflows/build-ubuntu.yml
```diff
@@ -32,6 +30,14 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y build-essential cmake libx11-dev clang-format-14 ccache
 
+    # Hunter cache - caches depthai dependencies (OpenSSL, CURL, etc.)
+    # This saves ~7 minutes on each build
+    - name: Cache Hunter packages
+      uses: actions/cache@v4
+      with:
+        path: ~/.hunter
+        key: hunter-${{ runner.os }}-depthai-v2.29.0
```
jiwenc-nv: does runner.os cover CPU arch as well? we probably want to key against that, so that arm will work smoothly.

.github/workflows/build-ubuntu.yml
```diff
@@ -32,6 +30,13 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y build-essential cmake libx11-dev clang-format-14 ccache
 
+    # Hunter cache - caches depthai dependencies (OpenSSL, CURL, etc.)
+    - name: Cache Hunter packages
+      uses: actions/cache@v4
+      with:
+        path: ~/.hunter
+        key: hunter-${{ runner.os }}-${{ runner.arch }}-depthai-v2.29.0
+
```
aristarkhovNV, line 41: We can factor out hunter setup in a separate cmake:

```suggestion
        key: hunter-${{ runner.os }}-${{ runner.arch }}"-${{ matrix.build_type }}-${{ hashFiles('cmake/HunterGateSetup.cmake') }}
```

CMakeLists.txt
aristarkhovNV: move this out to a new file to be hashed for the hunter cache key

.github/workflows/build-ubuntu.yml
shaosu-nvidia: Yeah I have added runner.arch as well

.github/workflows/build-ubuntu.yml
```diff
@@ -32,6 +30,15 @@ jobs:
...
         sudo apt-get install -y build-essential cmake libx11-dev clang-format-14 ccache
 
+    # Hunter cache - caches depthai dependencies (OpenSSL, CURL, etc.)
+    - name: Cache Hunter packages
+      uses: actions/cache@v4
+      with:
+        path: ~/.hunter
+        key: hunter-${{ runner.os }}-${{ runner.arch }}-depthai-v2.29.0-${{ hashFiles('cmake/SetupHunter.cmake') }}
+        restore-keys: |
+          hunter-${{ runner.os }}-${{ runner.arch }}-depthai-v2.29.0-
```
aristarkhovNV: we no longer need to explicitly spell out the "depthai-v2.29.0" here, is is already mentioned inside SetupHunter.cmake : `set(DEPTHAI_VERSION "v2.29.0")`

shaosu-nvidia: Good catch!

PR #86, opened 2026-02-03
Add data schema for Pico full body tracking

jiwenc-nv:


src/core/schema/fbs/full_body.fbs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// Fixed-size array of 24 body joint locations. This follows the joint format defined in:
+// See: https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_BD_body_tracking
+struct BodyJointsPico {
+    joints: [BodyJointPose:24];
+}
+
+// Full body pose data for PICO body tracking (XR_BD_body_tracking extension).
+table FullBodyPosePico {
```
nv-jakob, line 51: The names looks good, I was going say they should have `Pico` in the name.

src/core/schema_tests/python/test_full_body.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class TestBodyJointPicoEnum:
+    """Tests for BodyJointPico enum (joint index mapping for XR_BD_body_tracking)."""
+
+    def test_joint_count(self):
+        """Test that BodyJointPico has exactly 24 members."""
+        assert len(BodyJointPico) == 24
+
+    def test_joint_values(self):
```
aristarkhovNV, line 202: this tests nothing really, just double checking that the var has the value we assigned earlier

src/core/schema/python/schema_init.py
```diff
@@ -35,8 +37,52 @@
...
+    XR_BD_body_tracking extension. Use these to access specific joints
+    from BodyJointsPico by index.
+
+    Example:
+        body_pose = get_body_pose()  # Get from tracker
+        head = body_pose.joints[BodyJointPico.HEAD]
+        left_hand = body_pose.joints[BodyJointPico.LEFT_HAND]
+    """
+
+    PELVIS = 0
```
aristarkhovNV: is there a way to avoid duplicating enum values in py? By exposing them from c++ via bindings or something

src/core/schema_tests/python/test_full_body.py
jiwenc-nv: lol, I put it here, not AI...

was meant to make sure no one messes up with the ordering of things...

src/core/schema/python/schema_init.py
jiwenc-nv: right... Ideally we should put it in the fbs file. Will make some changes

src/core/schema/fbs/full_body.fbs
jiwenc-nv: I may want to change it to `PicoFullBodyPose`

src/core/schema/python/schema_init.py
jiwenc-nv: moved those enums into FBS

src/core/schema_tests/python/test_full_body.py
jiwenc-nv: moved those joint name enum into fbs and the tests are making sure the enum doesn't get to change.

PR #94, opened 2026-02-06
Adding support for passing external OpenXR handles to Teleop Session

rwiltz:
Adding support for passing external OpenXR handles to Teleop Session API, and querying the OpenXR extensions requirments based on the trackers used.

### DIFF
--- a/src/core/teleop_session_manager_tests/python/CMakeLists.txt
+++ b/src/core/teleop_session_manager_tests/python/CMakeLists.txt
@@ -22,7 +22,7 @@ foreach(test_file ${TEST_FILES})
     # Add the test
     add_test(
         NAME "teleop_session_manager_${test_name}"
-        COMMAND uv run --python ${TELEOPCORE_PYTHON_VERSION} --extra dev pytest -v --tb=short "${test_file}"
+        COMMAND uv run --python ${ISAAC_TELEOP_PYTHON_VERSION} --extra dev pytest -v --tb=short "${test_file}"
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
     
--- a/src/core/teleop_session_manager_tests/python/test_external_oxr_handles.py
+++ b/src/core/teleop_session_manager_tests/python/test_external_oxr_handles.py
@@ -56,11 +56,13 @@ def _make_empty_pipeline():
 
 @contextmanager
 def _mock_deviceio_and_oxr():
-    """Patch DeviceIOSession.run and OpenXRSession.create to avoid native calls.
+    """Patch DeviceIOSession.run and the OpenXRSession constructor to avoid native calls.
 
     Yields a namespace object with attributes:
-        - deviceio_run:  the mock replacing DeviceIOSession.run
-        - oxr_create:    the mock replacing OpenXRSession.create
+        - deviceio_run:      the mock replacing DeviceIOSession.run
+        - oxr_cls:           the mock replacing the OpenXRSession class
+        - mock_dio_session:  the fake DeviceIOSession instance
+        - mock_oxr_session:  the fake OpenXRSession instance
     """
     mock_dio_session = MagicMock()
     mock_dio_session.__enter__ = MagicMock(return_value=mock_dio_session)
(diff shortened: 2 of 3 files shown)


src/core/teleop_session_manager/python/config.py
```diff
@@ -70,11 +76,24 @@ class TeleopSessionConfig:
...
+            pipeline=pipeline,
+            oxr_handles=handles,  # Skip internal OpenXR session creation
+        )
     """
     app_name: str
     pipeline: Any
     trackers: List[Any] = field(default_factory=list)
     plugins: List[PluginConfig] = field(default_factory=list)
     verbose: bool = True
+    oxr_handles: Optional[Any] = None
```
nvddr: We have an openxr session handles type somewhere

rwiltz: ACK

src/core/teleop_session_manager/python/helpers.py
```diff
@@ -3,12 +3,77 @@
...
+
+    Returns:
+        Deduplicated list of tracker instances used by the pipeline.
+    """
+    from teleopcore.retargeting_engine.deviceio_source_nodes import IDeviceIOSource
+
+    leaf_nodes = pipeline.get_leaf_nodes()
+    sources = [node for node in leaf_nodes if isinstance(node, IDeviceIOSource)]
+
+    # Deduplicate by identity — two sources may share the same tracker.
```
nvddr: This is not true, trackers won't really be shared as every source owns its own tracker. The deduplication should be in get_required_extensions_from_pipeline, as extensions may be duplicated.

src/core/teleop_session_manager/python/helpers.py
```diff
@@ -3,12 +3,77 @@
...
-Helper utilities for creating input nodes and retargeters.
+Helper utilities for pipeline introspection and input node creation.
 """
 
+from typing import Any, List
+
 from teleopcore.retargeting_engine.deviceio_source_nodes import HandsSource, ControllersSource, HeadSource
 
 
+def get_trackers_from_pipeline(pipeline: Any) -> List[Any]:
```
nvddr: Do we really need this to be part of the public API? Can we just have the second one?

src/core/teleop_session_manager/python/helpers.py
```diff
@@ -3,12 +3,77 @@
...
+        tracker = source.get_tracker()
+        tracker_id = id(tracker)
+        if tracker_id not in seen:
+            seen.add(tracker_id)
+            trackers.append(tracker)
+
+    return trackers
+
+
+def get_required_extensions_from_pipeline(pipeline: Any) -> List[str]:
```
nvddr: get_required_oxr_extensions_from_pipeline, as this is only for external openxr sessions.

src/core/teleop_session_manager/python/teleop_session.py
```diff
@@ -224,13 +228,17 @@ def __enter__(self):
...
-        # Get OpenXR handles
-        handles = self.oxr_session.get_handles()
+        # Resolve OpenXR handles
+        if self.config.oxr_handles is not None:
+            # Use externally provided handles (e.g. from Kit's XR bridge).
+            # The caller owns the OpenXR session lifetime; we only consume handles.
+            handles = self.config.oxr_handles
+        else:
+            # Create our own OpenXR session (standalone mode)
+            self.oxr_session = self._exit_stack.enter_context(
```
aristarkhovNV: after retrieving the handles `self.oxr_session` member variable does not seem to be used for anything else and could be removed from the class.

src/core/teleop_session_manager/python/teleop_session.py
```diff
@@ -221,13 +230,17 @@ def __enter__(self):
...
-        # Create OpenXR session
-        self.oxr_session = self._exit_stack.enter_context(
-            oxr.OpenXRSession.create(self.config.app_name, required_extensions)
-        )
-        
-        # Get OpenXR handles
-        handles = self.oxr_session.get_handles()
+        # Resolve OpenXR handles
+        if self.config.oxr_handles is not None:
+            # Use externally provided handles (e.g. from Kit's XR bridge).
```
nvddr: Should we reference any NV proprietary things like Kit in this repo's documentation?

jiwenc-nv: I think it's OK to reference Kit. But we might want to call out the full name "Omniverse Kit" just to reduce confusion.

PR #97, opened 2026-02-07
Add FullBodyTrackerPico for PICO XR body tracking

jiwenc-nv:
- Implement full body tracker using XR_BD_body_tracking (24 joints)
- Add C++ tracker (FullBodyTrackerPico), Python bindings, and body joint constants
- Add test script examples/oxr/python/test_full_body_tracker.py

### DIFF
--- a/src/core/deviceio/cpp/full_body_tracker_pico.cpp
+++ b/src/core/deviceio/cpp/full_body_tracker_pico.cpp
@@ -9,9 +9,6 @@
 #include <cstring>
 #include <iostream>
 
-// XR_BD_body_tracking extension types are defined in <openxr/openxr.h>
-// which is included via oxr_utils/oxr_funcs.hpp
-
 namespace core
 {
 
@@ -81,13 +78,12 @@ FullBodyTrackerPicoImpl::FullBodyTrackerPicoImpl(const OpenXRSessionHandles& han
         throw std::runtime_error("OpenXR: failed to get system: " + std::to_string(result));
     }
 
-    // Get extension function pointers using the provided xrGetInstanceProcAddr
-    handles.xrGetInstanceProcAddr(
-        handles.instance, "xrCreateBodyTrackerBD", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_create_body_tracker_));
-    handles.xrGetInstanceProcAddr(
-        handles.instance, "xrDestroyBodyTrackerBD", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_destroy_body_tracker_));
-    handles.xrGetInstanceProcAddr(
-        handles.instance, "xrLocateBodyJointsBD", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_locate_body_joints_));
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrCreateBodyTrackerBD",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_create_body_tracker_));
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrDestroyBodyTrackerBD",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_destroy_body_tracker_));
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrLocateBodyJointsBD",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_locate_body_joints_));
 
     if (!pfn_create_body_tracker_ || !pfn_destroy_body_tracker_ || !pfn_locate_body_joints_)
     {
--- a/src/core/deviceio/cpp/hand_tracker.cpp
+++ b/src/core/deviceio/cpp/hand_tracker.cpp
@@ -50,18 +50,12 @@ HandTracker::Impl::Impl(const OpenXRSessionHandles& handles)
         }
     }
 
-    // Get extension function pointers using the provided xrGetInstanceProcAddr
-    handles.xrGetInstanceProcAddr(
-        handles.instance, "xrCreateHandTrackerEXT", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_create_hand_tracker_));
-    handles.xrGetInstanceProcAddr(
-        handles.instance, "xrDestroyHandTrackerEXT", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_destroy_hand_tracker_));
-    handles.xrGetInstanceProcAddr(
-        handles.instance, "xrLocateHandJointsEXT", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_locate_hand_joints_));
-
-    if (!pfn_create_hand_tracker_ || !pfn_destroy_hand_tracker_ || !pfn_locate_hand_joints_)
-    {
-        throw std::runtime_error("Failed to get hand tracking function pointers");
-    }
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrCreateHandTrackerEXT",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_create_hand_tracker_));
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrDestroyHandTrackerEXT",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_destroy_hand_tracker_));
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrLocateHandJointsEXT",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_locate_hand_joints_));
 
     // Create hand trackers
     XrHandTrackerCreateInfoEXT create_info{ XR_TYPE_HAND_TRACKER_CREATE_INFO_EXT };
--- a/src/core/deviceio/cpp/schema_tracker.cpp
+++ b/src/core/deviceio/cpp/schema_tracker.cpp
@@ -4,6 +4,7 @@
 #include "inc/deviceio/schema_tracker.hpp"
 
 #include <openxr/openxr.h>
+#include <oxr_utils/oxr_funcs.hpp>
 
 #include <XR_NVX1_tensor_data.h>
 #include <cassert>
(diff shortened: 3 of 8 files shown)


src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -0,0 +1,246 @@
...
+        result = core_funcs.xrGetSystemProperties(handles.instance, system_id, &system_props);
+        if (XR_SUCCEEDED(result) && !body_tracking_props.supportsBodyTracking)
+        {
+            throw std::runtime_error("Body tracking not supported by this system");
+        }
+    }
+
+    // Get extension function pointers using the provided xrGetInstanceProcAddr
+    handles.xrGetInstanceProcAddr(
+        handles.instance, "xrCreateBodyTrackerBD", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_create_body_tracker_));
```
life1ess: Why is the pointer type so different that it needs to be reinterpreted?

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -0,0 +1,246 @@
...
+    body_pose_.timestamp = std::make_shared<Timestamp>(time, time);
+
+    // Ensure joints struct is allocated
+    if (!body_pose_.joints)
+    {
+        body_pose_.joints = std::make_unique<BodyJointsPico>();
+    }
+
+    // Get mutable access to the joints array
+    auto* mutable_array = const_cast<flatbuffers::Array<BodyJointPose, 24>*>(body_pose_.joints->joints());
```
life1ess: Why is it safe to do a const cast here? Is const coming from an external library?

examples/oxr/python/test_full_body_tracker.py
```diff
@@ -0,0 +1,165 @@
...
+            print("[Test 9] Cleanup...")
+            print("✓ Resources will be cleaned up when exiting 'with' blocks (RAII)")
+            print()
+
+except RuntimeError as e:
+    print(f"❌ Failed to initialize: {e}")
+    print()
+    print("This test requires a PICO device with XR_BD_body_tracking extension support.")
+    print("If you're running on a non-PICO device, this is expected.")
+    sys.exit(1)
```
aristarkhovNV: This is too big of an assumption to make in a generic error handler. Message could be misleading if the actual error is unrelated to PICO.

Recommendation: 1. Let the error propagate. 2. Ensure that the exception description contains relevant and actionable info when it is actually due to "missing a PICO device with XR_BD_body_tracking extension support"

examples/oxr/python/test_full_body_tracker.py
```diff
@@ -0,0 +1,165 @@
...
+                                status += f" | Head: [{pos.x:+.2f}, {pos.y:+.2f}, {pos.z:+.2f}]"
+
+                        print(f"  [{elapsed:5.2f}s] Frame {frame_count:4d} | {status}")
+                        last_status_print = current_time
+
+                    frame_count += 1
+                    time.sleep(0.016)  # ~60 FPS
+
+            except KeyboardInterrupt:
+                print("\n⚠️  Interrupted by user")
```
aristarkhovNV: we should be able to rely on default interrupt handling

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -0,0 +1,246 @@
...
+        XrSystemBodyTrackingPropertiesBD body_tracking_props{ XR_TYPE_SYSTEM_BODY_TRACKING_PROPERTIES_BD };
+        XrSystemProperties system_props{ XR_TYPE_SYSTEM_PROPERTIES };
+        system_props.next = &body_tracking_props;
+
+        result = core_funcs.xrGetSystemProperties(handles.instance, system_id, &system_props);
+        if (XR_SUCCEEDED(result) && !body_tracking_props.supportsBodyTracking)
+        {
+            throw std::runtime_error("Body tracking not supported by this system");
+        }
+    }
```
aristarkhovNV, line 75: else: throw

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -0,0 +1,246 @@
...
+    if (XR_SUCCEEDED(result))
+    {
+        XrSystemBodyTrackingPropertiesBD body_tracking_props{ XR_TYPE_SYSTEM_BODY_TRACKING_PROPERTIES_BD };
+        XrSystemProperties system_props{ XR_TYPE_SYSTEM_PROPERTIES };
+        system_props.next = &body_tracking_props;
+
+        result = core_funcs.xrGetSystemProperties(handles.instance, system_id, &system_props);
+        if (XR_SUCCEEDED(result) && !body_tracking_props.supportsBodyTracking)
+        {
+            throw std::runtime_error("Body tracking not supported by this system");
```
aristarkhovNV, line 73: also throw if `!XR_SUCCEEDED(result)`

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    // Get extension function pointers using the provided xrGetInstanceProcAddr
+    handles.xrGetInstanceProcAddr(
+        handles.instance, "xrCreateBodyTrackerBD", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_create_body_tracker_));
+    handles.xrGetInstanceProcAddr(
+        handles.instance, "xrDestroyBodyTrackerBD", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_destroy_body_tracker_));
+    handles.xrGetInstanceProcAddr(
+        handles.instance, "xrLocateBodyJointsBD", reinterpret_cast<PFN_xrVoidFunction*>(&pfn_locate_body_joints_));
```
aristarkhovNV: check return values, throw on failure

src/core/deviceio/cpp/full_body_tracker_pico.cpp
aristarkhovNV: Seems unsafe - `const_cast` should not be used to modify the subject bypassing write protections

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -0,0 +1,246 @@
...
+        "Head", // 15
+        "Left_Shoulder", // 16
+        "Right_Shoulder", // 17
+        "Left_Elbow", // 18
+        "Right_Elbow", // 19
+        "Left_Wrist", // 20
+        "Right_Wrist", // 21
+        "Left_Hand", // 22
+        "Right_Hand" // 23
+    };
```
aristarkhovNV: we should be able to programmatically enumerate enum entries generated from fbs

src/core/deviceio/cpp/full_body_tracker_pico.cpp
jiwenc-nv: Short answer is that OpenXR designed it to be used like this.

Long answer:
That's because `xrGetInstanceProcAddr` itself lets us load any function pointers without specifying the concrete function pointer type.

As openXR spec state: "The returned function pointer is of type [PFN_xrVoidFunction](https://registry.khronos.org/OpenXR/specs/1.0/man/html/PFN_xrVoidFunction.html), and must be cast to the type of the function being queried."

https://registry.khronos.org/OpenXR/specs/1.0/man/html/xrGetInstanceProcAddr.html

src/core/deviceio/cpp/full_body_tracker_pico.cpp
jiwenc-nv: ah, actually, flatbuffer handles that with `--gen-mutable` options:
https://flatbuffers.dev/flatc/

fixed all uses.

examples/oxr/python/test_full_body_tracker.py
jiwenc-nv: This is actually the top level main already. We probably want to catch all and be more generic on the error message here.

examples/oxr/python/test_full_body_tracker.py
jiwenc-nv: done.

src/core/deviceio/cpp/full_body_tracker_pico.cpp
jiwenc-nv: done.

jiwenc-nv: done.

examples/oxr/python/test_full_body_tracker.py
aristarkhovNV: Yea this is exactly the point. Top level handler could catch anything, including stuff unrelated to PICO. We could tweak the message to be more generic, or might remove the handler altogether. If we let the exception propagate out of main then Python interpreter will print a full traceback ending with the exception type and its message

src/core/deviceio/cpp/full_body_tracker_pico.cpp
jiwenc-nv: done.

src/core/deviceio/cpp/full_body_tracker_pico.cpp
jiwenc-nv: done. Addded an inline util function `loadExtensionFunction` in oxr_funcs.hpp

src/core/deviceio/cpp/full_body_tracker_pico.cpp
jiwenc-nv: actually, this whole get_joint_name method is just redundant. removed.

examples/oxr/python/test_full_body_tracker.py
```diff
@@ -0,0 +1,158 @@
...
+            print()
+
+            # Cleanup
+            print("[Test 9] Cleanup...")
+            print("✓ Resources will be cleaned up when exiting 'with' blocks (RAII)")
+            print()
+
+except Exception as e:
+    print(f"Unknown error: {e}")
+    sys.exit(1)
```
aristarkhovNV: lets adopt unix style error reporting:
1. use stderr
2. prepend the message with the program name

```suggestion
except Exception as e:
    sys.exit(f"{sys.argv[0]}: {e}")
```

src/core/deviceio/cpp/CMakeLists.txt
```diff
@@ -32,9 +33,7 @@ add_library(deviceio_core STATIC
     inc/deviceio/schema_tracker.hpp
     inc/deviceio/generic_3axis_pedal_tracker.hpp
     inc/deviceio/frame_metadata_tracker.hpp
-    inc/deviceio/deviceio_session.hpp
```
aristarkhovNV, line 35: why removed

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -0,0 +1,214 @@
...
+#include "inc/deviceio/full_body_tracker_pico.hpp"
+
+#include "inc/deviceio/deviceio_session.hpp"
+
+#include <cassert>
+#include <cstring>
+#include <iostream>
+
+// XR_BD_body_tracking extension types are defined in <openxr/openxr.h>
+// which is included via oxr_utils/oxr_funcs.hpp
```
aristarkhovNV: obsolete comment, remove

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -0,0 +1,214 @@
...
+    {
+        throw std::runtime_error("OpenXR: failed to get system: " + std::to_string(result));
+    }
+
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrCreateBodyTrackerBD",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_create_body_tracker_));
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrDestroyBodyTrackerBD",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_destroy_body_tracker_));
+    loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrLocateBodyJointsBD",
+                          reinterpret_cast<PFN_xrVoidFunction*>(&pfn_locate_body_joints_));
```
aristarkhovNV, line 86: you could make `loadExtensionFunction` a template and embed `reinterpret_cast` into it

src/core/deviceio/cpp/full_body_tracker_pico.cpp
jiwenc-nv: removed.

src/core/deviceio/cpp/full_body_tracker_pico.cpp
aristarkhovNV: also make it member of handles, the code would look like
```cpp
handles.loadExtensionFunction("xrLocateBodyJointsBD", &pfn_locate_body_joints_);
```

examples/oxr/python/test_full_body_tracker.py
jiwenc-nv: done.

src/core/deviceio/cpp/CMakeLists.txt
jiwenc-nv: rebased it wong.

src/core/deviceio/cpp/full_body_tracker_pico.cpp
jiwenc-nv: Let's refactor that another time:

```
// Wrapper for OpenXR session handles
// This struct is used to pass OpenXR session information between modules
// without requiring them to link against the OpenXR loader library
struct OpenXRSessionHandles
{
    XrInstance instance{};
    XrSession session{};
    XrSpace space{};
    PFN_xrGetInstanceProcAddr xrGetInstanceProcAddr{};
};
```

The `OpenXRSessionHandles` is a relatively thin wrapper. Probably not ideally to make `loadExtensionFunction` part of it yet.

PR #98, opened 2026-02-07
Update Numpy version & uv.lock files

jiwenc-nv:


examples/oxr/python/pyproject.toml
```diff
@@ -7,13 +7,10 @@ version = "0.0.0"  # Internal examples - not versioned
 description = "OpenXR Tracking Python Examples"
 requires-python = ">=3.10,<3.13"
 dependencies = [
-    "isaacteleop",
```
aristarkhovNV, line 10: why would it not require isaacteleop? also inconsistent with the readme

examples/oxr/python/README.md
```diff
@@ -80,7 +80,7 @@ uv run test_session_sharing.py
 
 Dependencies are managed through `pyproject.toml` and automatically handled by `uv`:
 - `isaacteleop` - The OpenXR tracking wheel (built from source)
-- `numpy>=1.19.0` - Numerical computations used by the examples
+- `numpy>=1.22.0` - Numerical computations used by the examples
```
aristarkhovNV: do we really want to list the deps in the readme? They are supposed do me the implementation details.

scripts/update_uv_lockfiles.sh
aristarkhovNV, line 1: Ohh if we've got so many locks to require a maintanance script then maybe we are doing it wrong? I wonder if it's possible to incorporate all samples under the same pyproject or something such that we have one centralized uv.lock

jiwenc-nv: I'm wondering about that too, felt we have so many `pyproject.toml` files...

examples/oxr/python/README.md
jiwenc-nv: right. let's remove that.

examples/oxr/python/pyproject.toml
jiwenc-nv: to prevent `uv lock` locals against a locally built wheel.

jiwenc-nv: see internal slack for discussions.

scripts/update_uv_lockfiles.sh
jiwenc-nv: Given we decided to remove all the uv.lock files, this script becomes irrelevant.

examples/oxr/python/pyproject.toml
jiwenc-nv: No need to change `pyproject.toml` for now, as we are moving away from uv.lock.

PR #105, opened 2026-02-10
Implement foot pedal device to read from Linux joystick API

jiwenc-nv:


src/plugins/generic_3axis_pedal/generic_3axis_pedal_plugin.cpp
```diff
@@ -0,0 +1,146 @@
...
+    FD_SET(device_fd_, &read_fds);
+    struct timeval timeout = { 0, 0 };
+
+    int ret = select(device_fd_ + 1, &read_fds, nullptr, nullptr, &timeout);
+    if (ret < 0)
+    {
+        if (errno == EINTR)
+            return;
+        close_device();
+        push_current_state();
```
aristarkhovNV: I suppose this is here to push a tensor with `is_valid==false`, but does it actually matter? The plugin might as well just crash and push nothing, so the receiving side must take dependency on `is_valid` turning false if things go south - they should be looking for tensor lost instead.

src/plugins/generic_3axis_pedal/generic_3axis_pedal_plugin.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+        if ((event.type & JS_EVENT_AXIS) != 0u && event.number <= 2)
+        {
+            axes_[event.number] = normalize_axis(event.value);
+        }
+
+        FD_ZERO(&read_fds);
+        FD_SET(device_fd_, &read_fds);
+        timeout = { 0, 0 };
+        ret = select(device_fd_ + 1, &read_fds, nullptr, nullptr, &timeout);
```
aristarkhovNV: handle errors from this select

src/plugins/generic_3axis_pedal/generic_3axis_pedal_plugin.cpp
```diff
@@ -0,0 +1,146 @@
...
+void Generic3AxisPedalPlugin::update()
+{
+    assert(device_fd_ >= 0);
+
+    fd_set read_fds;
+    FD_ZERO(&read_fds);
+    FD_SET(device_fd_, &read_fds);
+    struct timeval timeout = { 0, 0 };
+
+    int ret = select(device_fd_ + 1, &read_fds, nullptr, nullptr, &timeout);
```
aristarkhovNV: Would it be possible to refactor this to have a single select in the loop? something like `while(select()) { read(); push() }`. Two selects make error handling inconsistent.

src/plugins/generic_3axis_pedal/generic_3axis_pedal_plugin.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+    void update();
+
+private:
+    void open_device();
+    void close_device();
+    void push_current_state();
+
+    std::string device_path_;
+    std::string collection_id_;
```
aristarkhovNV: these are used to pass value to open_device once, no need to be stored

src/plugins/generic_3axis_pedal/main.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    // Push data at 90 Hz
+    // TODO: Make the device push rate configurable
+    const auto frame_duration = std::chrono::nanoseconds(1000000000 / 90);
+
+    while (true)
+    {
+        auto frame_start = std::chrono::steady_clock::now();
+        plugin.update();
+        std::this_thread::sleep_until(frame_start + frame_duration);
```
aristarkhovNV: to minimize drift slip until `program_start + frame_duration * frame_count`

src/plugins/generic_3axis_pedal/main.cpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    return 0;
+}
+catch (const std::exception& e)
+{
+    std::cerr << (argc >= 1 ? argv[0] : "generic_3axis_pedal_plugin") << ": " << e.what() << std::endl;
```
aristarkhovNV: This seems so unusual that most programs don't bother handling `argc==0`. An alternative to hardcoding the program name could be adding `if (!argc) return 1;` in the beginning of main (just to keep static analysis tools happy)

src/plugins/generic_3axis_pedal/generic_3axis_pedal_plugin.cpp
jiwenc-nv: done.

src/plugins/generic_3axis_pedal/generic_3axis_pedal_plugin.cpp
jiwenc-nv: done.

src/plugins/generic_3axis_pedal/generic_3axis_pedal_plugin.cpp
jiwenc-nv: Your comment actually reminded some some feature I implemented back in Isaac OS: "device hot plug".

So basically each `update` can fail, close device, but don't throw. next update we can open device again.

jiwenc-nv: adding a follow up commit for now.

src/plugins/generic_3axis_pedal/generic_3axis_pedal_plugin.hpp
jiwenc-nv: removed.

src/plugins/generic_3axis_pedal/main.cpp
jiwenc-nv: done. good point.

src/plugins/generic_3axis_pedal/main.cpp
jiwenc-nv: updated.

But tbh, i just don't know how to trigger a case where `argc == 0`

src/plugins/generic_3axis_pedal/main.cpp
```diff
@@ -0,0 +1,45 @@
...
+    return 1;
+}
+catch (...)
+{
+    std::cerr << (argc >= 1 ? argv[0] : "generic_3axis_pedal_plugin") << ": Unknown error" << std::endl;
```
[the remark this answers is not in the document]
jiwenc-nv: done.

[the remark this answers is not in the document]
jiwenc-nv: done.

src/plugins/generic_3axis_pedal/main.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <iostream>
+#include <string>
+#include <thread>
+
+using namespace plugins::generic_3axis_pedal;
+
+int main(int argc, char** argv)
+try
+{
+    if (argc == 0)
```
aristarkhovNV, line 17:
```suggestion
    if (argc <= 1)
```

PR #110, opened 2026-02-10
Refactor the oak-d plugin

shaosu-nvidia:
As we have discussed the camera recording during the [design](https://docs.google.com/document/d/1F34x8NclO8SX5z-WZdziciD2fX6VdDFoUvPppk1A4y8/edit?tab=t.0#heading=h.hctreigdgjzi) session and become more clear on the future feature request around camera recording (and oakd device specifically), the original design idea to have a generic plugin interface is not great anymore. In this code review, I have:

- Deleted src/core/camera/ entirely (ICamera, CameraPlugin, CameraConfig, RecordConfig, FrameMetadataPusher)
Moved OAK-specific code into src/plugins/oakd/core/ under plugins::oakd with concrete types (OakConfig, OakFrame, FrameSink)
- Renamed generic identifiers to OAK-specific: camera.fbs -> oakd.fbs, FrameMetadataTracker -> - - FrameMetadataTrackerOak, bindings/tests updated accordingly
- Simplified main.cpp: capture loop runs inline, signal handling uses a plain atomic flag instead of a global pointer
- Also rename oakd to oak across multiple files

The reasons behind are:

1. Unclear of other camera devices, which may include different sensor data. Unless we have introduced more camera devices, it's better to have the current plugin for oakd specific instead of a blindly self-defined camera interface.
2. We may have to extend more functionality for oak cameras. E.g.
    a. Make oak support multiple camera streams, 1 rgb + 2 stereo.
    b. have a single oak plugin implementation but multiple plugin instance connecting to oak-d, oak-1, etc. 
3. Even though we will consider a generic camera instance, we would like to do more research to find a library/sdk that supports multiple vendors instead of doing by ourselves.

### DIFF
--- a/src/plugins/oak/core/frame_sink.cpp
+++ b/src/plugins/oak/core/frame_sink.cpp
@@ -1,12 +1,9 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 #include "frame_sink.hpp"
 
 #include <flatbuffers/flatbuffers.h>
-#include <oxr/oxr_session.hpp>
-#include <pusherio/schema_pusher.hpp>
-#include <schema/oak_generated.h>
 
 #include <filesystem>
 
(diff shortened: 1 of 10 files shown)


src/plugins/oak/core/frame_sink.cpp
```diff
@@ -0,0 +1,83 @@
...
+              return record_path;
+          }())
+{
+    if (!collection_id.empty())
+    {
+        m_pusher = std::make_unique<MetadataPusher>(collection_id);
+    }
+}
+
+FrameSink::~FrameSink() = default;
```
qingsi-at-nv: nit: not sure about the style, I typically move this to header.

src/plugins/oak/core/frame_sink.cpp
```diff
@@ -0,0 +1,83 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
qingsi-at-nv: 2025-2026, here and else where

PR #112, opened 2026-02-11
add plugin_args support to plugin_manager

shaosu-nvidia:


src/core/plugin_manager/cpp/plugin.cpp
```diff
@@ -128,6 +134,13 @@ void Plugin::start_process(const std::string& command, const std::string& workin
             args_str.push_back("--plugin-root-id=" + plugin_root_id);
         }
 
+        // Append any plugin arguments (e.g. --device-path=..., --device-mxid=...)
```
jiwenc-nv: is there any particular format a user need to follow for the argument lists? 

or it's just `argv[]` passing to the plugin binary?

if I read this correctly, it's the later?

src/core/teleop_session_manager/README.md
```diff
@@ -73,7 +73,8 @@ PluginConfig(
     plugin_name="controller_synthetic_hands",
     plugin_root_id="synthetic_hands",
     search_paths=[Path("/path/to/plugins")],
-    enabled=True
+    enabled=True,
+    plugin_args=[],
```
jiwenc-nv: Similar to my last question: can you please document how users can use the `plugin_args`?

src/core/plugin_manager/cpp/plugin.cpp
jiwenc-nv: also, what if users accidentally specify `--plugin-root-id=test`? That will conflict with the "plugin root ID argument" set above?

shaosu-nvidia: that's a good catch about `--plugin-root-id`. I should add some format check probably.

but it is used with manager.start(plugin_name, plugin_root_id, ["--arg1=val1", "--arg2=val2"]).

shaosu-nvidia: Let me know if this looks good.

src/core/plugin_manager/cpp/plugin.cpp
```diff
@@ -125,7 +132,14 @@ void Plugin::start_process(const std::string& command, const std::string& workin
...
         if (!plugin_root_id.empty())
         {
-            args_str.push_back("--plugin-root-id=" + plugin_root_id);
+            args_str.push_back(plugin_root_id_prefix + plugin_root_id);
+        }
+
+        // Append any plugin arguments (skip --plugin-root-id= so plugin_root_id parameter cannot be overridden)
+        for (const auto& arg : plugin_args)
+        {
+            if (!arg.empty() && !arg.starts_with(plugin_root_id_prefix))
```
jiwenc-nv: instead of hard code `plugin_root_id_prefix` here, why don't we just check whether `arg` is already in the `args_str`.

`args_str.find(arg) == args_str.end()`?

So that it will work in the future in case we add more args in the framework.

We can log an error as well.

jiwenc-nv: feel free to land the PR after you make the change. no need to go another round.

PR #113, opened 2026-02-11
Fixes quat ordering w.r.t. retargeting

rwiltz:


src/core/schema/fbs/pose.fbs
```diff
@@ -8,6 +8,6 @@ struct Pose {
   // Position of the pose (x, y, z) in meters.
   position: Point (id: 0);
 
-  // Orientation of the pose (w, x, y, z).
+  // Orientation of the pose (x, y, z, w).
```
jiwenc-nv, line 11: no need to block this PR, but wondering how we can better enforce the order there.

PR #117, opened 2026-02-12
Add transform utilities, TransformMatrix type, and .transformed() methods

nvddr:
Add a utilities module with reusable transform nodes and math:
- TransformMatrix tensor type for 4x4 homogeneous transforms
- ControllerTransform, HandTransform, HeadTransform retargeter nodes
- Shared transform_utils with position/orientation math (pure numpy)

Each DeviceIO source node gains a .transformed() convenience method that wires up the appropriate transform node, allowing callers to apply a coordinate-frame transform to tracked data inline.

### DIFF
--- a/src/core/retargeting_engine/python/utilities/hand_transform.py
+++ b/src/core/retargeting_engine/python/utilities/hand_transform.py
@@ -1,0 +1,128 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Hand Transform Node - Applies a 4x4 transform to hand tracking data.
+
+Transforms all hand joint positions and orientations using a homogeneous
+transformation matrix while preserving joint radii, validity, active state,
+and timestamp fields.
+
+The transform matrix is received as a tensor input from the graph, typically
+provided by a TransformSource node.
+
+Note: All 26 joint positions and orientations are transformed (not just the
+wrist) since all joints share the same coordinate frame. Transforming only
+the wrist while leaving other joints untransformed would break the hand skeleton.
+"""
+
+import copy
+
+import numpy as np
+
+from ..interface.base_retargeter import BaseRetargeter
+from ..interface.retargeter_core_types import RetargeterIO, RetargeterIOType
+from ..interface.tensor_group import TensorGroup
+from ..tensor_types import HandInput, HandInputIndex, TransformMatrix
+from .transform_utils import (
+    decompose_transform,
+    transform_positions_batch,
+    transform_orientations_batch,
+)
+
+
+# Index of the matrix field within the TransformMatrix TensorGroupType
+_MATRIX_INDEX = 0
+
+
+class HandTransform(BaseRetargeter):
+    """
+    Applies a 4x4 homogeneous transform to hand tracking data.
+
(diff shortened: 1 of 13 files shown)


src/core/retargeting_engine/python/utilities/hand_transform.py
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+    def output_spec(self) -> RetargeterIOType:
+        """Declare transformed hand output specs for left and right."""
+        return {
+            self.LEFT: HandInput(),
+            self.RIGHT: HandInput(),
+        }
+
+    def compute(self, inputs: RetargeterIO, outputs: RetargeterIO) -> None:
```
jiwenc-nv, line 96: Wondering whether we can reduce some boilerplate code here.

I can't think this through immediately. But intuitively, all the transforms we are dealing with is 4x4 matrix, then maybe we can just have one 

`SpatialTransform` node that takes in a 4x4xN tensor, where N is number of joints, then for the "Pass through unchanged fields", they probably don't need to be passed into the retargeter here?

src/core/retargeting_engine/python/utilities/hand_transform.py
```diff
@@ -0,0 +1,137 @@
...
+            positions, rotation, translation
+        )
+
+        # Transform all joint orientations (26, 4)
+        orientations = np.from_dlpack(inp[HandInputIndex.JOINT_ORIENTATIONS])
+        out[HandInputIndex.JOINT_ORIENTATIONS] = transform_orientations_batch(
+            orientations, rotation
+        )
+
+        # Pass through unchanged fields
```
jiwenc-nv: is there a way we can do:
```
# duplicate the object
out = in.duplicate()

# update specific fields that need update
out[HandInputIndex.JOINT_POSITIONS] = ...
out[HandInputIndex.JOINT_ORIENTATIONS] = ...
```

This way we don't need to explicitly list all the fields and risking missing something.

src/core/retargeting_engine/python/utilities/hand_transform.py
jiwenc-nv: Feel free to land this as-is.  We can figure out reducing boilerplate code in the future

src/core/retargeting_engine/python/utilities/hand_transform.py
nvddr: Out is already a pre-alocated tensor, what you are suggesting is an extra copy... I think this just points to requiring a robust unit test that checks for all fields for an identity transform.

jiwenc-nv: where does the "pre-alocated tensor" happen? Is there a way to do a "copy ctor" style allocation?

nvddr: I think it makes sense for controller though.

nvddr: Done, I think it's cleaner.

src/core/retargeting_engine/python/utilities/hand_transform.py
nvddr: Sounds good. Added tests to the code, so if we change, we can use the same tests to validate behavior.

src/core/retargeting_engine/python/utilities/hand_transform.py
jiwenc-nv: thanks!

PR #115, opened 2026-02-11
Move [tool.uv] config to CMake install step

nvddr:
The source pyproject.toml files now stay tool-agnostic. The [tool.uv] block (find-links, python-preference, environments) is appended only to the installed copy by CMake at configure time. This keeps the source files clean and avoids duplicating uv-specific settings across examples.

PR #121, opened 2026-02-12
Add ROS2 RA

life1ess:


CMakeLists.txt
```diff
@@ -115,8 +115,8 @@ if(BUILD_EXAMPLES)
     add_subdirectory(examples/oxr)
     add_subdirectory(examples/retargeting)
     add_subdirectory(examples/teleop_session_manager)
+    add_subdirectory(examples/teleop_ros2)
     add_subdirectory(examples/schemaio)
-    add_subdirectory(examples/native_openxr)
```
jiwenc-nv, line 120: looks like a rebase hiccup

examples/teleop_ros2/README.md
```diff
@@ -0,0 +1,63 @@
...
+SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Teleop ROS2 Reference (Python)
+
+Reference ROS2 bridge for Isaac Teleop data. It publishes the same topic set as
+`xr_teleop_ros2` using standard ROS2 message types.
+
+## Published Topics
```
jiwenc-nv, line 18: is there a way to run some robot backend so that mujoco can be executed as well? then we can have something end to end testable.

life1ess: Once we have integration done on the Gear (Sonic) side, we can use it for testing

CMakeLists.txt
nv-mhaselton: thanks, fixed

examples/teleop_ros2/python/teleop_ros2_bridge.py
```diff
@@ -0,0 +1,265 @@
...
+            )
+        )
+
+    config = TeleopSessionConfig(
+        app_name="TeleopRos2Reference",
+        pipeline=pipeline,
+        plugins=plugins,
+    )
+
+    rate = node.create_rate(args.rate_hz)
```
kajananchinniahNV: Nit: we should consider putting it in a class and using a timer callback 

Some potentially useful reading:
- https://docs.ros.org/en/jazzy/How-To-Guides/Using-callback-groups.html
- https://docs.ros.org/en/jazzy/Tutorials/Beginner-Client-Libraries/Writing-A-Simple-Py-Publisher-And-Subscriber.html#write-the-publisher-node

examples/teleop_ros2/python/teleop_ros2_bridge.py
```diff
@@ -0,0 +1,265 @@
...
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description="Teleop ROS2 reference bridge")
+    parser.add_argument("--hand-topic", default="xr_teleop/hand")
+    parser.add_argument("--twist-topic", default="xr_teleop/root_twist")
+    parser.add_argument("--pose-topic", default="xr_teleop/root_pose")
+    parser.add_argument("--controller-topic", default="xr_teleop/controller_data")
+    parser.add_argument("--frame-id", default="world")
+    parser.add_argument("--rate-hz", type=float, default=60.0)
+    parser.add_argument("--use-mock-operators", action="store_true")
```
kajananchinniahNV: Once we're ready to finalize this better, it might to declare these as parameters.

Some useful examples / reading:
- https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_cumotion/blob/main/isaac_ros_cumotion/isaac_ros_cumotion/cumotion_planner.py#L70
- https://docs.ros.org/en/jazzy/Tutorials/Beginner-Client-Libraries/Using-Parameters-In-A-Class-Python.html

examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -0,0 +1,265 @@
...
+    parser.add_argument("--use-mock-operators", action="store_true")
+    args = parser.parse_args()
+
+    rclpy.init()
+    node = Node("teleop_ros2_reference")
+
+    pub_hand = node.create_publisher(PoseArray, args.hand_topic, 10)
+    pub_twist = node.create_publisher(TwistStamped, args.twist_topic, 10)
+    pub_pose = node.create_publisher(PoseStamped, args.pose_topic, 10)
+    pub_controller = node.create_publisher(ByteMultiArray, args.controller_topic, 10)
```
kajananchinniahNV, line 181: I would recommend avoiding using a ByteMultiArray just because they're slightly harder to use for users

Some other possibilities are:
- We could publish each part as a separate topic (probably not ideal) 
- See if we can use ROS' joint states topic: https://docs.ros2.org/foxy/api/sensor_msgs/msg/JointState.html and split the parts where we can't use it into separate messages (we can discuss what makes the most sense)
- Use Isaac ROS' TensorList msg type: https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_common/tree/main/isaac_ros_tensor_list_interfaces/msg
- Create our own msg type (probably not ideal)


I personally would prefer option 2 and 3

examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -0,0 +1,265 @@
...
+        )
+
+    config = TeleopSessionConfig(
+        app_name="TeleopRos2Reference",
+        pipeline=pipeline,
+        plugins=plugins,
+    )
+
+    rate = node.create_rate(args.rate_hz)
+    with TeleopSession(config) as session:
```
kajananchinniahNV, line 226: Is the idea that TeleopSession handles receiving joint information, joystick information, etc?

examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -0,0 +1,312 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES.
```
kajananchinniahNV, line 2: Also just curious are we planning on converting the stack into C++ or will we be sticking to Python?

examples/teleop_ros2/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+- `xr_teleop/hand` (`geometry_msgs/PoseArray`)
+  - `poses[0]`: Right hand wrist pose (if active)
+  - `poses[1]`: Left hand wrist pose (if active)
+  - `poses[2+]`: Finger joint poses (all joints except palm/wrist, right then left)
+- `xr_teleop/root_twist` (`geometry_msgs/TwistStamped`)
+- `xr_teleop/root_pose` (`geometry_msgs/PoseStamped`)
+- `xr_teleop/controller_data` (`std_msgs/ByteMultiArray`, msgpack-encoded dictionary)
```
tifchen-nvda, line 26: Can we make sure we have head pose in controller data? Thanks

life1ess: Let's limit the scope of this PR and add head pose in a follow up PR. I added a task to the tracker to make sure we don't forget about it.

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: Currently Teleop Core is primarily in Python. In the future we may consider introducing a C++-only code path for performance reasons if needed

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: Yes, the session encapsulates Isaac Teleop state

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: Right now we're intentionally trying not to introduce dependencies on Isaac ROS to maintain compatibility with vanilla ROS2. For Option 2, what would be the type of those separate messages?

examples/teleop_ros2/python/teleop_ros2_publisher.py
kajananchinniahNV: Is there any way we can try making this more modular or ROS-like? 

Typically we expect joint states to be subscribed to and coming directly from the robot. ROS also provides a joystick msg (https://github.com/ros2/common_interfaces/blob/jazzy/sensor_msgs/msg/Joy.msg). 

Would this current structure limit the embodiments we can support is my big concern

examples/teleop_ros2/python/teleop_ros2_publisher.py
kajananchinniahNV: For Option 2:
- We could create our own msg and package it as part of this package
- We can split the controller msg into multiple smaller chunks. Button presses can use: https://github.com/ros2/common_interfaces/blob/jazzy/sensor_msgs/msg/Joy.msg, some of the ones based more on feedback can use: https://github.com/ros2/common_interfaces/blob/jazzy/sensor_msgs/msg/JoyFeedback.msg but I'm not sure how we'd want to handle the position / orientation msgs

Admittedly the second option might be a bit awkward because I think the joystick msgs doesn't fully encapsulate everything we'd want from a msg. I'll think about this more and think if there's a better for us to do this

examples/teleop_ros2/python/teleop_ros2_bridge.py
life1ess: Thanks for the feedback, will do it in a follow up PR

examples/teleop_ros2/python/teleop_ros2_bridge.py
life1ess: Thanks for the feedback, will do it in a follow up PR

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: Agree, let's go for the option 2.1 in a follow up PR

examples/teleop_ros2/README.md
```diff
@@ -0,0 +1,76 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Teleop ROS2 Reference (Python)
```
jaiveersinghNV: Nit - "ROS 2" not "ROS2" in writing. `ros2` is fine in code

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: Isaac Teleop session exposes all signals that are available from the CloudXR runtime, that cannot be split into modules. The teleop publisher then splits them into different ROS2 topics that can be subscribed to independently. What modularity did you have in mind?

examples/teleop_ros2/README.md
jaiveersinghNV, line 26: Remind me, `controller_data` is an arbitrary per-HW-config list of bytes, right?

Of course, if there is any known structure to this data, we should prefer to represent it explicitly through additional fields

examples/teleop_ros2/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+- `xr_teleop/hand` (`geometry_msgs/PoseArray`)
+  - `poses[0]`: Right hand wrist pose (if active)
+  - `poses[1]`: Left hand wrist pose (if active)
+  - `poses[2+]`: Finger joint poses (all joints except palm/wrist, right then left)
+- `xr_teleop/root_twist` (`geometry_msgs/TwistStamped`)
+- `xr_teleop/root_pose` (`geometry_msgs/PoseStamped`)
+- `xr_teleop/controller_data` (`std_msgs/ByteMultiArray`, msgpack-encoded dictionary)
+
+## Run in Docker (ROS2 Humble)
```
jaiveersinghNV: Do we have the capacity to also test using ROS 2 Jazzy? Isaac ROS 4.X uses Jazzy, and that's the platform we expect to eventually ship the Isaac ROS Teleop integration on. It would be nice to catch compatibility issues earlier rather than later - but nbd if we don't have resources for that now

life1ess: In theory we can look into supporting Jazzy in addition to Humble once we're done with Sonic integration

examples/teleop_ros2/README.md
life1ess: Done

examples/teleop_ros2/README.md
life1ess: controller_data is a msgpack-encoded dictionary, and the dictionary is HW-config specific

life1ess: per comment above, I'll replace it with a custom message in a follow up PR

src/core/python/requirements-retargeters.txt
```diff
@@ -2,3 +2,5 @@ dex-retargeting>=0.4.6,<0.6.0
 scipy>=1.15.0
 pyyaml>=6.0.3
 torch>=2.7.0
+# nlopt: no aarch64 wheel/sdist on PyPI for >=2.7; pin to last version with sdist
+nlopt==2.6.2
```
jiwenc-nv, line 6: nlopt by default is LGPL, which can be a problem? How were we dealing with this in the past?

https://github.com/stevengj/nlopt?tab=License-1-ov-file#readme

examples/teleop_ros2/python/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+]
+
+[tool.uv]
+find-links = ["/opt/isaacteleop/install/wheels"]
+python-preference = "only-managed"
+
+# Build-time dependencies for packages that must compile from source on aarch64.
+# nlopt has no pre-built wheel for Linux ARM64; swig + numpy are needed to build it.
+[tool.uv.extra-build-dependencies]
+nlopt = ["numpy"]
```
jiwenc-nv: We are not discouraging listing the [tool.uv] blocks in the pytproject.toml files.  Instead, use `install_python_example` like other samples.

This way we can stay tool-agnostic. See: #115 for more details.

Also `find-links = ["/opt/isaacteleop/install/wheels"]` is Dockerfile specific, not very useful to hard code here. Again, `install_python_example` will use cmake to handle it automatically for you.

examples/teleop_ros2/python/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jiwenc-nv: nit: 2026

examples/teleop_ros2/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+RUN pip3 install --no-cache-dir uv
+
+WORKDIR /opt/isaacteleop
+COPY . /opt/isaacteleop
+
+RUN uv python install 3.10 --quiet \
+    && uv python find 3.10
+
+RUN --mount=type=cache,target=/opt/isaacteleop/build,id=isaacteleop-teleop-ros2-build \
+    --mount=type=cache,target=/root/.hunter,id=isaacteleop-hunter \
```
jiwenc-nv: I think we don't need hunter here, as it's only used by the OakD plugin for data collection projects. We can disable that plugin for now.

Probably reasonable to add a CMAKE flag called `BUILD_PLUGIN_OAK_CAMERA`, so that `BUILD_PLUGINS` and `BUILD_PLUGIN_OAK_CAMERA` both needs to be true for OakD to be built.

Just a suggestion and I don't have strong opinions. Feel free to take it or leave it.

@shaosu-nvidia WDYT?

examples/teleop_ros2/python/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+[project]
+name = "teleop-ros2-example"
+version = "0.1.0"
+description = "Isaac Teleop ROS2 Example"
+requires-python = ">=3.10,<3.13"
```
jiwenc-nv: given the Dockerfile hard coded to 3.10, we should lock it to 3.10 here as well.

examples/teleop_ros2/python/pyproject.toml
jiwenc-nv: @nvddr to keep me honest.

examples/teleop_ros2/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20)
+project(teleop_ros2_examples)
+
+# Python examples don't need CMake - they use uv
```
jiwenc-nv: please use the new `install_python_example` instead.

examples/teleop_ros2/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+exec "$@"
+EOF
+
+# Set BASH_ENV so non-interactive bash shells automatically source the environment setup.
+ENV BASH_ENV=/usr/local/bin/teleop-env-setup
+
+# Ensure interactive bash shells source the environment setup.
+RUN echo "source /usr/local/bin/teleop-env-setup" >> /etc/bash.bashrc
+
+WORKDIR /opt/isaacteleop/examples/teleop_ros2/python
```
jiwenc-nv: with `install_python_example` mentioned above, we can just do:

```
WORKDIR /opt/isaacteleop/install/examples/teleop_ros2/python
ENTRYPOINT ["/usr/local/bin/teleop-entrypoint", "uv", "run", "teleop_ros2_publisher.py"]
```

no need to run `uv sync` explicitly.

examples/teleop_ros2/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+    libxrender-dev \
+    libxfixes-dev \
+    libxxf86vm-dev \
+    swig \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip3 install --no-cache-dir uv
+
+WORKDIR /opt/isaacteleop
+COPY . /opt/isaacteleop
```
jiwenc-nv, line 53: I think we can get away with copy the `install` folder. it should have everything we need. this way, the container will be smaller if we ever want to distribute the container image.

src/core/python/requirements-retargeters.txt
life1ess: This doesn't introduce any new dependencies, only pins existing dependency to 2.6.2. On a side note, LGPL is very permissive - it allows linking with proprietary applications without requiring the application's source code to be released

examples/teleop_ros2/python/pyproject.toml
life1ess: Done

examples/teleop_ros2/python/pyproject.toml
life1ess: Done

examples/teleop_ros2/Dockerfile
life1ess: Done

examples/teleop_ros2/python/pyproject.toml
life1ess: Done

examples/teleop_ros2/CMakeLists.txt
life1ess: Done

examples/teleop_ros2/Dockerfile
life1ess: Done

examples/teleop_ros2/Dockerfile
life1ess: Done

PR #122, opened 2026-02-12
Enable external inputs in TeleopSession

nvddr:
Add support for pipelines with non-DeviceIO leaf nodes whose inputs are provided by the caller at each step():

- PassthroughInput: generic typed placeholder for external graph inputs
- TeleopSession discovers external leaves and validates inputs each frame
- step() accepts optional external_inputs dict for non-DeviceIO leaves
- get_external_input_specs() / has_external_inputs() for introspection
- Factory methods on TeleopSession for testability (mock hardware)
- Comprehensive unit tests (mocked hardware, ~1200 lines)
- External inputs example with coordinate-frame transform

### DIFF
--- a/examples/teleop_session_manager/python/external_inputs_example.py
+++ b/examples/teleop_session_manager/python/external_inputs_example.py
@@ -1,0 +1,326 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+External Inputs Example with TeleopSessionManager
+
+Demonstrates using TeleopSession with a pipeline that has both DeviceIO sources
+(auto-polled from hardware trackers) and external leaf nodes whose inputs are
+provided by the caller at each step().
+
+This example shows:
+1. Creating a pipeline with mixed DeviceIO and non-DeviceIO leaf nodes
+2. Using ValueInput for simple external values (robot EE state, transform matrix)
+3. Using get_external_input_specs() to discover required external inputs
+4. Passing external inputs to step()
+
+External input options -- any non-DeviceIO leaf node becomes an external input:
+
+    1. ValueInput (shortcut): Use for any external value you want to feed into
+       the graph without any computation. One typed input, one output, data
+       passed straight through. Good for sim state, transform matrices,
+       configuration parameters, etc.
+
+    2. Any existing BaseRetargeter: Any retargeter that is a leaf node (i.e. not
+       connected to a DeviceIO source) automatically becomes an external input.
+       You provide its inputs via step(external_inputs=...). This is useful when
+       you already have a node class and want the caller to drive it directly --
+       no need to wrap it in ValueInput.
+
+    3. Custom BaseRetargeter subclass: Write a new subclass when you need
+       computation on the external inputs before they enter the graph (e.g.,
+       fusing multiple sensor readings, applying filtering).
+
+Scenario:
+    A "delta position" retargeter takes two inputs:
+    - Controller grip position (from DeviceIO via ControllersSource, transformed)
+    - Current robot end-effector position (from the caller / simulator)
+
+    A coordinate-frame transform is applied to the controller data before
+    computing the delta, allowing the caller to align the VR tracking frame
+    with the robot's world frame.
+
+    Both the robot EE state and the transform matrix are simple external values,
+    so they both use ValueInput rather than custom subclasses.
+"""
+
+import sys
+import time
+import numpy as np
+from pathlib import Path
+
+from isaacteleop.retargeting_engine.deviceio_source_nodes import ControllersSource
+from isaacteleop.retargeting_engine.interface import (
+    BaseRetargeter,
+    OutputCombiner,
+    ValueInput,
+    TensorGroup,
+    TensorGroupType,
+)
+from isaacteleop.retargeting_engine.tensor_types import (
+    ControllerInput,
+    TransformMatrix,
+    NDArrayType,
+    DLDataType,
+)
+from isaacteleop.teleop_session_manager import (
+    TeleopSession,
+    TeleopSessionConfig,
+    PluginConfig,
+)
+
+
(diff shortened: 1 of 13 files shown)


src/core/teleop_session_manager/python/teleop_session.py
```diff
@@ -169,6 +299,72 @@ def get_elapsed_time(self) -> float:
...
+        Extracted as a separate method so that unit tests can override or mock
+        this without requiring actual plugin binaries.
+        
+        Args:
+            search_paths: List of directory paths to search for plugins
+        
+        Returns:
+            A PluginManager instance
+        """
+        return pm.PluginManager(search_paths)
```
aristarkhovNV: Providing these backdoor overrides just for the sake of tests feels wrong when they don't serve any real purpose in production. I wonder if it'd be possible to intercept the object construction with something like
```python
from unittest.mock import patch

@patch("mylib.Client")
def test_thing(mock_client):
    mock_client.return_value = MyClientMock()
    # calls inside mylib that do Client(...) now get MyClientMock()
```


Or inside a test context:
```python
from unittest.mock import patch

def test_thing():
    with patch("mylib.Client", return_value=MyClientMock()):
        # any Client(...) inside this block returns MyClientMock()
        ...
```

examples/teleop_session_manager/python/external_inputs_example.py
```diff
@@ -0,0 +1,296 @@
...
+                print(
+                    f"[{elapsed:5.1f}s] "
+                    f"delta_L=[{delta_l[0]:+.3f}, {delta_l[1]:+.3f}, {delta_l[2]:+.3f}]  "
+                    f"delta_R=[{delta_r[0]:+.3f}, {delta_r[1]:+.3f}, {delta_r[2]:+.3f}]"
+                )
+
+            # Slowly drift the simulated EE position (pretend robot is moving)
+            sim_ee_position += np.array([0.0001, 0.0, 0.0], dtype=np.float32)
+
+            time.sleep(0.016)  # ~60 FPS
```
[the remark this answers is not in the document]
nvddr: We are not handling exceptions,a nd allowing exceptiosn to propagate to console.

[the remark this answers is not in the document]
jiwenc-nv: I think @aristarkhovNV is advocating for try catch for the entire main. I don’t have strong opinion, but prefer we keep things consistent.

src/core/teleop_session_manager_tests/python/conftest.py
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Pytest configuration and fixtures for teleop_session_manager tests."""
+
+import pytest
```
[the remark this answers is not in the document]
jiwenc-nv: Is this file absolutely needed? Or it’s just copilot being consumed?

examples/teleop_session_manager/python/external_inputs_example.py
```diff
@@ -0,0 +1,296 @@
...
+            ee_input[0] = sim_ee_position
+
+            # Build TensorGroup for the transform_input leaf's "value" input
+            xform_input = TensorGroup(TransformMatrix())
+            xform_input[0] = vr_to_robot
+
+            # Call step() with external inputs for the non-DeviceIO leaves
+            result = session.step(external_inputs={
+                "sim_state": {"ee_state": ee_input},
+                "transform_input": {PassthroughInput.VALUE: xform_input},
```
jiwenc-nv: After looking at the example here at 2 mins, I’m still a bit confused about what does `PassthroughInput` do that is different from `ee_input`.

jiwenc-nv: After looking at this a bit longer, I think I maybe get it:

Is `PassthroughInput` all about reusing the code input node for arbitrary data type, so that a developer doesn’t need to write specific new `SourceNode` everything they want to pass in some data?

Assuming I’m getting this right, can we think about not implementing the other `SimStateSource`? We can just reuse something that’s already exiting, say the head pose source.

examples/teleop_session_manager/python/external_inputs_example.py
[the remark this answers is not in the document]
aristarkhovNV: I only advocate for C++, but in Python we can let the system handle uncaught exceptions which by default prints the stack trace (useful!) and all the exception info. This also helps to keep example code cleaner from boilerplate

examples/teleop_session_manager/python/external_inputs_example.py
aristarkhovNV: Maybe we could have a better name for `PassthroughInput` to better reflect the intended use case such as `StaticValue` or `ValueInput`

nvddr: You are correct, passthrough is an input node for an arbitrary data type. I think this example can be cleaned up. I wanted an example of an external input for a pre-defined node somewhere, and an example of how to use transforms. For transforms passthrough is necessary, for external input it could have been any node, the any node is what sim state is trying to do. ValueInput seems reasonable. Will update

aristarkhovNV: Could you elaborate on why passthrough is required for transforms?

src/core/teleop_session_manager_tests/python/test_teleop_session.py
```diff
@@ -0,0 +1,1282 @@
...
+
+Tests the TeleopSession class without requiring OpenXR hardware by mocking
+the hardware-dependent factory methods. Covers source discovery, external
+input validation, step execution, session lifecycle, and plugin management.
+"""
+
+import time
+import pytest
+from pathlib import Path
+from unittest.mock import MagicMock, patch, PropertyMock
```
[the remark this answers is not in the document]
nvddr: patch is used

examples/teleop_session_manager/python/external_inputs_example.py
nvddr: We could have defined a transform input. I defined this as a ValueInput since it's generally a useful mechansim,

PR #126, opened 2026-02-13
Handle numpy ABI to support both 1.x and 2.x

jiwenc-nv:
1. Build against numpy 2.x ABI
2. Declare numpy dependencies for numpy 1.x


CMakeLists.txt
```diff
@@ -95,6 +95,47 @@ if(WIN32)
     endif()
 endif()
 
+# When building Python bindings, extensions must be compiled against NumPy 2.x so a single
```
aristarkhovNV: can we merge this with SetupPython.cmake

jiwenc-nv: Makes sense. I can just be a macro from there. Will update this soon.

jiwenc-nv: Done.

PR #130, opened 2026-02-15
Update ControllerSnapshot to be an FBS table

jiwenc-nv:
This way, future APIs may spit out single controller instead of being forced to provide both controllers.
    
This also changes the following:
1. Rename is_active to is_valid, so that it's consistent with all other schemas.
2. Assign IDs to the fields in the schema, and make is_valid the first field (i.e. id: 0).
3. Update tests to reflect the new schema

### DIFF
--- a/src/core/deviceio/cpp/controller_tracker.cpp
+++ b/src/core/deviceio/cpp/controller_tracker.cpp
@@ -278,15 +278,15 @@ bool ControllerTracker::Impl::update(XrTime time)
         return false;
     }
 
-    // Helper to update a single controller - creates a new immutable struct snapshot
+    // Helper to update a single controller - creates a new ControllerSnapshotT (table)
     auto update_controller = [&](XrPath hand_path, const XrSpacePtr& grip_space, const XrSpacePtr& aim_space,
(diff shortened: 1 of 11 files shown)


src/core/deviceio/cpp/controller_tracker.cpp
```diff
@@ -360,13 +366,15 @@ Timestamp ControllerTracker::Impl::serialize(flatbuffers::FlatBufferBuilder& bui
...
     // Use left controller timestamp (or right if left is inactive)
-    if (controller_data_.left_controller && controller_data_.left_controller->is_active())
+    if (controller_data_.left_controller && controller_data_.left_controller->is_valid &&
+        controller_data_.left_controller->timestamp)
     {
-        return controller_data_.left_controller->timestamp();
+        return *controller_data_.left_controller->timestamp;
     }
-    else if (controller_data_.right_controller && controller_data_.right_controller->is_active())
+    else if (controller_data_.right_controller && controller_data_.right_controller->is_valid &&
```
aristarkhovNV, line 374: if controller is present, should the `!is_valid` state still be time stamped?

jiwenc-nv:
This is no longer needed. @nvddr has a uber clean up going on.

PR #137, opened 2026-02-17
Adding tunable params to se3 retargeter

rwiltz:


src/core/retargeting_engine/python/retargeters/se3_retargeter.py
```diff
@@ -42,16 +44,25 @@ class Se3RetargeterConfig:
     use_wrist_rotation: bool = False
     use_wrist_position: bool = True
 
-    # Abs specific
-    target_offset_pos: Tuple[float, float, float] = (0.0, 0.0, 0.0)
-    target_offset_rot: Tuple[float, float, float, float] = (0.7071068, 0.0, 0.0, 0.7071068) # x, y, z, w
+    # Abs specific - Position offset (meters)
+    target_offset_x: float = 0.0
```
nvddr, line 48: Can these be defined as arrays?

nvddr: optional, probably better this way for rot so one avoids confusion of xyzw ordering?

PR #138, opened 2026-02-17
add the ability to record metadata and raw data as mcap

shaosu-nvidia:
Note: This is only used within the oak camera plugin it self. It has added the ability to use the plugin without any cloudxr runtime as the dependency. We want the plugin to be self-contained to record both the frame metadata and the raw pixel at the same time.


src/plugins/oak/mcap/mcap_frame.fbs
```diff
@@ -0,0 +1,11 @@
+include "oak.fbs";
```
jiwenc-nv, line 1: nit: we need license header for fbs files as well.

PR #143, opened 2026-02-18
Use CloudXR Runtime SDK and Dockerfile

nv-mhaselton:
Download and build the CloudXR Runtime via SDK package following the pattern established in ./scripts/download_cloudxr_sdk.sh.

- Refactor README.md to clarify optional pre-download steps for CloudXR Runtime and Web SDKs.
- Modify GitHub Actions workflow to parse and utilize CloudXR SDK versions from .env.default.
- Implement new scripts for downloading CloudXR Runtime SDK and update existing scripts for better integration.
- Update Docker Compose configuration to build the CloudXR runtime image using the specified SDK version.
- Introduce a new Dockerfile for the CloudXR runtime to streamline the build process.
- Adjust environment variable paths in setup scripts for consistency.

### DIFF
--- a/deps/cloudxr/Dockerfile.runtime
+++ b/deps/cloudxr/Dockerfile.runtime
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-# Minimal CloudXR runtime image (~205 MB).
+# Minimal CloudXR runtime image.
 # The SDK must be downloaded first (see scripts/download_cloudxr_runtime_sdk.sh).
 #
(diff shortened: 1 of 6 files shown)


deps/cloudxr/Dockerfile.runtime
```diff
@@ -0,0 +1,151 @@
...
+int main(void) {
+    if (nv_cxr_service_create(&svc)) return 1;
+    signal(SIGINT, stop);
+    signal(SIGTERM, stop);
+    if (nv_cxr_service_start(svc)) { nv_cxr_service_destroy(svc); return 1; }
+    nv_cxr_service_join(svc);
+    nv_cxr_service_destroy(svc);
+    return 0;
+}
+EOF
```
nv-jakob: Would it be annoying to put this in an actual file inside of the repo? No worries if you prefer to lean it in the docker file.

Even larger pull would be to use a python script that interfaces to the runtime instead of a cpp file.

nv-mhaselton: sure, updated

PR #148, opened 2026-02-20
Build debug on main to keep caches warm

aristarkhovNV:


.github/workflows/build-ubuntu.yml
```diff
@@ -17,8 +17,7 @@ jobs:
 
     strategy:
       matrix:
-        # On tag or branch pushes, only build Release (Debug is skipped). On PRs, build both Debug and Release.
-        build_type: ${{ fromJSON((github.event_name == 'push') && '["Release"]' || '["Debug","Release"]') }}
+        build_type: ["Debug", "Release"]
```
jiwenc-nv, line 20: might worth adding a comment here that building debug for the main branch helps with cache hit.

PR #152, opened 2026-02-21
Multi stream and sensor check for oak camera

shaosu-nvidia:
1. This PR has updated to use the latest depthai sdk version to have the most stable behavior.
2. This PR makes oak camera support multiple streams, Color MonoLeft and MonoRight
3. Add a sensor check on the device. For oakdw camera, there are 2 color sensors that support different resolution
4. Add an optional device_id to match with GEAR stack. This is future supporting multiple oak camera device
5. some other clean up and fix, e.g. image size (use the full camera resolution.)

6. Maintain a schema pusher for each stream
7. the frame_metadata_tracker_oak will pull the metadata from all the streams by maintaining a schema_tracker for each.

### DIFF
--- a/examples/oxr/python/test_oak_camera.py
+++ b/examples/oxr/python/test_oak_camera.py
@@ -139,7 +139,7 @@ def run_test(duration: float = 10.0, metadata_track: bool = True):
                 with deviceio.DeviceIOSession.run(trackers, handles) as session:
                     print("  ✓ DeviceIO session initialized")
 
-                    # Create MCAP recorder with single OakMetadata channel
+                    # Create MCAP recorder with single CameraMetadataOak channel
                     mcap_entries = [(tracker, "oak_metadata")]
                     with mcap.McapRecorder.create(
                         mcap_filename,
--- a/src/core/schema/fbs/oak.fbs
+++ b/src/core/schema/fbs/oak.fbs
@@ -9,7 +9,7 @@ namespace core;
 enum StreamType : byte { Color = 0, MonoLeft = 1, MonoRight = 2 }
 
 // Per-frame metadata pushed by the OAK camera plugin (one per stream).
-table FrameMetadata {
+table FrameMetadataOak {
   // Which camera sensor this metadata belongs to.
   stream: StreamType (id: 0);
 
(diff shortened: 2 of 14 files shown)


examples/oxr/python/test_oak_camera.py
```diff
@@ -206,9 +214,11 @@ def run_test(duration: float = 10.0, metadata_track: bool = True):
     print()
     print("=" * 80)
     print("Test completed successfully!")
-    print(f"  Video: {PLUGIN_ROOT_DIR / 'oak_camera' / 'recordings'}")
+    recordings_dir = PLUGIN_ROOT_DIR / "oak_camera" / "recordings"
+    print(f"  Color H.264:    {recordings_dir / Path(color_path).name}")
+    print(f"  MonoLeft H.264: {recordings_dir / Path(mono_left_path).name}")
     if metadata_track:
-        print(f"  MCAP:  {Path.cwd() / mcap_filename}")
+        print(f"  MCAP:           {Path.cwd() / mcap_filename}")
```
qingsi-at-nv, line 229: Not sure if this indentation is intended.

cmake/SetupHunter.cmake
```diff
@@ -63,8 +63,8 @@ if(BUILD_PLUGINS)
     endif()
 
     # Download DepthAI Hunter config from GitHub if not present or empty
-    set(DEPTHAI_VERSION "v2.29.0")
-    set(DEPTHAI_CONFIG_SHA1 "a88698ab81b7edef79a2446edcf649dc92cddcbc")
+    set(DEPTHAI_VERSION "v2.32.0")
```
qingsi-at-nv: Just to double check if the version bump is intended.

shaosu-nvidia: not necessary reverted

examples/oxr/python/test_oak_camera.py
shaosu-nvidia: this is for better printing

src/core/schema/fbs/oak.fbs
```diff
@@ -5,12 +5,24 @@ include "timestamp.fbs";
...
-  timestamp: Timestamp (id: 0);
+  timestamp: Timestamp (id: 1);
+
+  // Per-stream frame sequence number from the device.
+  sequence_number: uint64 (id: 2);
+}
 
-  sequence_number: int (id: 1);
+// Composed metadata from all active OAK camera streams.
+table OakMetadata {
```
jiwenc-nv: nit: `FrameMetadataOak` which matches the same pattern of how we are naming vendor specific bits, e.g. `FullBodyPosePico`

shaosu-nvidia: Sounds good. Done

PR #159, opened 2026-02-22
Update docs workflow

jiwenc-nv:
1. Skip docs build if nothing changed in docs folder
3. Add release/** branches to the list of branches to build docs for

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -9,9 +9,14 @@ name: Build & deploy docs
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - 'release/**'
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
+      - "docs/**"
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yaml
```diff
@@ -9,9 +9,17 @@ name: Build & deploy docs
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - 'release/**'
+    paths:
+      - "docs/**"
+      - ".github/workflows/docs.yaml"
```
aristarkhovNV: we don't do releases very often, lets always build just in case

jiwenc-nv: good point. I just need to reduce the PR preview noise.

aristarkhovNV: https://github.com/NVIDIA/IsaacTeleop/pull/161 removes the noise completely

PR #162, opened 2026-02-23
Create SchemaRecord types for storage

nvddr:


src/core/schema/fbs/hand.fbs
```diff
@@ -36,4 +36,9 @@ table HandPose {
     timestamp: Timestamp (id: 2);
 }
 
-root_type HandPose;
+// Record wrapper for a single hand pose (used as root type for serialization).
+table HandPoseRecord {
+    data: HandPose (id: 0);
+}
```
aristarkhovNV, line 46: Is the purpose of the wrapper so that the data is nullable?

nvddr: It will eventually also contain the timestamp. The following PR, yes the intent is so that it is nullable. Might contain other internal metadata in the future (e.g. timestamps for certain trackers if we think its useful).

src/core/schema/fbs/pedals.fbs
```diff
@@ -24,4 +24,9 @@ table Generic3AxisPedalOutput {
   rudder: float (id: 4);
 }
 
-root_type Generic3AxisPedalOutput;
+// Record wrapper for a single pedal output (used as root type for serialization).
+table Generic3AxisPedalOutputRecord {
```
jiwenc-nv, line 32: maybe the original schema was poorly named as xxxOutput. Since we are here, can you rename it to `Generic3AxisPedalSnapshot` so that it's consistent with controller?

src/core/schema/fbs/controller.fbs
```diff
@@ -47,9 +47,9 @@ struct ControllerSnapshot {
     timestamp: Timestamp;
 }
 
-// Controller data wrapper for a single controller (table type for top-level serialization).
-table ControllerData {
+// Record wrapper for a single controller snapshot (used as root type for serialization).
+table ControllerSnapshotRecord {
```
jiwenc-nv, line 55: not sure where is the best place to document this.  But we need to clearly document the naming convention and how they are intended to be used.

```
// This is the base schema.
// Use it when xxx.
table DataSchmeForXyz {
}

// This is wrapper for a data record when 
table DataSchmeForXyzRecord {

  // always having this set of timestamps.
}

// This is wrapper for output data recording
table DataSchmeForXyzTracked {
  // always having that set of timestamps.
}
```

I think `...Record` and `...Tracked` having the benefits of being concise, but may need a bit documentation to boost clarity.

src/core/schema/fbs/oak.fbs
```diff
@@ -13,4 +13,9 @@ table FrameMetadata {
   sequence_number: int (id: 1);
 }
 
-root_type FrameMetadata;
+// Record wrapper for a single frame metadata (used as root type for serialization).
+table FrameMetadataRecord {
```
jiwenc-nv: rebase conflict alert:
https://github.com/NVIDIA/IsaacTeleop/pull/152/changes#diff-c73d1f7a8eec46fc642f86cbe627d5f42005dff517b775698cb80af498b5ccd0

src/core/schema_tests/cpp/test_full_body.cpp
```diff
@@ -207,7 +209,7 @@ TEST_CASE("FullBodyPosePicoT serialization and deserialization", "[full_body][fl
 
     // Deserialize.
     auto buffer = builder.GetBufferPointer();
-    auto deserialized = core::GetFullBodyPosePico(buffer);
+    auto deserialized = flatbuffers::GetRoot<core::FullBodyPosePico>(buffer);
```
jiwenc-nv, line 212: when using `flatbuffers::GetRoot`, the `FullBodyPosePico` doesn't need to be marked as `root_type` in the `fbs`?

src/core/schema/fbs/pedals.fbs
nvddr: Folllow-up if absolutely critical

src/core/schema_tests/cpp/test_full_body.cpp
nvddr: No, this works. Basically if you don't declare root_type, you don't get the core::GetFullBodyPosePico accessor, but you can still use GetRoot, and then use it the same way.

src/core/schema/fbs/controller.fbs
nvddr: Done, added documengtation

PR #165, opened 2026-02-24
Bundle CloudXR runtime into TeleopCore

jiwenc-nv:


.github/workflows/build-ubuntu.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     # Pin to 22.04 so the wheel is built with GCC 11's libstdc++. The resulting wheel
     # then runs on both Ubuntu 22.04 and 24.04 (24.04's libstdc++ is backward compatible).
     runs-on: ${{ matrix.arch == 'arm64' && 'ubuntu-22.04-arm' || 'ubuntu-22.04' }}
+    environment: dev
```
aristarkhovNV, line 17: now every build will be considered deployment. We should perhaps move NGC secrets out of dev to the general GH secrets.

.github/actions/setup-cloudxr-sdk/action.yml
```diff
@@ -0,0 +1,67 @@
...
+    value: ${{ steps.parse.outputs.web_version }}
+
+runs:
+  using: composite
+  steps:
+    - name: Parse CloudXR SDK versions from .env.default
+      id: parse
+      shell: bash
+      run: |
+        CXR_RUNTIME_SDK_VERSION=$(grep '^CXR_RUNTIME_SDK_VERSION=' deps/cloudxr/.env.default | cut -d= -f2)
```
aristarkhovNV: could we get https://github.com/NVIDIA/IsaacTeleop/pull/155 in first? It hardens error handling in these scripts

src/core/cloudxr/python/runtime.py
```diff
@@ -0,0 +1,143 @@
...
+import ctypes
+import os
+import shutil
+import signal
+import sys
+import threading
+
+
+_EULA_URL = (
+    "https://github.com/NVIDIA/IsaacTeleop/blob/main/deps/cloudxr/CLOUDXR_LICENSE"
```
nv-mhaselton, line 12: Note there is also public pdf version @ https://developer.download.nvidia.com/cloudxr/EULA/NVIDIA_CloudXR_GA_License_without_Data_Collection_25Feb2025.pdf

.github/workflows/build-ubuntu.yml
```diff
@@ -121,48 +128,11 @@ jobs:
...
-      env:
-        CXR_WEB_SDK_VERSION: ${{ steps.cloudxr.outputs.web_version }}
-        NGC_CLI_API_KEY: ${{ secrets.NGC_TELEOP_CORE_GITHUB_SERVICE_KEY }}
-      run: |
-        ./scripts/download_cloudxr_sdk.sh
+    - name: Setup CloudXR SDK
+      id: cloudxr-sdk
+      uses: ./.github/actions/setup-cloudxr-sdk
+      with:
+        ngc-cli-api-key: ${{ secrets.NGC_TELEOP_CORE_GITHUB_SERVICE_KEY }}
```
aristarkhovNV, line 135: now that cloudxr is packaged with isaacteleop we no longer need to pull it from NGC here, use the packaged version instead

.github/actions/setup-cloudxr-sdk/action.yml
jiwenc-nv: done.

.github/workflows/build-ubuntu.yml
jiwenc-nv: We are still using the SDK for now to build the docker containers. I will change that in a follow up PR.

.github/workflows/build-ubuntu.yml
jiwenc-nv: but then we are not gating the deployment? (i.e. access to the SDK?)

aristarkhovNV: PRs from forks will not have access to repo secrets by default, so should be safe to move secrets outside of environment, keep the environment for actual deployments to Artifactory and Kitmaker

src/core/cloudxr/python/runtime.py
jiwenc-nv: Yep. I'm aware.

PR #166, opened 2026-02-24
Make the top-level data containers struct for cleaner access patterns.

nvddr:


jiwenc-nv:
@nvddr you don't need this one anymore, right? seems the changes are merged in other PRs already.

PR #173, opened 2026-02-25
Make camera tracker use multiple channels

nvddr:


src/core/deviceio/cpp/frame_metadata_tracker_oak.cpp
shaosu-nvidia, line 1: do we want to change the file name or frame_metadata_tracker_oak is fine?

nvddr: I think that's okay.

PR #174, opened 2026-02-25
Add a flag to only build ROS2 RA

life1ess:


CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ endif()
 # Note: BUILD_PLUGINS is defined in cmake/SetupHunter.cmake before project()
 # Note: BUILD_PYTHON_BINDINGS is defined in cmake/SetupPython.cmake
 option(BUILD_EXAMPLES "Build examples" ON)
+option(BUILD_EXAMPLE_TELEOP_ROS2 "Build only the teleop_ros2 example (e.g. for Docker)" OFF)
```
jiwenc-nv, line 62: why not default to be ON?

life1ess: It's already included in BUILD_EXAMPLES that is on by default. I think it's better not to build teleop_ros2 if BUILD_EXAMPLES is OFF unless it's explicitly enabled by BUILD_EXAMPLE_TELEOP_ROS2

PR #175, opened 2026-02-25
Safely discard external OpenXR handles and resources on Teleop close

hougantc-nvda:
Summary
This PR adds a safe shutdown path for externally-owned OpenXR lifecycles (e.g., Kit-managed sessions) to prevent crashes when Teleop/DeviceIO outlives runtime teardown. It introduces explicit resource-discard behavior so tracker and session destructors do not call xrDestroy* on handles that were already invalidated by the external runtime.

Why
When OpenXR handles are passed from an external owner, that owner may tear down runtime resources first. Without explicit discard, Teleop/DeviceIO destructors can attempt cleanup against stale handles, leading to crashes. This PR makes teardown ownership explicit and provides a safe close path for externally-invalidated runtimes.


src/core/deviceio/python/deviceio_bindings.cpp
```diff
@@ -100,6 +100,8 @@ PYBIND11_MODULE(_deviceio, m)
     // PyDeviceIOSession& directly, calling .native() internally in C++ code.
     py::class_<PyDeviceIOSession>(m, "DeviceIOSession")
         .def("update", &PyDeviceIOSession::update, "Update session and all trackers")
+        .def("_discard_oxr_handles", &PyDeviceIOSession::discard_oxr_resources,
+             "Discard XR handle ownership so destructors do not call xrDestroy* (internal: call when runtime was invalidated externally, e.g. Stop XR).")
```
aristarkhovNV, line 104: This is gross hack that throws RAII out the window. Would it be possible to let isaacteleop know before the owner destroy its xr session, not after the fact? Then isaacteleop would properly tear down its resources without leaving dangling objects.

jiwenc-nv: I think the issue with sharing session with Kit. When "Step XR" is hit, the XrInstance and session went away there.

PR #180, opened 2026-02-26
Fix multiple controller trackers using temporary workaround.

nvddr:
Multiple controller trackers did not work prior to this, because we cannot have multiple action sets. 

This sets up a temporary workaround that caches tracker implementation instances and reuses them across different controller trackers, and also avoids duplicate updates for multiple controller tracker instances.


src/core/python/requirements-retargeters-lite.txt
```diff
@@ -0,0 +1 @@
+scipy>=1.15.0
```
jiwenc-nv, line 1: Add a TODO here to state that this is temporary and the plan is to move actual retargeters implementations out of src/core.

nvddr: I think that can be handled when we actually move it out, not necessary here as this particular approach is still reasonable by itself.

PR #186, opened 2026-02-26
enable build oak for github ci

shaosu-nvidia:
## Summary by CodeRabbit

* **Chores**
  * Updated build workflows for Ubuntu and Windows to enable building the Oak camera plugin.
  * Updated copyright year to 2025–2026.


jiwenc-nv (approved):
LGTM.

If CI becomes too slow, we might come back asking for only doing that when OakD plugin is changed.  Currently we don't have a very elegant way to doing it.

PR #222, opened 2026-03-05
Use controller aim pose if hand tracking is not available

life1ess:
## Summary by CodeRabbit

* **New Features**
  * Publishes AIM-based hand pose data for both left and right controllers alongside existing wrist-based poses, with automatic fallback when AIM data is unavailable.

* **Improvements**
  * Simplified controller data handling by removing redundant operations while preserving hand arrays, root pose/twist, and full-body data publishing.

### DIFF
--- a/examples/teleop_ros2/python/teleop_ros2_publisher.py
+++ b/examples/teleop_ros2/python/teleop_ros2_publisher.py
@@ -290,6 +290,8 @@ class TeleopRos2PublisherNode(Node):
 
                         left_hand = result["hand_left"]
                         right_hand = result["hand_right"]
+                        left_ctrl = result["controller_left"]
+                        right_ctrl = result["controller_right"]
 
                         if not right_hand.is_none:
                             right_positions = np.asarray(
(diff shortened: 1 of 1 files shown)


examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -304,6 +306,22 @@ def run(self) -> int:
...
+                                    np.asarray(
+                                        right_ctrl[ControllerInputIndex.AIM_POSITION]
+                                    ),
+                                    np.asarray(
+                                        right_ctrl[
+                                            ControllerInputIndex.AIM_ORIENTATION
+                                        ]
+                                    ),
+                                )
+                            )
```
kajananchinniahNV, line 319: We'd need to do tf publishing on this branch as well.

It might be cleaner if we did something along the lines of:
```
left_hand = result["hand_left"]
right_hand = result["hand_right"]
left_ctrl = result["controller_left"]
right_ctrl = result["controller_right"]

# convert to pose encapsulates the two if conditions 
left_pose = convert_to_pose(left_hand, left_ctrl)
right_pose = convert_to_pose_right_hand, right_ctrl)

# continue
if left_pose: 
```

But I don't mind delaying to a refactoring pass after we get the rest of the features we need in

examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -304,6 +306,20 @@ def run(self) -> int:
                                     right_orientations[HandJointIndex.WRIST],
                                 )
                             )
+                        elif (
```
jiwenc-nv: actually, do we want to prefer controller if both are available?

will controller be more accurate and lower latency vs hand tracking?

kajananchinniahNV (approved):

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: Hand tracking is not just EE poses, is includes other joints, so while it may be less accurate and fast, it provides more data

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: Let's delay this to a refactoring pass

PR #225, opened 2026-03-06
Introduce ROS teleop mode

life1ess:
## Summary by CodeRabbit

* **New Features**
  * Added a mode parameter to select four teleoperation scenarios (controller_teleop, hand_teleop, controller_raw, full_body). Publishing now switches per mode and emits only relevant topics, including explicit root twist/root pose and an optional full-body command stream. Introduced a unified ee_poses topic and a dedicated end-effector topic for controller EE poses.

* **Documentation**
  * README updated with new mode descriptions, renamed topic/parameter names (ee_pose_topic, root_twist_topic, root_pose_topic, mode), a published-topics table, and an example invocation.

### DIFF
--- a/examples/teleop_ros2/python/teleop_ros2_publisher.py
+++ b/examples/teleop_ros2/python/teleop_ros2_publisher.py
@@ -406,23 +406,24 @@ class TeleopRos2PublisherNode(Node):
                                 self._pub_ee_pose.publish(ee_msg)
 
                         if self._mode in ("hand_teleop", "controller_teleop"):
-                            root_command = result["root_command"]
-                            cmd = np.asarray(root_command[0])
-                            twist_msg = TwistStamped()
-                            twist_msg.header.stamp = now
-                            twist_msg.header.frame_id = self._frame_id
-                            twist_msg.twist.linear.x = float(cmd[0])
-                            twist_msg.twist.linear.y = float(cmd[1])
-                            twist_msg.twist.linear.z = 0.0
-                            twist_msg.twist.angular.z = float(cmd[2])
-                            self._pub_root_twist.publish(twist_msg)
+                            root_command = result.get("root_command")
+                            if not root_command.is_none:
+                                cmd = np.asarray(root_command[0])
+                                twist_msg = TwistStamped()
+                                twist_msg.header.stamp = now
+                                twist_msg.header.frame_id = self._frame_id
+                                twist_msg.twist.linear.x = float(cmd[0])
+                                twist_msg.twist.linear.y = float(cmd[1])
+                                twist_msg.twist.linear.z = 0.0
+                                twist_msg.twist.angular.z = float(cmd[2])
+                                self._pub_root_twist.publish(twist_msg)
 
-                            pose_msg = PoseStamped()
-                            pose_msg.header.stamp = now
-                            pose_msg.header.frame_id = self._frame_id
-                            pose_msg.pose.position.z = float(cmd[3])
-                            pose_msg.pose.orientation.w = 1.0
-                            self._pub_root_pose.publish(pose_msg)
+                                pose_msg = PoseStamped()
+                                pose_msg.header.stamp = now
+                                pose_msg.header.frame_id = self._frame_id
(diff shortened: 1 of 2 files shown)


examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -284,114 +331,125 @@ def run(self) -> int:
...
+                                hand_msg.poses.append(
+                                    _to_pose(
+                                        left_positions[HandJointIndex.WRIST],
+                                        left_orientations[HandJointIndex.WRIST],
+                                    )
+                                )
+                            else:
+                                hand_msg.poses.append(_to_pose([0.0, 0.0, 0.0]))
+
+                            if not right_hand.is_none:
```
kajananchinniahNV: Should we consider also making adding the finger joints optional as another parameter in the PoseArray or should we always want to keep it?

examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -284,114 +331,125 @@ def run(self) -> int:
...
-                                    ),
-                                    np.asarray(
-                                        right_ctrl[ControllerInputIndex.AIM_ORIENTATION]
-                                    ),
+                        hand_msg = None
+                        if self._mode == "hand_teleop":
+                            hand_msg = PoseArray()
+                            hand_msg.header.stamp = now
+                            hand_msg.header.frame_id = self._frame_id
+                            left_hand = result["hand_left"]
```
kajananchinniahNV: Could we move the logic for hand_teleop in a function similar to what we did for controller_teleop?

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: I don't think they should be optional when there's no retargeting - most downstream components will need all hand poses

life1ess: When we add retargeting, it's reasonable to only send wrist poses when retargeting is on

kajananchinniahNV: Got it, and we definitely want to keep it in the same topic too? I think I'd prefer the topic to be consistent wrt output between the two modes (e.g. we could have /xr_teleop/finger_poses)

examples/teleop_ros2/python/teleop_ros2_publisher.py
life1ess: Good idea, done

examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -284,114 +381,67 @@ def run(self) -> int:
...
+                            if not left_ctrl.is_none or not right_ctrl.is_none:
+                                controller_payload = _build_controller_payload(
+                                    left_ctrl, right_ctrl
+                                )
+                                payload = msgpack.packb(
+                                    controller_payload, default=mnp.encode
+                                )
+                                payload = tuple(bytes([a]) for a in payload)
+                                controller_msg = ByteMultiArray()
+                                controller_msg.data = payload
```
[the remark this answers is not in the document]
life1ess: No it won't fail

examples/teleop_ros2/python/teleop_ros2_publisher.py
```diff
@@ -284,114 +381,67 @@ def run(self) -> int:
...
+                            twist_msg.twist.linear.z = 0.0
+                            twist_msg.twist.angular.z = float(cmd[2])
+                            self._pub_root_twist.publish(twist_msg)
+
+                            pose_msg = PoseStamped()
+                            pose_msg.header.stamp = now
+                            pose_msg.header.frame_id = self._frame_id
+                            pose_msg.pose.position.z = float(cmd[3])
+                            pose_msg.pose.orientation.w = 1.0
+                            self._pub_root_pose.publish(pose_msg)
```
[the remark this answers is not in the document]
life1ess: Done

PR #242, opened 2026-03-11
Proxy shows a friendlier page when certificate is accepted

blongs-nv:
For both the docker and python wss proxies, instead of a 501 error page this will show a friendlier page when a self-signed certificate is accepted:
image

## Release Notes

* **Improvements**
  * Non-WebSocket HTTP requests now receive a direct certificate-accepted confirmation response, eliminating unnecessary backend forwarding operations.
  * Updated request handling infrastructure for improved system efficiency and cleaner resource utilization.
  * WebSocket proxy functionality remains fully operational and stable.

### DIFF
--- a/src/core/cloudxr/python/wss.py
+++ b/src/core/cloudxr/python/wss.py
@@ -5,7 +5,6 @@
 
 import asyncio
 import errno
-import http.client
 import logging
 import shutil
 import ssl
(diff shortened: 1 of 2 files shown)


src/core/cloudxr/python/wss.py
```diff
@@ -112,56 +111,6 @@ def build_ssl_context(cert_paths: CertPaths) -> ssl.SSLContext:
 }
 
 
-def _forward_http(backend_host, backend_port, request):
```
jiwenc-nv, line 115: Nice. Cursor was talking about this as well.  Thanks for the clean up.

PR #253, opened 2026-03-12
Fix CloudXR runtime crash from OpenSSL symbol conflict

rwiltz:
Load libcloudxr.so with RTLD_DEEPBIND so NVST resolves OpenSSL symbols from its own scope instead of picking up Python's incompatible libssl.so.3 from the global scope.
Add a patchelf build step to strip the spurious libssl.so.3 NEEDED entry from libcloudxr.so after extracting the SDK tarball, since the library never calls any OpenSSL functions directly.

### DIFF
--- a/src/core/cloudxr/python/CMakeLists.txt
+++ b/src/core/cloudxr/python/CMakeLists.txt
@@ -78,6 +78,24 @@ else()
         COMMENT "Extracting CloudXR SDK tarball into python_package/.../cloudxr/native/"
     )
     add_dependencies(cloudxr_native_bundle cloudxr_native_dir)
+
+    # libcloudxr.so in the SDK tarball carries a NEEDED entry for libssl.so.3 even
+    # though it never calls any OpenSSL functions. When Python (which bundles its own
+    # OpenSSL) loads libcloudxr.so, the loader pulls in a second, incompatible
+    # libssl.so.3 and NVST crashes. Strip the spurious dependency at build time.
+    # A future SDK built with -Wl,--as-needed will make this step unnecessary.
+    find_program(PATCHELF_EXECUTABLE patchelf)
+    if(PATCHELF_EXECUTABLE)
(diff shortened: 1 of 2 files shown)


src/core/cloudxr/python/CMakeLists.txt
```diff
@@ -78,6 +78,24 @@ else()
...
+
+    # libcloudxr.so in the SDK tarball carries a NEEDED entry for libssl.so.3 even
+    # though it never calls any OpenSSL functions. When Python (which bundles its own
+    # OpenSSL) loads libcloudxr.so, the loader pulls in a second, incompatible
+    # libssl.so.3 and NVST crashes. Strip the spurious dependency at build time.
+    # A future SDK built with -Wl,--as-needed will make this step unnecessary.
+    find_program(PATCHELF_EXECUTABLE patchelf)
+    if(PATCHELF_EXECUTABLE)
+        add_custom_target(cloudxr_patchelf ALL
+            COMMAND ${PATCHELF_EXECUTABLE} --remove-needed libssl.so.3
```
aristarkhovNV, line 90: What if instead of stripping the dependency we add `libssl.so` to the ignore list at the wheel repair step at https://github.com/NVIDIA/IsaacTeleop/blob/main/.github/workflows/build-ubuntu.yml#L118? That might be a safer option in case libssl is actually pulled for something

rwiltz: Only doing this didn't work. Looking into it with Cursor:



> Removing the NEEDED entry with patchelf works because NVST's internal code then loads its own compatible OpenSSL through the SDK's bundled copy (via LD_LIBRARY_PATH or its own dlopen), bypassing conda's entirely.

rwiltz: making a lot of assumptions but it works for now

PR #283, opened 2026-03-16
Clean up trackers

nvddr:
## Summary by CodeRabbit

* **Refactor**
  * Reorganized tracker implementations to separate declarations from implementations and centralize initialization/serialization.

* **Bug Fixes / Reliability**
  * More robust startup, error handling, and state clearing for consistent tracker outputs and safer resource cleanup.

* **New Features**
  * Exposed schema-text accessors for head, hand, and full-body trackers and switched serialization to use monotonic timestamps.

### DIFF
--- a/src/core/deviceio/cpp/inc/deviceio/head_tracker.hpp
+++ b/src/core/deviceio/cpp/inc/deviceio/head_tracker.hpp
@@ -5,9 +5,6 @@
 
 #include "tracker.hpp"
 
-#include <oxr_utils/oxr_funcs.hpp>
-#include <oxr_utils/oxr_time.hpp>
-#include <schema/head_bfbs_generated.h>
 #include <schema/head_generated.h>
 
 #include <memory>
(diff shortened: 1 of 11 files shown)


src/core/deviceio/cpp/inc/deviceio/head_tracker.hpp
```diff
@@ -52,26 +40,7 @@ class HeadTracker : public ITracker
...
-
-    private:
-        const OpenXRCoreFunctions core_funcs_;
-        XrTimeConverter time_converter_;
-        XrSpace base_space_;
-        XrSpacePtr view_space_;
-        HeadPoseTrackedT tracked_;
-        XrTime last_update_time_ = 0;
-    };
+    class Impl;
```
aristarkhovNV, line 43: btw the impl could be a standalone class, does not have to be a member of the tracker

aristarkhovNV (approved):

src/core/deviceio/cpp/inc/deviceio/head_tracker.hpp
nvddr: Having it in the class gives it a namespace, but yeah it could be a completely different class, I don't really have opinions..

PR #287, opened 2026-03-17
Bypass body tracker if not supported by the runtime

nvddr:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for device initialization and tracking scenarios. The system now gracefully handles missing components and continues operation with limited functionality instead of failing completely.
  * Enhanced robustness during body tracker setup and joint detection, with better fallback behavior when tracking data is unavailable.

### DIFF
--- a/src/core/deviceio/cpp/full_body_tracker_pico.cpp
+++ b/src/core/deviceio/cpp/full_body_tracker_pico.cpp
@@ -71,7 +71,9 @@ FullBodyTrackerPico::Impl::Impl(const OpenXRSessionHandles& handles)
         }
         if (!body_tracking_props.supportsBodyTracking)
         {
-            throw std::runtime_error("Body tracking not supported by this system");
+            std::cerr << "[FullBodyTrackerPico] Body tracking not supported by this system, running in limp mode"
+                      << std::endl;
+            return;
         }
     }
     else
@@ -85,11 +87,6 @@ FullBodyTrackerPico::Impl::Impl(const OpenXRSessionHandles& handles)
                           reinterpret_cast<PFN_xrVoidFunction*>(&pfn_destroy_body_tracker_));
     loadExtensionFunction(handles.instance, handles.xrGetInstanceProcAddr, "xrLocateBodyJointsBD",
                           reinterpret_cast<PFN_xrVoidFunction*>(&pfn_locate_body_joints_));
-
-    if (!pfn_create_body_tracker_ || !pfn_destroy_body_tracker_ || !pfn_locate_body_joints_)
-    {
-        throw std::runtime_error("Failed to get body tracking function pointers");
-    }
 
     XrBodyTrackerCreateInfoBD create_info{ XR_TYPE_BODY_TRACKER_CREATE_INFO_BD };
     create_info.next = nullptr;
@@ -106,10 +103,9 @@ FullBodyTrackerPico::Impl::Impl(const OpenXRSessionHandles& handles)
 
 FullBodyTrackerPico::Impl::~Impl()
 {
-    assert(pfn_destroy_body_tracker_ != nullptr && "pfn_destroy_body_tracker must not be null");
-
     if (body_tracker_ != XR_NULL_HANDLE)
     {
+        assert(pfn_destroy_body_tracker_ != nullptr && "pfn_destroy_body_tracker must not be null");
         pfn_destroy_body_tracker_(body_tracker_);
         body_tracker_ = XR_NULL_HANDLE;
     }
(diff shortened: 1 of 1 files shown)


src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -216,12 +216,28 @@ std::string_view FullBodyTrackerPico::get_schema_text() const
...
+    try
+    {
+        return std::make_shared<Impl>(handles);
+    }
+    catch (const std::runtime_error& e)
+    {
+        std::cerr << "[FullBodyTrackerPico] Body tracking unavailable: " << e.what() << std::endl;
+        return nullptr;
+    }
 }
```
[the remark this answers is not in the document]
nvddr: This is intended, as others will succeed, but body tracker may not if body tracker is absent. Head tracker will succeed if hmd doesnt support head tracking anyway.

src/core/deviceio/cpp/deviceio_session.cpp
```diff
@@ -19,10 +19,25 @@ DeviceIOSession::DeviceIOSession(const std::vector<std::shared_ptr<ITracker>>& t
...
 {
-    // Initialize all trackers and collect their implementations
+    // Initialize all trackers and collect their implementations (skip trackers that fail to create)
     for (const auto& tracker : trackers)
     {
-        tracker_impls_.emplace(tracker.get(), tracker->create_tracker(handles_));
+        if (!tracker)
+        {
+            throw std::invalid_argument("DeviceIOSession: null tracker in trackers list");
+        }
```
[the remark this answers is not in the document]
nvddr: This is fine, the tracker implementation creation fails. Ideally tracker being null is bad input, and the other null check was jsut defensive.

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -98,17 +100,16 @@ FullBodyTrackerPico::Impl::Impl(const OpenXRSessionHandles& handles)
...
 
     std::cout << "FullBodyTrackerPico initialized (24 joints)" << std::endl;
 }
 
 FullBodyTrackerPico::Impl::~Impl()
 {
-    assert(pfn_destroy_body_tracker_ != nullptr && "pfn_destroy_body_tracker must not be null");
-
-    if (body_tracker_ != XR_NULL_HANDLE)
+    if (body_tracker_ != XR_NULL_HANDLE && pfn_destroy_body_tracker_ != nullptr)
```
aristarkhovNV: if we managed to create the tracker then there must be the destroy function. We can enforce it in ctor and assert here

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -88,7 +89,8 @@ FullBodyTrackerPico::Impl::Impl(const OpenXRSessionHandles& handles)
 
     if (!pfn_create_body_tracker_ || !pfn_destroy_body_tracker_ || !pfn_locate_body_joints_)
     {
-        throw std::runtime_error("Failed to get body tracking function pointers");
+        std::cerr << "[FullBodyTrackerPico] Failed to get body tracking function pointers, limp mode" << std::endl;
+        return;
```
aristarkhovNV: if we end up here while `body_tracking_props.supportsBodyTracking == true` then something bad is happening, better throw than ignore.

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -58,25 +58,26 @@ FullBodyTrackerPico::Impl::Impl(const OpenXRSessionHandles& handles)
     system_info.formFactor = XR_FORM_FACTOR_HEAD_MOUNTED_DISPLAY;
 
     XrResult result = core_funcs.xrGetSystem(handles.instance, &system_info, &system_id);
-    if (XR_SUCCEEDED(result))
+    if (XR_FAILED(result))
+    {
+        std::cerr << "[FullBodyTrackerPico] OpenXR get system failed, running in limp mode: " << result << std::endl;
```
aristarkhovNV: this should be an actual failure, seems unrelated to body tracking

src/core/deviceio/cpp/full_body_tracker_pico.cpp
```diff
@@ -58,25 +58,26 @@ FullBodyTrackerPico::Impl::Impl(const OpenXRSessionHandles& handles)
...
-        result = core_funcs.xrGetSystemProperties(handles.instance, system_id, &system_props);
-        if (XR_FAILED(result))
-        {
-            throw std::runtime_error("OpenXR: failed to get system properties: " + std::to_string(result));
-        }
-        if (!body_tracking_props.supportsBodyTracking)
-        {
-            throw std::runtime_error("Body tracking not supported by this system");
-        }
+        std::cerr << "[FullBodyTrackerPico] OpenXR get system properties failed, limp mode: " << result << std::endl;
```
aristarkhovNV: same here, the limp mode should be decided per `body_tracking_props.supportsBodyTracking`, not any arbitrary failed call

src/core/deviceio/cpp/full_body_tracker_pico.cpp
aristarkhovNV: btw `loadExtensionFunction` throws on error, this check is redundant

PR #313, opened 2026-03-24
Move get_required_extensions to the session level 

nvddr:
Change get_required_extensions to become a session implementation detail (i.e openxr backed live session).

## Summary by CodeRabbit

* **Refactor**
  * Session-level extension discovery centralized in the live backend; per-tracker extension-reporting and per-tracker factory hooks removed, and tracker implementation creation consolidated.

* **Documentation**
  * Guides, examples, and API docs updated to describe session-level extension aggregation and the revised live-backend registration/creation workflow.

* **Tests**
  * Tests updated to verify platform-specific baseline extensions, deduplication/sorting, and session-level aggregation behavior; example demos/docstrings adjusted accordingly.


jiwenc-nv (approved):

src/core/deviceio_session/cpp/deviceio_session.cpp
aristarkhovNV, line 79: this is now redundant, `create_tracker_impl` itself will throw in case of error

PR #323, opened 2026-03-25
Add custom bt601 full range cuda kernel and fix color conversion

farbod-nv:
Nvenc color conversion is CSC so for custom encoding we need csc nv12->rgb
On the other hand, Oak-d uses bt601 fullrange color conversion ondevice for rgb->nv12 which is not supported by nppi. This adds a custom cuda kernel to do the correct color conversion for oakd. Alternative is to use 2 nppi methods (go to yuv420 then rgb) which is not as efficient as this.

## Summary by CodeRabbit

* **New Features**
  * Added GPU-accelerated full-range BT.601 NV12→RGB conversion option for decoded video.
  * Added a force_full_range toggle to the stream decoder API to override automatic range detection.
  * Per-camera color_range setting ("auto"/"full"/"limited"); "auto" resolves per camera type (OAKD defaults to full-range) and is passed to the decoder.

### DIFF
--- a/examples/camera_streamer/operators/nv_stream_decoder/nv_stream_decoder_op.cpp
+++ b/examples/camera_streamer/operators/nv_stream_decoder/nv_stream_decoder_op.cpp
@@ -58,6 +58,9 @@ void NvStreamDecoderOp::initialize()
     cuda_check(cuDevicePrimaryCtxRetain(&cu_context_, cu_device_));
 
     // Initialize NPP stream context manually.
+    // Push the target device context so the CUDA runtime API calls below
+    // query the correct GPU (matters on multi-GPU systems).
+    cuda_check(cuCtxPushCurrent(cu_context_));
     {
         npp_ctx_.hStream = 0; // Default (NULL) stream.
 
@@ -78,6 +81,7 @@ void NvStreamDecoderOp::initialize()
         npp_ctx_.nMaxThreadsPerBlock = deviceProperties.maxThreadsPerBlock;
         npp_ctx_.nSharedMemPerBlock = deviceProperties.sharedMemPerBlock;
     }
+    cuda_check(cuCtxPopCurrent(nullptr));
 
     if (verbose_.get())
     {
@@ -215,6 +219,10 @@ void NvStreamDecoderOp::compute(holoscan::InputContext& op_input,
 
     auto dst = static_cast<uint8_t*>(out_tensor.value()->pointer());
 
+    // Push the decoder's CUDA context so the conversion runs on the correct
+    // GPU.  pFrame and dst reside on cu_device_; without this, multi-GPU
+    // setups would target the wrong device after the decode context pop above.
+    cuda_check(cuCtxPushCurrent(cu_context_));
     if (use_full_range_)
     {
         // BT.601 full-range (ITU-T T.871).  NPP has no NV12 variant for this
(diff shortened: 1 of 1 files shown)


examples/camera_streamer/teleop_camera_subgraph.py
```diff
@@ -485,12 +485,16 @@ def _compose_rtp_sources(
                     port=stream_cfg.port,
                     verbose=verbose,
                 )
+                # OAK-D VPU encoder produces full-range NV12; the default
+                # limited-range BT.709 conversion causes over-saturated colors.
+                full_range = cam_cfg.camera_type == "oakd"
```
nvddr: Can we make this more extensible as you add mode cameras? Maybe have a per-camera settings profile and this reads from that.

farbod-nv: good idea. added

examples/camera_streamer/operators/nv_stream_decoder/nv_stream_decoder_op.cpp
```diff
@@ -196,15 +222,25 @@ void NvStreamDecoderOp::compute(holoscan::InputContext& op_input,
...
-    NppStatus status = nppiNV12ToRGB_8u_P2C3R_Ctx(pSrc, pitch, dst, width * 3, roi, npp_ctx_);
-    if (status != NPP_SUCCESS)
+    if (use_full_range_)
     {
-        HOLOSCAN_LOG_ERROR("NPP NV12->RGB failed: {}", static_cast<int>(status));
-        decoder_->UnlockFrame(&pFrame);
-        return;
+        // BT.601 full-range (ITU-T T.871).  NPP has no NV12 variant for this
+        // combination so we use a single-pass CUDA kernel.  See nv12_to_rgb.cu.
+        nv12_to_rgb_fullrange_bt601(pFrame, pFrame + lumaSize, pitch, dst, width * 3, width, height);
```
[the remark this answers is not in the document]
farbod-nv: added

examples/camera_streamer/operators/nv_stream_decoder/nv12_to_rgb.cu
```diff
@@ -0,0 +1,63 @@
...
+                                 int height,
+                                 cudaStream_t stream)
+{
+    dim3 block(16, 16);
+    dim3 grid((width + block.x - 1) / block.x,
+              (height + block.y - 1) / block.y);
+
+    nv12_to_rgb_fullrange_kernel<<<grid, block, 0, stream>>>(
+        y_plane, uv_plane, y_pitch, dst, dst_pitch, width, height);
+}
```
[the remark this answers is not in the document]
farbod-nv: added

examples/camera_streamer/teleop_camera_subgraph.py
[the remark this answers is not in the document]
farbod-nv: addressed above

examples/camera_streamer/operators/nv_stream_decoder/nv_stream_decoder_op.cpp
```diff
@@ -172,6 +178,20 @@ void NvStreamDecoderOp::compute(holoscan::InputContext& op_input,
         }
     }
 
+    // Detect full-range vs limited-range once after first successful decode.
```
life1ess, line 185: nit: Do we really need to check decoder format info when force_full_range_ is set to true?

farbod-nv: i think it's still useful to log it, in case we are forcing but the decoder metadata says we shouldn't

PR #337, opened 2026-03-31
Replace manual sccache installation with github action

aristarkhovNV:
## Summary by CodeRabbit

* **Chores**
  * Improved Windows build workflow by switching to a dedicated sccache action, removing manual installation, legacy cache steps and related environment variables, and simplifying verification and post-build reporting. The change keeps a lightweight pre-build verification step and reduces redundant post-build reporting for more reliable, maintainable builds.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: Verify sccache
       run: |
-        & "${env:SCCACHE_PATH}" --version
+        sccache --version
 
     # Put cl.exe and other build tools in PATH
     - name: Setup MSVC (Developer Command Prompt)


.github/workflows/build-windows.yml
```diff
@@ -39,8 +38,8 @@ jobs:
     - name: Install uv
       uses: ./.github/actions/setup-uv
 
-    - name: Install sccache
-      run: choco install -y sccache
+    - name: Setup sccache
+      uses: Mozilla-Actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
```
jiwenc-nv: Do we want to lock GitHub actions to version numbers or hashes?  seems we are a bit inconsistent here.

aristarkhovNV: Security best practice is to use hashes for more hermetic builds. Also for this particular action provider only the hash appears to be in the allow-list, the version tag was not allowed.

PR #342, opened 2026-04-01
fix[client]: Exit WebXR when failed to connect or streaming error

yanziz-nvidia:
## Summary by CodeRabbit

* **New Features**
  * Immersive XR exit events now trigger the same cleanup and UI teardown as the manual disconnect action, ensuring teleop timers and state reset.

* **Bug Fixes**
  * More robust error handling and resource cleanup for XR session failures (connection, stream, proxy, and reference-space issues), improving reliability of session termination and UI state.

### DIFF
--- a/deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
+++ b/deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
@@ -58,6 +58,12 @@ interface CloudXRComponentProps {
   /** Callback fired when an error occurs. Receives error message string. */
   onError?: (error: string) => void;
 
+  /**
+   * Called when CloudXR fails to connect or streaming stops with an error.
+   * Use this to end the immersive WebXR session (same as the user pressing Disconnect).
+   */
+  onExitImmersiveXR?: () => void;
+
   /** Callback fired when CloudXR session is created or destroyed. Receives session instance or null. */
   onSessionReady?: (session: CloudXR.Session | null) => void;
 
@@ -91,6 +97,7 @@ export default function CloudXRComponent({
   applicationName,
   onStatusChange,
   onError,
+  onExitImmersiveXR,
   onSessionReady,
   onServerAddress,
   onRenderPerformanceMetrics,
@@ -101,6 +108,8 @@ export default function CloudXRComponent({
   const { session } = useXR();
   // React reference to the CloudXR session that persists across re-renders.
   const cxrSessionRef = useRef<CloudXR.Session | null>(null);
+  const onExitImmersiveXRRef = useRef(onExitImmersiveXR);
+  onExitImmersiveXRRef.current = onExitImmersiveXR;
 
   // Metrics trackers for averaging performance metrics
   // Use prop values if provided, otherwise use defaults
@@ -192,6 +201,7 @@ export default function CloudXRComponent({
           } catch (error) {
             onStatusChange?.(false, 'Configuration Error');
             onError?.(`Proxy configuration failed: ${error}`);
+            onExitImmersiveXRRef.current?.();
             return;
           }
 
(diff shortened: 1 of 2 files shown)


deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
```diff
@@ -323,9 +335,19 @@ export default function CloudXRComponent({
             onStatusChange?.(false, 'Connection Failed');
             // Report error via callback
             onError?.('Failed to connect CloudXR session');
-            // Clean up the failed session
+            try {
```
nvddr, line 339: This seems spurious, why try to disconnect if connection failed?

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
```diff
@@ -323,9 +335,19 @@ export default function CloudXRComponent({
...
+              cxrSession.disconnect();
+            } catch (_) {
+              /* best-effort cleanup */
+            }
             cxrSessionRef.current = null;
+            onSessionReady?.(null);
```
nvddr, line 345: Why is this needed?

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
```diff
@@ -323,9 +335,19 @@ export default function CloudXRComponent({
...
+            onExitImmersiveXRRef.current?.();
           }
+        } else {
```
nvddr, line 348: What is this else?

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
yanziz-nvidia: disconnect triggers a set of cleanups

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
yanziz-nvidia: this gates the useEffect for MessageChannel in case the xr session is not null. We always clear this when stream properly stopped. Here it's triggered by error so should clear this here too

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
yanziz-nvidia: this is failing reference space request from webxr manager from three.js. We previously didn't catch this error. We should.

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
```diff
@@ -323,9 +335,19 @@ export default function CloudXRComponent({
             onStatusChange?.(false, 'Connection Failed');
             // Report error via callback
             onError?.('Failed to connect CloudXR session');
-            // Clean up the failed session
+            try {
+              cxrSession.disconnect();
+            } catch (_) {
+              /* best-effort cleanup */
```
blongs-nv: This comment could be confusing here since there's no additional cleanup in this catch block. If the best effort is to disconnect() and swallow errors, this comment should be above the try and here a comment would just say to ignore errors.

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
nvddr: The "if" body is very long. We should do early exit with reference space failure, or document what the else is. Prefer early exit.

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
nvddr: I dont understand. Can you add comments here? Can we have uniform cleanup instead of fragmented cleanup?

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
nvddr: That seems like code-smell.. but can likely defer for future.

deps/cloudxr/webxr_client/helpers/react/CloudXRComponent.tsx
yanziz-nvidia: makes sense, fixed

PR #352, opened 2026-04-03
Migrate XrTime to monotonic time in devicetracker API

nvddr:
## Summary by CodeRabbit

* **Documentation**
  * Added mandatory agent guidance files documenting contribution requirements and API contracts across the tracker subsystem and core modules.

* **Refactor**
  * Refactored time handling throughout tracker implementations to use monotonic nanosecond timestamps instead of OpenXR-specific time values.
  * Removed internal time converter state from session management, simplifying timing propagation.

### DIFF
--- a/src/core/deviceio_base/cpp/inc/deviceio_base/tracker.hpp
+++ b/src/core/deviceio_base/cpp/inc/deviceio_base/tracker.hpp
@@ -3,17 +3,13 @@
 
 #pragma once
 
-#include <openxr/openxr.h>
-
+#include <cstdint>
 #include <memory>
 #include <string>
 #include <string_view>
 
 namespace core
 {
-
-// Forward declarations
-struct OpenXRSessionHandles;
 
 // Base interface for tracker implementations.
 // The actual worker objects updated each frame by DeviceIOSession.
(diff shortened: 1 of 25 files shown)


src/core/deviceio_session/cpp/deviceio_session.cpp
```diff
@@ -103,11 +105,11 @@ std::unique_ptr<DeviceIOSession> DeviceIOSession::run(const std::vector<std::sha
 
 void DeviceIOSession::update()
```
aristarkhovNV, line 106: Consider passing the time externally:
- so users may coordinate a synchronized update across multiple sessions
- to allow more deterministic unit testing by passing known timestamps from the test body

src/core/deviceio_base/cpp/inc/deviceio_base/tracker.hpp
```diff
@@ -23,13 +19,16 @@ class ITrackerImpl
...
+     *
+     * @param monotonic_time_ns Current time from the system monotonic clock, in nanoseconds
+     *        (same domain as core::os_monotonic_now_ns()).
      *
      * @throws std::runtime_error On critical tracker/runtime failures.
      * @note A thrown exception indicates a fatal condition; the application is
      *       expected to terminate rather than continue running.
      */
-    virtual void update(XrTime time) = 0;
+    virtual void update(int64_t monotonic_time_ns) = 0;
```
aristarkhovNV, line 31: Consider using `std::chrono` for compile time type safety. Define a custom clock domain using something like `std::chrono::time_point<deviceio_monothonic_clock>` and a custom `deviceio_monothonic_clock` that would use platform specific query mechanisms.

src/core/deviceio_session/cpp/deviceio_session.cpp
nvddr: Yes I plan to do it in a follow-up. The current problem is that external users may pass in a completely arbitrary timestamp, and we don't really know what to do with that. So the API documentation needs some thinking, and what we do with the timestamp needs some thinking (e.g. if the timestamp is a future timestamp, is the future pose the one that goes into mcap or do we measure at current time as well and that goes into mcap? it's not clear).

src/core/deviceio_base/cpp/inc/deviceio_base/tracker.hpp
nvddr: We can defer this. Created an issue to track.

PR #353, opened 2026-04-03
Add CloudXRLauncher for programmatic runtime lifecycle management

rwiltz:
Extracts the CloudXR runtime + WSS proxy launch logic from __main__.py into a reusable CloudXRLauncher class that embedding applications (e.g. Isaac Lab Teleop) can drive without a separate terminal.
CloudXRLauncher provides start()/stop() and context-manager APIs to manage the full runtime lifecycle programmatically
Runtime is spawned via subprocess.Popen with start_new_session=True to isolate it in its own process group, avoiding CUDA context conflicts with the host application
WSS proxy runs in a background daemon thread with a dedicated event loop
Stale runtime cleanup detects and removes leftover IPC sockets before launching
Adds wait_for_runtime_ready_sync as a synchronous polling alternative to the existing async variant
Promotes runtime timeout constants to public module-level symbols
Refactors __main__.py to delegate to CloudXRLauncher, keeping the CLI as a thin entry point
Adds cloudxr_tests test package (27 tests) covering launcher lifecycle, sentinel polling, process termination escalation, and stale cleanup

## Summary by CodeRabbit

* **New Features**
  * Introduced CloudXRLauncher class for flexible, reusable runtime lifecycle management supporting both context-manager and explicit control patterns.

* **Improvements**
  * Streamlined application startup and shutdown with improved signal handling and runtime readiness coordination.

* **Tests**
  * Added comprehensive test suite validating launcher initialization, runtime startup/stop, and lifecycle edge cases.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -235,6 +235,7 @@ jobs:
       env:
         ACCEPT_CLOUDXR_EULA: Y
         CXR_BUILD_CONTEXT: ${{ github.workspace }}
+        CXR_RUNTIME_NETWORK_MODE: bridge
         PYTHON_VERSION: ${{ matrix.python_version }}
         ISAACTELEOP_PIP_DEBUG: "0"
         ISAACTELEOP_PIP_FIND_LINKS: /workspace/install/wheels


src/core/cloudxr/python/__main__.py
```diff
@@ -48,60 +37,54 @@ def _parse_args() -> argparse.Namespace:
...
     args = _parse_args()
-    env_cfg = EnvConfig.from_args(args.cloudxr_install_dir, args.cloudxr_env_config)
-    check_eula(accept_eula=args.accept_eula or None)
-    logs_dir_path = env_cfg.ensure_logs_dir()
-    wss_ts = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H%M%SZ")
-    wss_log_path = logs_dir_path / f"wss.{wss_ts}.log"
 
-    runtime_proc = multiprocessing.Process(target=runtime_run)
-    runtime_proc.start()
+    launcher = CloudXRLauncher(
```
aristarkhovNV: consider raii style initialization:
```python
with CloudXRLauncher(...) as launcher:
   # body - luncher enters in started state or raises an exception

# launcher automatically stops upon exiting the with: block
print("stopped")

File: src/core/cloudxr/python/runtime.py [hunk 2]
```diff
@@ -67,13 +76,48 @@ def _get_sdk_path() -> str | None:
     return native_dir
 
 
+def wait_for_runtime_ready_sync(
```
aristarkhovNV: this could be a thin wrapper over existing async function so the core waiting logic stays in one place:
```python
asyncio.run(wait_for_runtime_ready())
```
or
```python
loop.run_until_complete.run(wait_for_runtime_ready())
```

src/core/cloudxr/python/launcher.py
```diff
@@ -0,0 +1,309 @@
...
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        self.stop()
+
+    # ------------------------------------------------------------------
+    # Public API
+    # ------------------------------------------------------------------
+
+    def start(self) -> None:
```
aristarkhovNV: If we do not necessary require multiple re-starts of the same launcher instance then we could reduce the API surface and test burden by starting immediately upon creation and running until the end of launcher scope.

src/core/cloudxr/python/launcher.py
```diff
@@ -0,0 +1,309 @@
...
+        if self._runtime_proc is not None:
+            try:
+                self._terminate_runtime()
+            except RuntimeError:
+                logger.warning("Failed to cleanly terminate CloudXR runtime process")
+            self._runtime_proc = None
+            logger.info("CloudXR runtime process stopped")
+
+    @property
+    def is_running(self) -> bool:
```
aristarkhovNV: If the runtime or wss proxy spontaniously stop that could be treated as an exceptional situation that normally is not supposed to happen. So we could convert this into a `health_check()` function that would return immediately if all is good or raise an exception with details on what happened. This will provide a controllable teardown sequence and wipe unhealthy launcher from existence asap. This pattern is used in `isaacteleop` plugin management.

src/core/cloudxr_tests/python/test_runtime.py
```diff
@@ -0,0 +1,215 @@
...
+    """Guard against accidental changes to the public timeout constants."""
+
+    def test_startup_timeout(self):
+        assert RUNTIME_STARTUP_TIMEOUT_SEC == 10
+
+    def test_terminate_timeout(self):
+        assert RUNTIME_TERMINATE_TIMEOUT_SEC == 10
+
+    def test_poll_interval(self):
+        assert RUNTIME_POLL_INTERVAL_SEC == 0.5
```
aristarkhovNV: testing constants in general is not necessary, they will show up themselves in a PR diff if attempted to change

rwiltz: Yeah this is weird, I had cursor generate the tests. I'' fix

src/core/cloudxr/python/__main__.py
rwiltz: This was fixed in latest commit (caught by review bot)

src/core/cloudxr/python/__main__.py
```diff
@@ -48,60 +39,53 @@ def _parse_args() -> argparse.Namespace:
...
+            stop = True
+
+        signal.signal(signal.SIGINT, on_signal)
+        signal.signal(signal.SIGTERM, on_signal)
+
+        while not stop:
+            try:
+                launcher.health_check()
+            except RuntimeError:
+                break
```
aristarkhovNV: No need to catch here, let the exception propagate and crash the program. The default top level handler will print the call stack and exception info. The enclosing `with:` block shall guarantee a graceful teardown of the launcher upon leaving the scope regardless of the cause.

deps/cloudxr/docker-compose.runtime.yaml
```diff
@@ -19,10 +19,10 @@ services:
     network_mode: "${CXR_RUNTIME_NETWORK_MODE:-host}"
     healthcheck:
       test: ["CMD", "test", "-f", "/openxr/.cloudxr/run/runtime_started"]
-      interval: 1s
+      interval: 2s
```
yanziz-nvidia: why reducing ?

rwiltz: Trying to track down a CI failure, will revert once root caused

.github/workflows/build-ubuntu.yml
```diff
@@ -235,6 +235,7 @@ jobs:
       env:
         ACCEPT_CLOUDXR_EULA: Y
         CXR_BUILD_CONTEXT: ${{ github.workspace }}
+        CXR_RUNTIME_NETWORK_MODE: bridge
```
rwiltz, line 238: @aristarkhovNV Can you review this?

yanziz-nvidia: what does the test do btw? doing bridge mode means we go with testing within container

rwiltz: I'm not sure but cursor says:

The test validates that teleop_ros2_publisher.py can initialize a TeleopSession in each of its four modes (controller_teleop, hand_teleop, controller_raw, full_body) against a live CloudXR runtime, using mock operators (no real hardware). It starts the runtime, then for each mode runs the publisher for 5 seconds and asserts "TeleopSession started successfully" appears in the log.

Switching to bridge mode only affects the CloudXR runtime's streaming port (49100), which is used for WebRTC connections from remote VR headsets. The test never connects a headset -- the ROS2 publisher communicates with the runtime through OpenXR IPC via a Unix domain socket on the shared volume (CXR_HOST_VOLUME_PATH), which works regardless of network mode. The test-cloudxr job already uses bridge mode for exactly the same reason and works fine.

The only thing bridge isolates is port 49100 inside the container's network namespace, preventing conflicts when multiple matrix jobs land on the same self-hosted runner.

jiwenc-nv: The tests are all running inside the container. I already change the local run of the same test suite to use bridge mode, so during it for CI should be the same move.

https://github.com/NVIDIA/IsaacTeleop/pull/302

PR #380, opened 2026-04-09
Replace ROS test timeout with a readiness poll

sgrizan-nv:
## Summary by CodeRabbit

* **Chores**
  * Improved reliability and error reporting in the automated testing infrastructure to provide clearer diagnostics during the build and test process.


aristarkhovNV (approved):

.github/workflows/build-ubuntu.yml
```diff
@@ -302,33 +305,67 @@ jobs:
...
+
+          docker logs "$RUN_NAME" > "$LOG_FILE" 2>&1 || true
+
+          if [ "$started" = true ]; then
+            docker stop --time 5 "$RUN_NAME" >/dev/null || true
+            docker wait "$RUN_NAME" >/dev/null || true
           else
-            echo "No test log captured at $LOG_FILE"
+            running=$(docker inspect -f '{{.State.Running}}' "$RUN_NAME" 2>/dev/null || echo false)
+            if [ "$running" = "true" ]; then
```
jiwenc-nv, line 351: just being paranoid, any chance that two `build-ubuntu.yml` workflow running together and one will kill the other? can that happen?  can we check how long a container is running and only kill if it's long lasting?

sgrizan-nv: There is a chance, if the workflows are executed on the same runner. To handle that, we'll need something like https://github.com/NVIDIA/IsaacTeleop/pull/382

PR #389, opened 2026-04-13
[Replay] Add latest channel support in schema_tracker.

shaosu-nvidia:
- Enable pedals to write the latest channel

## Summary by CodeRabbit

* **New Features**
  * Enhanced MCAP recording to support optional latest-value tracking for pedal data, enabling efficient access to the most recent pedal data during playback.

### DIFF
--- a/src/core/live_trackers/cpp/inc/live_trackers/schema_tracker.hpp
+++ b/src/core/live_trackers/cpp/inc/live_trackers/schema_tracker.hpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
@@ -9,6 +9,7 @@
 #include <mcap/tracker_channels.hpp>
 
 #include <memory>
+#include <optional>
 
 namespace core
 {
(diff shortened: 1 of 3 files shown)


src/core/live_trackers/cpp/inc/live_trackers/schema_tracker.hpp
nvddr, line 70: Could you make "update" return "TrackedT" and then write the "TrackedT" in the mcap? I think that's more future proof.. It would move the "TrackedT" construction from pedal tracker into schema tracker..

nvddr: The rationale is that the mcap replayer can be less smart, it just records the "Tracked" type and the replay tracker can jsut return directly from mcap. Because the "Tracked" type also encapsulates whether there was any data or no and so on..

src/core/mcap/cpp/inc/mcap/recording_traits.hpp
```diff
@@ -44,7 +44,7 @@ struct FullBodyPicoRecordingTraits
 struct PedalRecordingTraits
 {
     static constexpr std::string_view schema_name = "core.Generic3AxisPedalOutputRecord";
-    static constexpr std::array channels = { "pedals" };
+    static constexpr std::array channels = { "pedals", "pedals_latest" };
```
nvddr: Instead of "_latest" suffix, I think we can go with "_tracked" suffix. It seems cleaner to me.

src/core/live_trackers/cpp/inc/live_trackers/schema_tracker.hpp
shaosu-nvidia: Talked offline and we stick with only writing RecordT into MCAP for now.

src/core/live_trackers/cpp/live_generic_3axis_pedal_tracker_impl.cpp
```diff
@@ -40,7 +40,7 @@ LiveGeneric3AxisPedalTrackerImpl::LiveGeneric3AxisPedalTrackerImpl(const OpenXRS
                                                                    const Generic3AxisPedalTracker* tracker,
                                                                    std::unique_ptr<PedalMcapChannels> mcap_channels)
     : mcap_channels_(std::move(mcap_channels)),
-      m_schema_reader(handles, make_pedal_tensor_config(tracker), mcap_channels_.get(), 0)
+      m_schema_reader(handles, make_pedal_tensor_config(tracker), mcap_channels_.get(), 0, 1)
```
nvddr: Is there a way to avoid hard-coding these integers? Or add the arg name, e.g. /*argname = */0, /*argname2 = */1 .. Because when I see this I don't know what 0 and 1 are.

nvddr (approved):

src/core/live_trackers/cpp/live_generic_3axis_pedal_tracker_impl.cpp
shaosu-nvidia: Updated. Thanks.

PR #391, opened 2026-04-14
Wire up ExecutionEvents.reset in stateful retargeters

rwiltz:
Handle context.execution_events.reset in all stateful retargeters (GripperRetargeter, LocomotionRootCmdRetargeter, Se3AbsRetargeter, Se3RelRetargeter) so cross-step state is reinitialized when the pipeline signals a reset.
Add bilateral tests (reset=True / reset=False) for each retargeter to verify sta

## Summary by CodeRabbit

* **Tests**
  * Added unit and end-to-end tests validating reset behavior across locomotion, absolute/relative pose, and gripper flows to ensure outputs reflect reset vs. non-reset scenarios.

* **Bug Fixes**
  * Improved reset handling so locomotion, pose (abs/rel), and gripper components reliably reinitialize state on system reset, yielding stable and predictable outputs.

### DIFF
--- a/src/core/teleop_session_manager_tests/python/test_session_reset.py
+++ b/src/core/teleop_session_manager_tests/python/test_session_reset.py
@@ -1,0 +1,222 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+End-to-end tests for reset propagation through TeleopSession.
+
+Verifies that ``TeleopSession.step(execution_events=ExecutionEvents(reset=True))``
+correctly propagates the reset signal to stateful retargeters in a real pipeline
+(not mocked). All OpenXR/DeviceIO dependencies are patched out.
+"""
+
+import numpy as np
+import pytest
+from contextlib import contextmanager
+from unittest.mock import MagicMock, patch
+
+from isaacteleop.retargeting_engine.interface import (
+    ExecutionEvents,
+    ExecutionState,
+    OptionalTensorGroup,
+)
+
+from isaacteleop.retargeters import (
+    GripperRetargeter,
+    GripperRetargeterConfig,
+    LocomotionRootCmdRetargeter,
+    LocomotionRootCmdRetargeterConfig,
+)
+
+from isaacteleop.retargeting_engine.tensor_types import ControllerInput, HandInput
+from isaacteleop.teleop_session_manager import TeleopSession, TeleopSessionConfig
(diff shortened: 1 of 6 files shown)


src/retargeters/se3_retargeter.py
```diff
@@ -333,6 +338,14 @@ def output_spec(self) -> RetargeterIOType:
         }
 
     def _compute_fn(self, inputs: RetargeterIO, outputs: RetargeterIO, context) -> None:
+        if context.execution_events.reset:
```
nvddr, line 343: When reset is observed, do you want to do an early return or something? Do these retargeters need warmup?

rwiltz: No early return needed — none of these retargeters require warmup after reset:

Se3AbsRetargeter: Resets _last_pose to identity. If valid input arrives on the same frame, it computes a valid pose immediately. If no input, identity is returned (same as a fresh start). No warmup needed.
Se3RelRetargeter: Resets _first_frame = True. The existing first-frame path (lines 387-393) already provides natural one-frame warmup — it captures the new baseline and outputs zero deltas. So reset effectively re-arms the warmup that's already built in.
LocomotionRootCmdRetargeter: Restores _hip_height to initial. Computes immediately from controller inputs. No warmup.
GripperRetargeter: Restores _previous_gripper_command to open. Computes immediately. No warmup.
An early return after the reset block would waste valid tracking data arriving on the same frame. I've added inline comments clarifying this intent on both Se3 retargeters.

src/core/teleop_session_manager_tests/python/test_session_reset.py
```diff
@@ -0,0 +1,222 @@
...
+                # Reset via session
+                result = session.step(
+                    external_inputs=ext,
+                    execution_events=_running_events(reset=True),
+                )
+
+                cmd = result["gripper_command"][0]
+                assert cmd == pytest.approx(1.0), (
+                    "gripper should be open after session-level reset"
+                )
```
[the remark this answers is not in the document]
rwiltz: Investigated — no change needed here. GripperRetargeter's output spec uses FloatType("command") (a scalar type), not NDArrayType. The FloatType.validate_value enforces that the stored value is a Python float, so result["gripper_command"][0] returns a plain float via Tensor.value — not a tensor buffer or ndarray. The pytest.approx(1.0) comparison works correctly against a native float.

This is distinct from NDArrayType outputs like root_command (shape (4,)), which do require np.from_dlpack() to extract. The existing unit test in test_retargeter_reset.py uses the same pattern and passes.

nvddr (approved):

PR #396, opened 2026-04-14
[Replay] Add McapTrackerViewers to read mcap file.

shaosu-nvidia:
Following the same pattern of McapTrackerChannels
- it takes a McapReader and a list of channels we want to replay
- internally it keeps the MessageView and iterator to read message one by one.

## Summary by CodeRabbit

* **New Features**
  * Added a templated reader component for extracting tracked data from MCAP files with multi-channel filtering support
  * Introduced a utility function for constructing MCAP topic identifiers

* **Tests**
  * Added test coverage for the new reader functionality, including channel filtering and null payload handling

* **Chores**
  * Updated copyright year range

### DIFF
--- a/src/core/mcap/cpp/inc/mcap/tracker_channels.hpp
+++ b/src/core/mcap/cpp/inc/mcap/tracker_channels.hpp
@@ -111,12 +111,14 @@ private:
 };
 
 /**
- * @brief Type-safe MCAP channel reader for FlatBuffer record types.
+ * @brief Type-safe MCAP channel reader returning deserialized Tracked wrappers.
  *
- * @tparam RecordT   The FlatBuffer record wrapper type (e.g. HeadPoseRecord).
- *                   Must expose a data() accessor returning the FlatBuffer data table.
- * @tparam DataTableT The FlatBuffer data table type (e.g. HeadPose).
- *                   Must expose UnPackTo() and NativeTableType.
+ * @tparam RecordT       The FlatBuffer record wrapper stored in MCAP (e.g. HeadPoseRecord).
+ *                       Must expose a data() accessor returning the inner data table.
+ * @tparam DataTableT    The FlatBuffer data table type (e.g. HeadPose).
+ *                       Must expose UnPackTo() and NativeTableType.
+ * @tparam TrackedTableT The FlatBuffer tracked wrapper returned to callers
+ *                       (e.g. HeadPoseTracked). Must share the same data table as RecordT.
  *
  * Read-side counterpart to McapTrackerChannels. Takes an externally-owned
  * McapReader& and a list of sub-channel names, creating one LinearMessageView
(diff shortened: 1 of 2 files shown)


src/core/mcap/cpp/inc/mcap/tracker_channels.hpp
```diff
@@ -103,4 +105,91 @@ class McapTrackerChannels
...
+template <typename RecordT, typename DataTableT>
+class McapTrackerViewers
+{
+public:
+    using NativeDataT = typename DataTableT::NativeTableType;
+
+    McapTrackerViewers(mcap::McapReader& reader, std::string_view base_name, const std::vector<std::string>& sub_channels)
+        : reader_(&reader)
+    {
+        auto on_problem = [](const mcap::Status& s) { std::cerr << "McapTrackerViewers: " << s.message << std::endl; };
```
nvddr, line 137: Why are these named topic and problem? Are these mcap-specific? May want to add some comments, the variable names are cryptic

shaosu-nvidia: Yeah they are all mcap definition. "topic" also applies to writer when we create a channel. I created an inline function for both writer and reader case. 
on_problem is a callback function that mcap is using as well.

qingsi-at-nv (approved):

src/core/mcap/cpp/inc/mcap/tracker_channels.hpp
```diff
@@ -103,4 +110,91 @@ class McapTrackerChannels
...
+
+        channels_.reserve(sub_channels.size());
```
nvddr: This needs some comments as it is important. The ChannelView stores iterators to itself, and the reserve prevents vector reallocation. 

If the reserve was removed, the and the vector got re-allocated midway, it would invalidate the iterator potentially because the LinearMessageView may be reallocated. 

Alternatively, make the vector hold uniqueptrs - then no issues with reallocation.

src/core/mcap/cpp/inc/mcap/tracker_channels.hpp
```diff
@@ -103,4 +110,91 @@ class McapTrackerChannels
...
+        }
+    }
+
+    /**
+     * @brief Read and deserialize the next data payload from the given channel.
+     * @param channel_index Index into the sub_channels list passed at construction.
+     * @return shared_ptr to the deserialized data (null if the record had no data),
+     *         or std::nullopt when no more messages remain.
+     */
+    std::optional<std::shared_ptr<NativeDataT>> read(size_t channel_index)
```
nvddr: From previous MR, we should be reading "TrackedT" from the mcap directly for this case.. "Record" types are for incoming data from runtime, "Tracked" types are what trackers return.

src/core/mcap/cpp/inc/mcap/tracker_channels.hpp
shaosu-nvidia: Good catch! Use the unique_ptr instead.

src/core/mcap/cpp/inc/mcap/tracker_channels.hpp
shaosu-nvidia: Discussed offline, let's just write RecordT into mcap but for read, read TrackedT directly

nvddr (approved):

PR #418, opened 2026-04-25
test(webclient): Adds e2e web client quick start test

yanziz-nvidia:
Example run: https://github.com/NVIDIA/IsaacTeleop/actions/runs/24949804846/job/73058257235

Note we don't have an IL or other app connected. So it shows empty frames. Also it's odd react control panel isn't shown. But at least we can get a connected session, a video recording (from artifacts), and some logs for e2e.

Download artifact https://github.com/NVIDIA/IsaacTeleop/actions/runs/24949804846/artifacts/6645351244 and see.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end testing for the web client, validating SDK initialization, UI state transitions, runtime connectivity, and frame delivery.

* **Chores**
  * Updated CI/CD pipeline to build web client artifacts and execute E2E tests as a release gate.
  * Updated project ignore patterns for test-generated artifacts.

* **Documentation**
  * Added E2E test suite documentation with local execution guidelines and environment variable configuration.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -818,7 +818,7 @@ jobs:
     # client) and the gripper retargeting example concurrently against the same
     # live runtime, exercising the full quick-start pipeline end-to-end.
     # Skipped on PRs that don't touch the web stack — see detect-changes.
-    runs-on: [self-hosted, linux, gpu]
+    runs-on: [self-hosted, linux, gpu, "x64"]
     needs:
       - detect-changes
       - build-ubuntu
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -818,12 +818,17 @@ jobs:
     # client) and the gripper retargeting example concurrently against the same
     # live runtime, exercising the full quick-start pipeline end-to-end.
     # Skipped on PRs that don't touch the web stack — see detect-changes.
-    runs-on: [self-hosted, linux, gpu, "x64"]
+    runs-on: [self-hosted, linux, gpu, "${{ matrix.arch }}"]
     needs:
       - detect-changes
       - build-ubuntu
       - build-webapp
     if: needs.detect-changes.outputs.web == 'true'
+
+    strategy:
+      fail-fast: ${{ github.event_name == 'pull_request' }}
+      matrix:
+        arch: ['x64', 'arm64']
 
     steps:
     - name: Checkout code
@@ -834,7 +839,7 @@ jobs:
     - name: Download install artifacts
       uses: actions/download-artifact@v7
       with:
-        name: isaacteleop-install-release-x64-py3.11
+        name: isaacteleop-install-release-${{ matrix.arch }}-py3.11
 
     - name: Extract tarball to preserve permissions
       run: |
@@ -869,7 +874,7 @@ jobs:
       if: always()
       uses: actions/upload-artifact@v6
       with:
-        name: web-e2e-results
+        name: web-e2e-results-${{ matrix.arch }}
         # Compose/runtime logs + Playwright JSON summary + screenshots +
         # video recording (every run) + trace zip (failure runs only,
         # per playwright.config.ts).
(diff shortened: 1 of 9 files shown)


scripts/run_test_e2e_with_web.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
```
gareth-morgan-nv, line 1: Do we need the script?  Could the logic be contained in Playwright?

gareth-morgan-nv (commented):
Wouldn't it be better to start with a simple playwright test that assumes no server?

scripts/run_test_e2e_with_web.sh
yanziz-nvidia: It's useful to run it locally imo.

yanziz-nvidia:
> Wouldn't it be better to start with a simple playwright test that assumes no server?

then that wouldn't be called e2e right? I would want to make sure it actually connects. 

~This PR also lacks an actual server running. Open for ideas what to run.~

I have it hooked up to the example

deps/cloudxr/docker-compose.test-e2e.yaml
```diff
@@ -0,0 +1,124 @@
...
+    volumes:
+      # /openxr/ stays in the container's own writable layer — Playwright
+      # never reads it, so there's no need for a shared named volume here
+      # (and a Docker-managed named volume seeded from the image's
+      # root-owned /openxr/ would break a non-root container user).
+      #
+      # Bind-mount the prebuilt web client so python -m http.server can serve
+      # it. CI sets CXR_WEBAPP_DIR to the downloaded webxr-client-build artifact
+      # directory; the default works for local runs after `npm run build`.
+      - ${CXR_WEBAPP_DIR:-./webxr_client/build}:/app/webapp:ro
```
yanziz-nvidia, line 66: I guess this is the culprit that caused root permission.

yanziz-nvidia: nvm not here

.github/workflows/build-ubuntu.yml
```diff
@@ -460,13 +463,120 @@ jobs:
...
+        name: webxr-client-build
+        path: deps/cloudxr/webxr_client/build
+        if-no-files-found: error
+        retention-days: 1
+
+  test-cloudxr-web-e2e:
+    # GPU runner required: the CloudXR runtime needs an NVIDIA GPU even when
+    # only one E2E spec is running. Single matrix entry — the web client
+    # behavior is independent of the matrix dims used by other test jobs.
+    runs-on: [self-hosted, linux, gpu, x64]
```
aristarkhovNV: may skip the arch tag so it'd use any available runner

aristarkhovNV (approved):

jiwenc-nv (approved):

.github/workflows/build-ubuntu.yml
```diff
@@ -460,13 +463,120 @@ jobs:
...
+
+    - name: Upload web client artifact
+      uses: actions/upload-artifact@v6
+      with:
+        name: webxr-client-build
+        path: deps/cloudxr/webxr_client/build
+        if-no-files-found: error
+        retention-days: 1
+
+  test-cloudxr-web-e2e:
```
jiwenc-nv: We should gate this test, so that it only runs when there are actual changes to the web app.  This way we can get CI pass faster.

jiwenc-nv (changes requested):

.github/workflows/build-ubuntu.yml
yanziz-nvidia: sounds good

yanziz-nvidia: done

.github/workflows/build-ubuntu.yml
yanziz-nvidia: done

aristarkhovNV (commented):
>On aarch64 runners the x64 wheels can't install

why don't we get an arm wheel as well?

yanziz-nvidia:
> > On aarch64 runners the x64 wheels can't install
> 
> why don't we get an arm wheel as well?

yes can enable let me add

PR #436, opened 2026-04-29
Use NV GHA runners for test-cloudxr job

aristarkhovNV:


fvaloppi-nv:
@jiwenc-nv to check if we still want to do this.

jiwenc-nv:
we probably still want this.  need more discussions.  keep it in draft for now.

jiwenc-nv:
Had a discussion with @aristarkhovNV and the learning was that GHA merging process is not ideal and we don't want to use this path. Closing.

PR #466, opened 2026-05-04
[TeleViz] Add QuadLayer + textured-quad pipeline + VizCudaArray

farbod-nv:
QuadLayer is the first concrete layer type: a fullscreen-blit textured quad sampling a CUDA-fed DeviceImage. Two submission paths:

- submit(VizBuffer)        Mode A: caller's CUDA buffer is copied
                           into the layer's DeviceImage.
- acquire() / release()    Mode B: caller writes into the layer's
                           tiled CUDA-Vulkan image directly. Zero
                           copy. acquire() returns VizCudaArray.

Two image-shape view types now sit side-by-side in viz/core:

- VizBuffer    — linear pointer-backed memory (CPU bytes or CUDA
                 device pointer). Exposes __cuda_array_interface__
                 in Python for kDevice.
- VizCudaArray — opaque CUDA cudaArray_t (tiled GPU memory). Does
                 NOT expose __cuda_array_interface__ — that protocol
                 means pointer-backed memory and a cudaArray_t isn't
                 one. Used as Mode B return.

Pipeline machinery (sampler, descriptor set layout, pipeline layout, VkPipeline, descriptor pool/set) lives inside QuadLayer. Pipelines build using a process-wide VkPipelineCache added to VkContext for driver-side compilation reuse.

## Summary by CodeRabbit

* **New Features**
  * QuadLayer: render CUDA-fed textures as fullscreen quads with two submission modes (copy and zero‑copy).
  * Vulkan pipeline cache exposed for faster pipeline reuse.
  * VizSession exposes a Vulkan context accessor.
  * Per-layer timeline-semaphore hooks for explicit GPU synchronization.

* **Tests**
  * Added unit and GPU tests for QuadLayer and end-to-end CUDA↔Vulkan interop (including multi-frame).

* **Documentation**
  * Clarified module docs on CUDA tiled arrays and static layer layout.

### DIFF
--- a/src/viz/AGENTS.md
+++ b/src/viz/AGENTS.md
@@ -16,12 +16,18 @@ single sub-module. Each sub-module is its own static library with its own
 sibling `<sub-module>_tests/` directory:
 
 - **`viz/core/`** — foundational types + Vulkan/CUDA infrastructure.
-  Library: `viz_core`. Today: `VkContext`, `VizBuffer`, `Pose3D`, `Fov`,
-  `Resolution`, `ViewInfo`, `PixelFormat`, `RenderTarget`, `FrameSync`,
-  `HostImage`, `DeviceImage`. `HostImage` / `DeviceImage` are the
-  symmetric pair of owning 2D pixel buffers (CPU bytes vs CUDA-Vulkan
-  interop) — both expose `VizBuffer view()` so generic helpers branch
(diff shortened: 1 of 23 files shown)


farbod-nv:
@coderabbitai review

junjeechao:
Do we need to change the MR summary? I see acquire() and release() was removed from QuadLayer

junjeechao (approved):

PR #478, opened 2026-05-07
perf(dex-hand): run retargeting optimizer under torch.no_grad

nvddr:
## Summary
- `DexHandRetargeter._compute_hand` was wrapping every call to `self._dex_hand.retarget(...)` in `torch.enable_grad()` + `torch.inference_mode(False)`. The dex_retargeting QP solver does not consume those gradients, so each frame paid for autograd bookkeeping that was immediately discarded.
- Switch to `torch.no_grad()` so forward tensor ops inside the optimizer skip grad tracking entirely. The optimizer is unchanged; only the surrounding context flips.

## Why this matters
At 90 Hz with two `DexHandRetargeter` instances (bimanual), the per-frame cost adds up. Disabling grad tracking on the hot path is a free win unless dex_retargeting internally requires autograd (which it does not — it's a NLOpt-style optimization).

## Risk
The previous explicit `enable_grad()` + `inference_mode(False)` reads as defensive escape from an outer no-grad context, but dex_retargeting's `retarget()` does not differentiate through any tensor operation, so running under `no_grad` is safe. If any downstream code actually relies on graph capture from `retarget(...)` (none in this repo), CI will flag it.

## Test plan
- [ ] `ctest` retargeting suite green
- [ ] Optional: run an existing dex bimanual example; confirm output joints match the previous run on the same recorded input.

## Summary by CodeRabbit

* **Bug Fixes**
  * Optimized hand retargeting execution context for improved performance during retargeting operations.

### DIFF
--- a/src/retargeters/dex_hand_retargeter.py
+++ b/src/retargeters/dex_hand_retargeter.py
@@ -412,10 +412,14 @@ class DexHandRetargeter(BaseRetargeter):
             ref_value = target_pos[task_indices, :] - target_pos[origin_indices, :]
 
         # 4. Run optimizer
+        # ``dex_retargeting`` solves a QP-style optimization that does not
+        # require autograd; running under ``torch.no_grad()`` avoids the
+        # per-step grad-tracking overhead the previous ``enable_grad`` /
+        # ``inference_mode(False)`` context was incurring on every frame.
         try:
             import torch  # type: ignore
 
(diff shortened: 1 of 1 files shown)


jiwenc-nv:
do you still want to follow up and land them?  Or close?  I'm moving them into draft and will close in a week if no actions by then:

PR #479, opened 2026-05-07
fix(dex-hand): release temp YAML config after optimizer build

nvddr:
## Summary
- `DexHandRetargeter._update_yaml` writes a temp file (`tempfile.mkstemp(suffix=".yml", prefix="dex_retarget_")`) per construction so the URDF path can be patched without mutating the original config. The fd was closed but the file was never unlinked.
- Track the temp path on the instance and unlink it right after `RetargetingConfig.load_from_file(...).build()` returns — the optimizer has parsed the YAML by then and no longer needs the file.
- Add `__del__` as a safety net for the construction path where the optimizer build raises after the temp file was written.
- `_cleanup_temp_config` is idempotent (safe to call from both the explicit cleanup and `__del__`).

## Why this matters
Long-running sessions, test suites, and pytest runs that instantiate many `DexHandRetargeter`s leak a `/tmp/dex_retarget_*.yml` file each, slowly filling `/tmp`. No functional impact, but it's the kind of leak that surfaces as an outage on a long-running production deployment.

## Test plan
- [ ] Run any existing dex hand example/test; confirm `/tmp/dex_retarget_*` count is unchanged after the run vs. before.
- [ ] CI retargeting suite green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved temporary file cleanup to ensure all temporary configuration files are properly removed after optimizer initialization.
  * Added additional safety measures to guarantee cleanup completion even in edge cases.

### DIFF
--- a/src/retargeters/dex_hand_retargeter.py
+++ b/src/retargeters/dex_hand_retargeter.py
@@ -135,6 +135,9 @@ class DexHandRetargeter(BaseRetargeter):
             config.hand_retargeting_config
         ).build()
 
+        # The optimizer has parsed the YAML; the temp file is no longer needed.
+        self._cleanup_temp_config()
+
         # Cache joint names from optimizer
         self._dof_names = self._dex_hand.optimizer.robot.dof_joint_names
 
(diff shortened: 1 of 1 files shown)


jiwenc-nv:
do you still want to follow up and land them?  Or close?  I'm moving them into draft and will close in a week if no actions by then:

jiwenc-nv:
do you still want to follow up and land them?  Or close?  I'm moving them into draft and will close in a week if no actions by then:

PR #489, opened 2026-05-09
fix(cmake): allow build wihtout git

jiwenc-nv:
When using locally, the version number is always X.Y-local, so we technically don't need to use git to calculate all the patch numbers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version detection and handling across different environments.
  * Enhanced CI environment version detection with better fallback behavior for local development builds.

* **Chores**
  * Refactored version computation logic for better maintainability.


cmake/IsaacTeleopVersion.cmake
aristarkhovNV: consider also removing git package dependency when not in CI.
BTW i made similar change, never landed though 358a422df2516d2377e89fc0cce41b39f0cb7266

aristarkhovNV (approved):

cmake/IsaacTeleopVersion.cmake
jiwenc-nv: done.

PR #516, opened 2026-05-12
add required-environments platform constraints

aristarkhovNV:
Enforces that a valid resolution must exist for every environment listed

## Summary by CodeRabbit

* **Chores**
  * Updated Python example installation configuration to explicitly support Linux aarch64 and x86_64 architectures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/IsaacTeleop/pull/516)

### DIFF
--- a/examples/teleop_ros2/Dockerfile
+++ b/examples/teleop_ros2/Dockerfile
@@ -109,8 +109,17 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /opt/isaacteleop
 COPY . /opt/isaacteleop
 
+# Wipe any stale wheels before building:
+#   - `install/` may arrive via `COPY . /opt/isaacteleop` when the host context
+#     contains an extracted build-ubuntu artifact tarball.
+#   - `build/wheels/` lives inside the persistent cmake build cache mount and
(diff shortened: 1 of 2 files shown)


nvddr (approved):

examples/teleop_ros2/Dockerfile
```diff
@@ -109,8 +109,17 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /opt/isaacteleop
 COPY . /opt/isaacteleop
 
+# Wipe any stale wheels before building:
+#   - `install/` may arrive via `COPY . /opt/isaacteleop` when the host context
+#     contains an extracted build-ubuntu artifact tarball.
```
nv-jakob, line 114: Feels like this shouldn't be in the artifact tarball in the first place? We can fix this in a follow tho, let's get this in.

nv-jakob (approved):

examples/teleop_ros2/Dockerfile
aristarkhovNV: yea test-teleop-ros2 pulls artifacts for some reason, then builds its own wheel

PR #527, opened 2026-05-13
Add frame-aligned MessageChannelTracker replay

rwiltz:
## Summary

Adds `ReplayMessageChannelTrackerImpl` so `ReplaySession` can re-emit the `_teleop_control` records captured by the live `LiveMessageChannelTrackerImpl`. Before this PR, `ReplayDeviceIOFactory::create_tracker_impl` rejected `MessageChannelTracker` outright, stranding the recorded START / STOP / RESET events in the MCAP.

## Design

Events are emitted **frame-aligned** with the rest of the recorded data: each `session.update()` advances all trackers by exactly one recorded frame regardless of replay-loop dt, so the relative position of control events vs the per-frame tracker stream is invariant under any playback rate. On construction the impl reads the MCAP summary, anchors `recording_start_ns_` to `messageStartTime`, and derives an inter-frame interval from the densest non-self channel. Each `update()` drains every pending record whose `logTime - messageStartTime <= frame_counter * recording_dt_ns` into the tracked batch. Wall-clock time is intentionally ignored.

If the summary is missing or no reference channel exists, the impl falls back to one record per `update()` so events still surface in recorded order.

## Changes

- `src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.{hpp,cpp}` -- new impl.
- `replay_deviceio_factory.{hpp,cpp}` -- new dispatch entry + public `create_message_channel_tracker_impl`.
- `CMakeLists.txt` -- new source listed.
- Three new `[message_channel]` Catch2 cases in `test_replay_session.cpp` covering same-frame drain, multi-tick fan-out, and frame-alignment under arbitrary replay-loop speed (no sleeps).

## Tests

`ctest -R replay_deviceio_session_tests` -- 93 assertions / 7 cases passing.
`SKIP=check-copyright-year pre-commit run --all-files` -- clean.

## Downstream

Consumed by `IsaacLab#<PR>` (`rwiltz/isaacteleop-record-replay`); requires republishing the `isacteleop` wheel before IsaacLab can land.

## Summary by CodeRabbit

* **New Features**
  * Added message-channel replay functionality for MCAP recordings, with automatic frame-based timing alignment derived from recording metadata.

* **Tests**
  * Added integration tests verifying message-channel replay behavior including correct timing alignment and message batching.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/IsaacTeleop/pull/527)

### DIFF
--- a/src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.cpp
+++ b/src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.cpp
@@ -1,0 +1,206 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "replay_message_channel_tracker_impl.hpp"
+
+#include <mcap/reader.hpp>
+#include <mcap/recording_traits.hpp>
+#include <schema/message_channel_bfbs_generated.h>
+#include <schema/timestamp_generated.h>
+
+#include <iostream>
+#include <string>
+#include <utility>
+
+namespace core
+{
+
+namespace
+{
+
+// Inspect the MCAP summary and pick the highest record count among channels
+// whose topic differs from the message-channel topic we are about to read.
+// The result is the live recorder's effective per-frame record count for any
+// dense tracker (hand / head / controller / ...), which gives us a basis to
+// derive the inter-frame interval below.
+//
+// Returns 0 when no other channel exists; the caller treats that as
+// "no frame-rate signal" and falls back to one-record-per-update emission.
+uint64_t resolve_reference_record_count(const mcap::McapReader& reader, std::string_view self_topic)
+{
+    const auto& stats_opt = reader.statistics();
+    if (!stats_opt.has_value())
+    {
+        return 0;
+    }
+    const auto& channels = reader.channels();
+
+    uint64_t max_count = 0;
+    for (const auto& [channel_id, count] : stats_opt->channelMessageCounts)
+    {
+        const auto channel_it = channels.find(channel_id);
+        if (channel_it == channels.end() || !channel_it->second)
+        {
+            continue;
+        }
+        if (channel_it->second->topic == self_topic)
+        {
+            continue;
+        }
+        if (count > max_count)
+        {
+            max_count = count;
+        }
+    }
+    return max_count;
+}
+
+} // namespace
+
+ReplayMessageChannelTrackerImpl::ReplayMessageChannelTrackerImpl(std::unique_ptr<mcap::McapReader> reader,
+                                                                 std::string_view base_name)
+{
+    // Read the MCAP summary BEFORE handing the reader to McapTrackerViewers
+    // so we can anchor ``recording_start_ns_`` to ``messageStartTime`` and
+    // derive a per-frame interval from the densest non-self channel. That
+    // interval is what makes ``update()``'s frame-counter math align with
+    // the rest of the recording's record stream (each non-message-channel
+    // replay impl emits one record per session update, so N updates ==
+    // recording frame N).
+    //
+    // ``AllowFallbackScan`` recovers when the Summary section is missing
+    // (older or corrupt writers); on total failure we leave the sentinels
+    // in place and ``update()`` falls back to one-record-per-update.
+    const auto summary_status = reader->readSummary(mcap::ReadSummaryMethod::AllowFallbackScan);
+    if (summary_status.ok())
+    {
+        const auto& stats_opt = reader->statistics();
+        if (stats_opt.has_value() && stats_opt->messageCount > 0)
+        {
+            recording_start_ns_ = static_cast<int64_t>(stats_opt->messageStartTime);
+
+            // The first (and currently only) sub-channel name in the traits
+            // is what the live recorder writes under; mirror it here so the
+            // reference-channel search excludes us from its own max search.
+            const std::string self_topic =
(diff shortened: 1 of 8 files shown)


src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.cpp
```diff
@@ -0,0 +1,206 @@
...
+    {
+        ++frame_counter_;
+        return;
+    }
+
+    if (recording_dt_ns_ <= 0 || recording_start_ns_ < 0)
+    {
+        // Fallback: no usable frame-rate reference. Emit one record per
+        // update so the events at least surface in recorded order on
+        // malformed / minimal MCAPs.
```
aristarkhovNV: if an MCAP is truly malformed we should throw on it right away to prevent ingesting bad data

src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    mcap_viewers_ = std::make_unique<MessageChannelMcapViewers>(
+        std::move(reader), base_name,
+        std::vector<std::string>(MessageChannelRecordingTraits::channels.begin(),
+                                 MessageChannelRecordingTraits::channels.end()));
+}
+
+void ReplayMessageChannelTrackerImpl::prime_pending_record()
+{
+    pending_record_ = mcap_viewers_->read(0);
```
aristarkhovNV: for readability it might be better to get rid of `prime_pending_record()` anf replace those function calls with this single line. Also what is 0 arg?

src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.hpp
```diff
@@ -0,0 +1,81 @@
...
+    const MessageChannelMessagesTrackedT& get_messages() const override;
+    void send_message(const std::vector<uint8_t>& payload) const override;
+
+private:
+    void prime_pending_record();
+    static int64_t record_monotonic_ns(const MessageChannelMessagesRecordT& record);
+
+    MessageChannelMessagesTrackedT messages_;
+    std::unique_ptr<MessageChannelMcapViewers> mcap_viewers_;
+    std::optional<MessageChannelMessagesRecordT> pending_record_;
```
aristarkhovNV, line 72: `pending_record_` seems like a temporary object, do we need to keep it in the class?

src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.cpp
```diff
@@ -0,0 +1,206 @@
...
+    if (recording_dt_ns_ <= 0 || recording_start_ns_ < 0)
+    {
+        // Fallback: no usable frame-rate reference. Emit one record per
+        // update so the events at least surface in recorded order on
+        // malformed / minimal MCAPs.
+        if (pending_record_->data)
+        {
+            messages_.data.push_back(std::move(pending_record_->data));
+        }
+        prime_pending_record();
```
aristarkhovNV: prime on next update, then would no longer have to keep `pending_record_` as a member.

src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.cpp
```diff
@@ -0,0 +1,206 @@
...
+                mcap_topic(base_name, std::string(MessageChannelRecordingTraits::channels[0]));
+            const uint64_t reference_count = resolve_reference_record_count(*reader, self_topic);
+
+            if (reference_count > 1)
+            {
+                const int64_t duration_ns = static_cast<int64_t>(stats_opt->messageEndTime) -
+                                            static_cast<int64_t>(stats_opt->messageStartTime);
+                if (duration_ns > 0)
+                {
+                    // Inter-frame interval in the recording. ``count - 1``
```
shaosu-nvidia: I read the comment in the header file and IIUC, the logic here is trying to figure out when it's time to read message channel records and replay. we use this logic to figure out the average delta_t between each frame so I guess we may have problem in the following 2 cases?
1. if the update frequency during live mcap recording changes. For example, we have a live session of 20 seconds and the first half of the live session with 30fps and second half of the live session with 60fps. And we have a opaque channel message at t=5 (which is after 150 frames). And when replay, the recording_dt_ would be 1/45 second. then we will wait for 5 / recording_dt_ = 225 frames to replay the message? (I am not sure if my understanding is correct, let me know).

2. This is also based on the assumption that the live session has at least one tracker that get recorded in each update call (hand/head from cloudxr)?

rwiltz: Yeah that seems wrong... I'll revisit this

shaosu-nvidia: I think an easier way may be recording regardless on the live session side. so if no message, record a no message or data null, as not available

src/core/replay_trackers/cpp/replay_message_channel_tracker_impl.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+void ReplayMessageChannelTrackerImpl::update(int64_t /*monotonic_time_ns*/)
+{
+    // Each update consumes exactly one recorded frame: all records
+    // sharing the first pending record's timestamp. See the class
+    // docstring for the invariant this relies on (the live recorder
+    // writes ≥1 record per session.update()).
+    messages_.data.clear();
+
+    if (!pending_record_)
```
shaosu-nvidia, line 50: alternatively we can vibe code a peek() function in McapTrackerViewers but this is also okay to me.

shaosu-nvidia (approved):

PR #529, opened 2026-05-14
ROS: Check that the poses are valid before using them

sgrizan-nv:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced hand pose construction with per-joint validity detection
  * Invalid joints now properly emit zero-position poses instead of being skipped
  * Improved controller and hand pose validation logic for teleoperation systems
  * Wrist frame publishing now based on joint validity state

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/IsaacTeleop/pull/529)

### DIFF
--- a/examples/teleop_ros2/python/teleop_ros2_node.py
+++ b/examples/teleop_ros2/python/teleop_ros2_node.py
@@ -157,15 +157,19 @@ def _append_hand_poses(
     poses: List[Pose],
     joint_positions: np.ndarray,
     joint_orientations: np.ndarray,
+    joint_valid: np.ndarray,
     transform_rot: Rotation | None = None,
     transform_trans: Sequence[float] | None = None,
 ) -> None:
     for joint_idx in range(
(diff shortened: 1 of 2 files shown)


examples/teleop_ros2/python/teleop_ros2_node.py
```diff
@@ -157,15 +157,19 @@ def _append_hand_poses(
     poses: List[Pose],
     joint_positions: np.ndarray,
     joint_orientations: np.ndarray,
+    joint_valid: np.ndarray,
```
kajananchinniahNV, line 160: Nit: call this joints_valid or something else to make it more clear this is an array

kajananchinniahNV (approved):

examples/teleop_ros2/python/teleop_ros2_node.py
sgrizan-nv: The array is called JOINT_VALID in the OpenXR joint index, and I'm trying to keep naming consistent

PR #534, opened 2026-05-14
viz/layers: stereo QuadLayer (paired mailbox + per-eye binding)

farbod-nv:
QuadLayer::Config gains ``stereo`` and ``stereo_baseline_m``. Stereo layers allocate two DeviceImages per mailbox slot (left + right) and expose a two-arg submit:

    void submit(left, right, stream = 0);

Both eyes are copied + the single cuda_done_writing signal is emitted on the same CUDA stream, so stream ordering — not a second semaphore — is what guarantees the renderer never sees a half-matched pair. Strict invariants: submit(src) throws std::logic_error on a stereo layer, submit(left, right) throws on a mono layer.

record() in kXr binds the left descriptor for view 0 and the right for view 1. Window / offscreen (single-view) renders left only. Baseline disparity is applied host-side by translating each eye's placement position by ±stereo_baseline_m/2 along the placement's local +x axis before composing the MVP — no shader change.

Mip generation runs on both eyes in the pre-render-pass when stereo
+ generate_mipmaps are both on. Descriptor pool sizing + writes double when stereo.

Memory: 2× per-slot allocation for stereo layers (~112 MB at 1080p RGBA8 + 7 slots vs ~56 MB mono).

Tests (catch2, gpu-gated where needed):
- ctor rejects non-finite stereo_baseline_m
- stereo allocates paired DeviceImages for every slot (distinct vk_image + cuda_array per eye)
- mono device_image_right is null
- mono submit(left, right) throws
- stereo submit(left) throws
- stereo submit lands matching L/R pair in the latest slot
- stereo rapid submits keep every L/R pair atomic (decodes the per-submit index from the left pixel and proves the right pixel is the matching one for every written slot — covers torn writes
  + cross-eye aliasing in one shot)

### DIFF
--- a/examples/camera_viz/camera_streamer.py
+++ b/examples/camera_viz/camera_streamer.py
@@ -30,7 +30,7 @@ from typing import List, Optional
 import yaml
 
 from pipeline import FrameSource
-from sources import build_local_camera
+from sources import PairedFrameSource, build_local_camera
 from transports import RtpH264Sender, make_encoder
 
 logger = logging.getLogger("camera_streamer")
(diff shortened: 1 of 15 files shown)


nv-jakob:
Are we still looking to land this?

farbod-nv:
> Are we still looking to land this?

ah i think this was just experimental, gonna close this. the one merged already supports stereo cameras with per-eye bindings.

PR #579, opened 2026-05-29
fix(cloudxr): add opt-in to avoid autoTSSkey mapping abort

nv-mhaselton:
Add NV_CXR_RUNTIME_JOIN_MAIN_THREAD to join the runtime on the main thread instead of a worker thread, avoiding a "Couldn't create autoTSSkey mapping" abort seen on some platforms. Default keeps the original worker-thread behavior.

### DIFF
--- a/src/core/cloudxr/python/runtime.py
+++ b/src/core/cloudxr/python/runtime.py
@@ -266,16 +266,23 @@ def run() -> None:
     state["service_created"] = True
     lib.nv_cxr_service_start(svc)
 
-    # Run the blocking join() in a worker thread so the main thread stays in Python
-    # and can run the signal handler. Otherwise Ctrl+C is not processed while we're
-    # inside the native nv_cxr_service_join() call.
-    def join_then_destroy() -> None:
+    join_on_main = os.environ.get("NV_CXR_RUNTIME_JOIN_MAIN_THREAD", "").strip().lower()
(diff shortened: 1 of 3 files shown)


src/core/cloudxr/python/runtime.py
```diff
@@ -266,16 +266,23 @@ def stop(sig: int, frame: object) -> None:
     state["service_created"] = True
     lib.nv_cxr_service_start(svc)
 
-    # Run the blocking join() in a worker thread so the main thread stays in Python
-    # and can run the signal handler. Otherwise Ctrl+C is not processed while we're
-    # inside the native nv_cxr_service_join() call.
-    def join_then_destroy() -> None:
+    join_on_main = os.environ.get("NV_CXR_RUNTIME_JOIN_MAIN_THREAD", "").strip().lower()
```
jiwenc-nv, line 269: Do we really need the feature flag here?  It will add burdens for users (then our support on educating users) to set the flag accordingly.

Can handle this automatically based on whether we are capable of "Couldn't create autoTSSkey mapping"?

nv-mhaselton: agreed, this is a draft branch, and the feature flag should not exist to begin with

nv-mhaselton:
Closed in favor of https://github.com/NVIDIA/IsaacTeleop/pull/808 and https://github.com/NVIDIA/IsaacTeleop/pull/816

Issue #578, opened 2026-05-29
Control pane to start teleop sometimes didn't appear in view


jiwenc-nv:
**Describe the bug**

The control pane to start teleop sometimes didn't appear in view.

Sometimes a small circle with "Show" tag appeared, but sometime nothing appeared and I would have to remove the headset and then wear again (keep trying)

** 100% Reproduceable on Pico **
Steps to reproduce the behavior in Pico Browser:
1. Open https://nvidia.github.io/IsaacTeleop/client
2. Connect to a CloudXR session for the first time
3. Observed that the "Play" screen pops out -> OK
4. Click "disconnect" to exit from a CloudXR session
5. Connect to a CloudXR session for the 2nd time
6. Observed that the "Play" screen DOES NOT pops out -> BAD!!!
7. Exit VR my building the menu button
8. Refresh the web page
9. Connect to a CloudXR session for the 3rd time
10. Observed that the "Play" screen pops out -> OK

**Expected behavior**

"Play" screen pops out all the time

**Observed behavior**

Observed that the "Play" screen DOES NOT pops out after "disconnect"

**Screenshots**

**Platforms**
 - Isaac Teleop Version: 1.2.9rc1 / 1.3.55rc1
 - OS: Ubuntu X86
 - Integrations: [e.g. Isaac Lab, Isaac Sim, ROS2]
 - XR Device: Pico 4 Ultra
 - Additional Devices [e.g. Manus]

**Additional context**

jiwenc-nv:
https://github.com/user-attachments/assets/683871bf-c586-4df8-9d48-4341918a78e4

yanziz-nvidia:
The second one is non-deterministic. See https://github.com/NVIDIA/IsaacTeleop/pull/586 for a workaround of this problem if you are not on the enterprise activated PICO which should have fixed the issue properly. It's PICO browser issue with R3F.

yanziz-nvidia:
this should be fixed now, close? @jiwenc-nv

PR #586, opened 2026-05-30
feat(webclient): add auto-refresh-on-session-end option

yanziz-nvidia:
## Summary
Adds a configurable **Auto Refresh on Session End** option to the web client's 2D settings UI, so the page can reload itself when the XR session ends and return to a clean initial state.

The option is a three-way selector rather than a boolean, letting users choose the behavior they want:

| Mode | Behavior |
|------|----------|
| `never` | Page never reloads |
| `clean` (default) | Reload only on an error-free end (Disconnect button, headset exit, server stop) |
| `any` | Reload on any end, **including** a streaming error |

## Implementation
- New `AutoRefreshMode` type + `parseAutoRefreshMode()` helper in `helpers/react/utils.ts` (mirrors the existing `parseControlPanelPosition` pattern for safe parsing of stored/URL/form values).
- Refresh is handled centrally in `App.tsx` `handleStatusChange`, keyed off the CloudXR stream-stop status (`Disconnected` vs `Error`), so it covers **every** session-end path — not just the Disconnect button. Pre-stream connection failures are excluded since they aren't session ends. Config is read live off the stable 2D UI object to avoid a stale closure.
- New `<select id="cloudxrAutoRefreshMode">` in `index.html`, wired through `CloudXR2DUI.tsx` (element lookup, default, `enableLocalStorage` persistence, change listener, URL-seed handling, and inclusion in the built config).
- Persists in `localStorage` and is overridable via the `autoRefreshMode` URL query param.

## Testing
- `tsc --noEmit` passes (the only error is a pre-existing, unrelated one in `PerformanceCanvasImage.tsx`).
- Pre-commit hooks (REUSE/SPDX, copyright year, etc.) pass.

## Summary by CodeRabbit

* **New Features**
  * Added "Auto Refresh on Session End" configuration option
  * Three modes: never refresh, refresh on clean session end (default), or refresh on any session end
  * Setting can be configured via dropdown, URL parameter, and is persisted locally
  * Automatic page reload on stream/session stop or error based on selected mode

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/IsaacTeleop/pull/586?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/deps/cloudxr/webxr_client/helpers/react/utils.ts
+++ b/deps/cloudxr/webxr_client/helpers/react/utils.ts
@@ -64,8 +64,30 @@ export interface ReactUIConfig {
   panelHiddenAtStart?: boolean;
   /** When true, all WebGL rendering is skipped. */
   headless?: boolean;
+  /** Page-refresh behavior when the XR session ends. See {@link AutoRefreshMode}. */
+  autoRefreshMode?: AutoRefreshMode;
   /** Active teleop project path (a key path in `TELEOP_PROJECTS`). */
   teleopPath: string;
+}
+
+/** When to reload the page once the XR session ends: never, only on a clean end, or on any end. */
+export type AutoRefreshMode = 'never' | 'clean' | 'any';
+
+const AUTO_REFRESH_MODES: readonly AutoRefreshMode[] = ['never', 'clean', 'any'];
+
+/**
+ * Parses a string into a valid auto-refresh mode.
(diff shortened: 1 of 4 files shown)


jiwenc-nv (approved):

jiwenc-nv:
fixes https://github.com/NVIDIA/IsaacTeleop/issues/578

Issue #452, opened 2026-05-01
libnlopt-dev system dependency undocumented in Quick Start and Build from Source guides


hesmilla1:
**Description:** No pre-built nlopt wheel exists for linux/aarch64 on PyPI, so pip falls back to building from source. That build requires libnlopt-dev to be installed on the system, but it is not listed as a prerequisite in either the [https://nvidia.github.io/IsaacTeleop/main/getting_started/build_from_source/index.htm](https://nvidia.github.io/IsaacTeleop/main/getting_started/quick_start.html)or [https://nvidia.github.io/IsaacTeleop/main/getting_started/build_from_source/index.htm](https://nvidia.github.io/IsaacTeleop/main/getting_started/build_from_source/index.htm) docs.

**To reproduce**
Follow the Quick Start guide: pip install 'isaacteleop[cloudxr,retargeters]~=1.0.0' --extra-index-url https://pypi.nvidia.com/

**Expected behavior**
Package installs successfully

**Observed behavior**
Build fails with could not build wheels for nlopt

**Screenshots**
<img width="1613" height="1325" alt="Image" src="https://github.com/user-attachments/assets/85f1c2a9-a6cb-4b38-ae7b-b3925c9d40f2" />
<img width="1618" height="1346" alt="Image" src="https://github.com/user-attachments/assets/e4f4b98a-7454-4fa8-94ca-807317cffca8" />
<img width="1621" height="1279" alt="Image" src="https://github.com/user-attachments/assets/2f7fb813-d3cf-486d-b066-88a4cd2bc0b2" />

**Platforms**
 - Isaac Teleop Version: N/A
 - OS: aarch64 (NVIDIA DGX Spark, Ubuntu)
 - Integrations: Isaac Teleop
 - XR Device: N/A

**Additional context**
I ran the steps on the build from source page, but I am getting multiple errors. It's not finding the c library, so the python wheel can't be built.

yanziz-nvidia:
Can repro

```
$ uv pip install 'isaacteleop[cloudxr,retargeters]~=1.0.0' --extra-index-url https://pypi.nvidia.com
Resolved 62 packages in 1.54s
  × Failed to build `nlopt==2.6.2`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit status: 1)

      [stdout]
      running bdist_wheel
      running build
      running build_ext
      cmake version 3.28.3

[...]

PR #590, opened 2026-05-30
wip: bundle nlopt into isaacteleop wheel [do not merge — LGPL]

jiwenc-nv:
## Why this exists

This PR records a dead end for future reference.

Bundling nlopt (LGPL v2.1) directly into the isaacteleop wheel was explored as a zero-friction fix for aarch64 (`pip install isaacteleop[retargeters]` failing because PyPI has no aarch64 nlopt wheel — see #452).

## Why it cannot be merged

nlopt is **LGPL v2.1**. Bundling it into a proprietary NVIDIA wheel would require:
- Shipping the LGPL license text with every distribution
- Guaranteeing users can relink against a modified nlopt `.so`
- Legal review / sign-off

The LGPL relinking requirement is incompatible with a tightly-packaged binary wheel without additional legal work.

## Alternative (tracked in #589)

Build the nlopt wheel as a separate artifact via CMake ExternalProject so it can be distributed under its own license, and users supply it via `--find-links`.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -42,7 +42,7 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y build-essential cmake libx11-dev clang-format-14 ccache libvulkan-dev glslang-tools \
           libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev libxext-dev libxkbcommon-dev \
-          libwayland-dev wayland-protocols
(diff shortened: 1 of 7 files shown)


jiwenc-nv:
Functionally this works well by bundling nlopt into IsaacTeleop wheel.  However, we can't ship this as nlopt is LGPL v2.1, we can't bundle and ship it within the IsaacTeleop wheel due to license concerns.

PR #606, opened 2026-06-02
refactor: schema - colocate flatbuffer generation with the library (no ../), drop unused tensor_utils.h

aristarkhovNV:
## Release Notes

* **Refactor**
  * Reorganized schema build infrastructure to consolidate FlatBuffers header generation into a unified location, improving build efficiency and maintainability
  * Updated internal schema reference paths to align with restructured directories
  * Removed unused utility functions to reduce code complexity

* **Chores**
  * Streamlined build configuration by removing redundant build directives


jiwenc-nv (approved):

jiwenc-nv (changes requested):
actually, having fbs inside the cpp folder reads a bit weird.  to some degree fbs itself can be treated as another language like cpp and python right?

aristarkhovNV:
> actually, having fbs inside the cpp folder reads a bit weird. to some degree fbs itself can be treated as another language like cpp and python right?

updated, PTAL

jiwenc-nv (approved):
actually, having fbs inside the cpp folder reads a bit weird.  to some degree fbs itself can be treated as another language like cpp and python right?

PR #618, opened 2026-06-03
ROS: publish head pose

sgrizan-nv:
Wire the existing HeadSource into the hand_teleop and controller_teleop session pipelines and surface the head pose as a new xr_teleop/head_pose PoseStamped topic plus a world_frame -> head_frame TF, mirroring how wrist poses are handled.

## Summary by CodeRabbit

* **New Features**
  * Added head pose tracking and publishing to the ROS 2 teleop example in both controller and hand modes.
  * Added support for MCAP replay mode testing without live XR hardware.

* **Documentation**
  * Updated ROS 2 teleop guides to document head pose topic and parameters.
  * Added Docker validation instructions and MCAP replay workflow examples.

* **Chores**
  * Enhanced development container with pre-commit hook support for automated code quality checks.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,12 +4,14 @@
   "features": {
     "ghcr.io/devcontainers/features/git:1": {},
     "ghcr.io/devcontainers/features/git-lfs:1": {},
+    "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {},
     "ghcr.io/devcontainers-extra/features/apt-packages:1": {
       "packages": "clang-format-14,libx11-dev,libxrandr-dev,libxxf86vm-dev,libgl1-mesa-dev"
     },
     "ghcr.io/devcontainers-extra/features/cmake:1": {},
     "ghcr.io/devcontainers-extra/features/uv:1": {}
   },
+  "postCreateCommand": "uv tool install pre-commit && pre-commit install",
   "customizations": {
     "vscode": {
       "extensions": [
--- a/examples/teleop_ros2/cpp/integration_tests/mcap_generator.cpp
+++ b/examples/teleop_ros2/cpp/integration_tests/mcap_generator.cpp
@@ -7,6 +7,7 @@
 #include <schema/controller_generated.h>
 #include <schema/full_body_generated.h>
 #include <schema/hand_generated.h>
+#include <schema/head_generated.h>
 #include <schema/pedals_generated.h>
 #include <schema/timestamp_generated.h>
 
@@ -24,6 +25,7 @@ namespace
 
 using ControllerChannels = core::McapTrackerChannels<core::ControllerSnapshotRecord, core::ControllerSnapshot>;
 using HandChannels = core::McapTrackerChannels<core::HandPoseRecord, core::HandPose>;
+using HeadChannels = core::McapTrackerChannels<core::HeadPoseRecord, core::HeadPose>;
 using PedalChannels = core::McapTrackerChannels<core::Generic3AxisPedalOutputRecord, core::Generic3AxisPedalOutput>;
 using FullBodyChannels = core::McapTrackerChannels<core::FullBodyPosePicoRecord, core::FullBodyPosePico>;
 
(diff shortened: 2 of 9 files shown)


.devcontainer/devcontainer.json
```diff
@@ -4,12 +4,17 @@
   "features": {
     "ghcr.io/devcontainers/features/git:1": {},
     "ghcr.io/devcontainers/features/git-lfs:1": {},
+    "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
+      "version": "latest",
```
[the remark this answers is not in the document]
sgrizan-nv: Done

examples/teleop_ros2/cpp/integration_tests/mcap_generator.cpp
```diff
@@ -74,6 +76,16 @@ std::shared_ptr<core::HandPoseT> make_hand_sample(bool left, int frame)
     return sample;
 }
 
+std::shared_ptr<core::HeadPoseT> make_head_sample(int frame)
+{
+    const float delta = 0.0005f * static_cast<float>(frame);
```
kajananchinniahNV, line 86: Nit: it's unclear to me where `0.0005f` is from. Can we make this a constexpr variable?

examples/teleop_ros2/cpp/integration_tests/mcap_generator.cpp
```diff
@@ -74,6 +76,16 @@ std::shared_ptr<core::HandPoseT> make_hand_sample(bool left, int frame)
...
+    auto sample = std::make_shared<core::HeadPoseT>();
+    sample->pose =
+        std::make_shared<core::Pose>(core::Point(0.0f, 0.10f + delta, 1.60f), core::Quaternion(0.0f, 0.0f, 0.0f, 1.0f));
```
kajananchinniahNV, line 89: Similarly can we make constexpr expressions for the other numbers?

examples/teleop_ros2/python/node_parameters.py
```diff
@@ -423,7 +423,7 @@ def create_node_parameters(node: Node) -> NodeParameters:
     session_mode, mcap_config = _load_mcap_replay(node)
     cloudxr_install_dir, cloudxr_env_config, cloudxr_accept_eula = _load_cloudxr(node)
     pedal_collection_id = _load_pedal_collection_id(node)
-    world_frame, right_wrist_frame, left_wrist_frame = _load_frames(node)
+    world_frame, right_wrist_frame, left_wrist_frame, head_frame = _load_frames(node)
```
kajananchinniahNV, line 426: I'm almost tempted to have this as a dict instead of as 4 params - but I'm not confident here

What do you think?

examples/teleop_ros2/cpp/integration_tests/mcap_generator.cpp
sgrizan-nv: These aren't domain constants — they're throwaway fixture values. The only real requirements are "valid, finite, non-zero, slightly time-varying." In my opinion naming them implies a significance they don't actually carry (they're arbitrary).

examples/teleop_ros2/python/node_parameters.py
sgrizan-nv: They are used later as individual variables, so in my opinion deconstruction makes sense. If they're aggregated into a dict, we'll have to extract values by key again instead of just accessing variables

kajananchinniahNV (approved):
LGTM beyond some minor comments!

examples/teleop_ros2/python/node_parameters.py
kajananchinniahNV: Sounds good, resolving it

examples/teleop_ros2/cpp/integration_tests/mcap_generator.cpp
kajananchinniahNV: Could we leave some sort of comment mentioning that and what / why they are here then?

sgrizan-nv: Sure, done

PR #629, opened 2026-06-05
fix(webclient): fix XR panel snapping to face away from user when moved

### LINKED ISSUE #546: IsaacTeleop WebXR client thumbsticks move and gets stuck
**Describe the bug**

**To reproduce**
1. Connect with the IsaacTeleop WebXR client
2. Wiggle controller thumbstick on one of the controllers

**Expected behavior**
* Controller thumbstick doesn't move when the other's is moving.
* Controller thumbstick doesn't get stuck.

**Observed behavior**
* The other controller's thumbstick moving when the first's is moving.
* The controller's thumbstick sometimes getting stuck in the right position when let go.

**Screenshots**

**Platforms**
 - Isaac Teleop Version: N/A
 - OS: Meta Horizon OS
 - Integrations: N/A
 - XR Device: Quest3
 - Additional Devices: Touch Plus Controllers

**Additional context**
N/A

yanziz-nvidia:
Closes #546

## Summary

- **Bug 1 (yaw accuracy):** The old code used `setFromUnitVectors(zAxis, dir3D)` + `setFromQuaternion('YXZ')` to extract horizontal yaw. Because `dir3D` included the camera's height offset, the extracted Euler Y was not the true horizontal yaw. Replaced with `Math.atan2(dx, dz)` on the XZ plane.
- **Bug 2 (±π wrap):** The face-camera `damp()` call had no angle wrapping. When the target yaw crossed the ±π boundary, `damp()` took the ~2π wrong-way path, spinning the panel to face backward. Normalized `diff` to `[-π, π]` before calling `damp()`.
- **Bug 3 (Handle quaternion reset — root cause of 90° snap after drag):** `@pmndrs/handle`'s `defaultApply` copies `state.current.quaternion` to the target on every drag frame (useFrame priority -1, before face-camera at priority 0) and again on drag release. With `rotate={false}`, that quaternion is always the one captured at drag-start, so it continuously reset the face-camera rotation mid-drag and wiped it entirely on release — snapping the panel up to 90° off after every drag. Fixed by passing a custom `apply` function (`applyPositionSkipRotation`) to `Handle` that copies only position, leaving quaternion/rotation entirely under face-camera control.

## Test plan

- [x] Enter XR session and drag the panel to the left or right — panel should rotate to face you as it moves
- [x] Release the handle — panel should stay facing you, not snap back to a wrong angle
- [x] Move/walk around the panel in the emulator — panel should track you continuously
- [x] Walk behind the panel — it should smoothly rotate to re-face you
- [x] Verify normal forward-facing behaviour is unchanged when standing still

### DIFF
--- a/deps/cloudxr/webxr_client/src/CloudXRUI.tsx
+++ b/deps/cloudxr/webxr_client/src/CloudXRUI.tsx
@@ -39,11 +39,11 @@ import { PerformanceCanvasImage } from '@helpers/react/PerformanceCanvasImage';
 import { useXRButton } from '@helpers/react/useXRButton';
 import { ReadonlySignal } from '@preact/signals-react';
 import { useFrame } from '@react-three/fiber';
-import { Handle, HandleTarget } from '@react-three/handle';
+import { Handle, HandleTarget, HandleState } from '@react-three/handle';
 import { Container, Text, Image } from '@react-three/uikit';
 import { Button } from '@react-three/uikit-default';
 import React, { useRef, useState, useEffect } from 'react';
-import { Color, Group, Mesh, MeshStandardMaterial, Vector3 } from 'three';
+import { Color, Group, Mesh, MeshStandardMaterial, Object3D, Vector3 } from 'three';
 import { damp } from 'three/src/math/MathUtils.js';
 
 // Face-camera rotation constants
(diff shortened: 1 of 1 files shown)


yanziz-nvidia:
See https://nvidia.github.io/IsaacTeleop/preview/pr-629/client/#/sim

gareth-morgan-nv (approved):
LGTM.  Maybe add a test?

PR #632, opened 2026-06-05
camera_viz: add --mode override and fix setup pipefail bugs

nv-mhaselton:
Add a --mode {window,xr} flag to camera_viz.py that overrides display.mode from the YAML, and forward extra args through camera_viz.sh run, so one config drives both a local window and a headless XR (CloudXR) run without editing the file.

Fix two `set -o pipefail` bugs in _install_deps.sh that blocked setup:
- the Jetson cuda-nvrtc probe used `find /usr ... | grep -q`, whose pipeline fails when find hits an unreadable /usr subdir, falsely flagging cuda-nvrtc as missing even when libnvrtc is present;
- the isaacteleop dist-info probe `ls ... | head -1` aborted the first full install on a fresh venv (glob matches nothing) before it ran.

## Summary by CodeRabbit

* **New Features**
  * Add CLI option to set display mode (window or XR) that overrides config; startup now reports the effective mode.

* **Bug Fixes**
  * Ensure additional CLI flags are forwarded to the visualization command.
  * Improve Jetson dependency detection and installation checks to avoid failures on systems with unreadable directories or missing package metadata.

### DIFF
--- a/examples/camera_viz/camera_viz.py
+++ b/examples/camera_viz/camera_viz.py
@@ -179,9 +179,10 @@ def _build_rtp_entries(cfg: dict, is_xr: bool) -> List[SourceEntry]:
     return entries
 
 
-def _make_session(cfg: dict) -> viz.VizSession:
+def _make_session(cfg: dict, mode_override: Optional[str] = None) -> viz.VizSession:
     display = cfg.get("display", {})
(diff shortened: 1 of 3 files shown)


farbod-nv (approved):
thanks!

PR #691, opened 2026-06-22
Add OGLO tactile glove device (plugin, tracker, schema, example)

styu12:
## Summary

Adds **OGLO tactile gloves** as a first-class IsaacTeleop device, following the
[Add a New Device](https://github.com/NVIDIA/IsaacTeleop/blob/main/docs/source/device/add_device.rst)
pattern (FlatBuffer schema → plugin `SchemaPusher` → tracker → MCAP), modeled on the
`generic_3axis_pedal` reference.

Each glove streams **80 taxels** (5 fingers × 4×4) + a **6-axis IMU** at **100 Hz** over BLE.
A MetaQuest (hand/head) and both gloves record into **one time-synchronized MCAP** on a shared
`DeviceDataTimestamp` clock, with an optional in-headset tactile heatmap overlay (TeleViz).

In-headset CloudXR view: OGLO gloves on the operator's hands with the live two-hand taxel heatmap (bottom-right) composited into the Quest view

## What's included

- **Schema** — `oglo_tactile.fbs` (`OgloGloveSample` / `…Tracked` / `…Record`) + C++/Python bindings + schema pytest.
- **Plugin** — `src/plugins/oglo_tactile`: BLE read → config-driven parser → push (Mode 2) or local MCAP (Mode 1).
- **Tracker** — `OgloTactileTracker` + `LiveOgloTactileTrackerImpl`, factory registration, recording traits.
- **Example + docs** — end-to-end record demo, heatmap renderer, Foxglove exporter, and `docs/source/device/oglo.rst`.

## BLE backend

The plugin talks to **BlueZ directly over libdbus (AFL-2.1, permissive)** — no copyleft dependency,
and it connects out of the box on any Linux host with the BlueZ daemon (`apt install libdbus-1-dev`).
The transport is isolated behind the `OgloBleClient` interface, so an alternative backend can be
dropped in via `make_ble_client()` without touching the parser, schema, or tracker.

**Verified on hardware** — both gloves connect and stream at 100 Hz into a synchronized MCAP
(the screenshot above is a live capture).

*Scope: BLE-only. A USB-C serial transport could slot into the same factory as future work.*

## Type of change

- [x] New feature (non-breaking)
- [x] Documentation update

## Testing

```bash
ctest --test-dir build -R oglo_packet_parser --output-on-failure   # wire decode vs firmware packing ref
ctest --test-dir build -R schema_test_oglo_tactile --output-on-failure  # schema bindings
```
- The parser test is dependency-free (`g++ -std=c++20 … && ./t` → `OK: all 350 checks passed`).
- **Manual:** MetaQuest (CloudXR) + two gloves → one MCAP with `hands/{left,right}`, `head/head`,
  `oglo_{left,right}`, all on `sample_time_local_common_clock`. Steps in `examples/oglo_tactile/README.md`.
- Tooling matched to CI: `clang-format-14`, `ruff`, and REUSE/SPDX all clean.

## Checklist

- [x] Read the [contribution guidelines](https://github.com/NVIDIA/IsaacTeleop/blob/main/CONTRIBUTING.md)
- [x] Ran `ruff` and `clang-format-14` (CI's version)
- [x] Updated documentation
- [x] Added tests
- [x] Signed off all commits (DCO)

---

Plugin is **off by default** (`-DBUILD_PLUGIN_OGLO=ON` to enable), Linux-only. Happy to open a
tracking issue or adjust the review shape if maintainers prefer.


src/plugins/oglo_tactile/oglo_tactile_plugin.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include "oglo_packet_parser.hpp"
+
+#include <atomic>
+#include <condition_variable>
+#include <cstdint>
+#include <deque>
+#include <memory>
+#include <mutex>
+#include <string>
+
```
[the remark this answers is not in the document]
styu12: Fixed in `65e0224`: added `#include <chrono>` and `#include <cstddef>` so the header is self-contained and no longer relies on transitive includes.

src/plugins/oglo_tactile/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+`test_oglo_packet_parser` validates the wire decode against the firmware's own
+12-bit packing reference (round-trip), plus the schema-4 fallback and
+malformed-packet rejection:
+
+```bash
+ctest --test-dir build -R oglo_packet_parser --output-on-failure
+# or standalone, no IsaacTeleop deps:
+cd src/plugins/oglo_tactile/tests
+g++ -std=c++17 -I.. test_oglo_packet_parser.cpp ../oglo_packet_parser.cpp -o t && ./t
```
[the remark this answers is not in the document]
styu12: Fixed in `65e0224`: aligned the standalone build command to `-std=c++20` to match the test file comment and the rest of the docs.

examples/oglo_tactile/oglo_teleop_record.py
```diff
@@ -0,0 +1,371 @@
...
+    vv = np.array(v, dtype=float)
+    t = 2.0 * np.cross(u, vv)
+    return vv + w * t + np.cross(u, t)
+
+
+def _head_locked_placement(
+    head_tracker: "deviceio.HeadTracker",
+    session: "viz.VizSession",
+    args: argparse.Namespace,
+) -> "viz.QuadLayerPlacement | None":
```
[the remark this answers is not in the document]
styu12: Fixed in `65e0224`: corrected the hint to `deviceio.DeviceIOSession`, which is the object `DeviceIOSession.run()` yields and is-a `ITrackerSession` (what `get_head()` consumes).

src/plugins/oglo_tactile/main.cpp
```diff
@@ -0,0 +1,132 @@
...
+            {
+                ms = -1; // force the validation error below
+            }
+            if (ms <= 0)
+            {
+                std::cerr << "Error: --scan-timeout-ms expects a positive integer (got '" << val << "')." << std::endl;
+                print_usage(argv[0]);
+                return 1;
+            }
+            opts.scan_timeout = std::chrono::milliseconds(ms);
```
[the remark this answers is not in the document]
styu12: Fixed in `65e0224`: now `std::stoi(val, &parsed)` and reject the value when `parsed != val.size()`, so trailing garbage like `15000ms` is no longer silently accepted.

src/plugins/oglo_tactile/oglo_tactile_plugin.cpp
```diff
@@ -0,0 +1,155 @@
...
+            catch (const std::exception& e)
+            {
+                std::cerr << "OGLO: reconnect failed: " << e.what() << " — retrying." << std::endl;
+                std::this_thread::sleep_for(std::chrono::milliseconds(500));
+            }
+        }
+    }
+
+    std::cout << "OGLO: stopped after " << m_total_samples << " samples." << std::endl;
+}
```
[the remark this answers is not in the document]
styu12: Good catch — fixed in `65e0224`. After the run loop exits on `stop`, we now call `m_ble->disconnect()` to quiesce the BLE producer thread and then drain the remaining queue into the sink, so tail samples are persisted instead of dropped.

src/core/mcap/cpp/inc/mcap/recording_traits.hpp
```diff
@@ -52,6 +52,13 @@ struct PedalRecordingTraits
     static constexpr std::array replay_channels = { "pedals_tracked" };
 };
 
+struct OgloRecordingTraits
+{
+    static constexpr std::string_view schema_name = "core.OgloGloveSampleRecord";
+    static constexpr std::array recording_channels = { "oglo", "oglo_tracked" };
+    static constexpr std::array replay_channels = { "oglo_tracked" };
```
jiwenc-nv, line 59: **please fix:** `replay_channels` is declared, but there's no `ReplayOgloTactileTrackerImpl` and no dispatch entry in `replay_deviceio_factory.cpp` — constructing a `ReplaySession` with an `OgloTactileTracker` over a recorded MCAP throws `std::invalid_argument("unsupported tracker type")`. The invariant elsewhere is *replay_channels ⇔ replay impl exists* (head/hand/controller/full-body/pedal all have both; OAK deliberately omits the field).

Either add the replay impl mirroring `replay_generic_3axis_pedal_tracker_impl.{hpp,cpp}` (~40 lines + factory entry — preferred, since record→replay is the primary offline workflow for a research device), or drop this line per the OAK precedent and note the gap in `oglo.rst`.

docs/source/device/oglo.rst
```diff
@@ -0,0 +1,94 @@
...
+   ./build/src/plugins/oglo_tactile/oglo_tactile_plugin --side right --collection-prefix=oglo
+
+The packet parser reads the device **Config characteristic** first and branches
+on the notify ``flags`` byte (``0x04`` packed12 v5 — primary; ``0x02``/``0x01``
+legacy schema-4 fallback), so payload sizes are never hardcoded.
+
+Recorded data
+-------------
+
+Channels ``oglo_left`` / ``oglo_right`` carry ``core.OgloGloveSampleRecord``:
```
jiwenc-nv: Maybe I'm confused... but these seem to be the Mode 1 channel names. Mode 2 session recordings (what the example produces) write `oglo_left/oglo` and `oglo_left/oglo_tracked` (via `mcap_topic()` + `OgloRecordingTraits`), so a new user's first `mcap info` check fails against this page.

Related: a newcomer meets five spellings of glove identity with no stated derivation — `--collection-prefix=oglo` (CLI) → collection `oglo/left` (tracker ctor) → recording prefix `oglo_left` (`McapRecordingConfig`) → channel `oglo_left/oglo` (MCAP) → `/glove/oglo_left` (plugin.yaml). One small table here stating that derivation (and the per-mode channel names) would resolve both issues.

examples/oglo_tactile/oglo_teleop_record.py
```diff
@@ -0,0 +1,371 @@
...
+            )  # verify cupy actually works
+            cp = _cp
+            overlay = _make_overlay_layer(viz_session, args)
+        except Exception as e:
+            print(
+                f"WARNING: headset overlay disabled — cupy unusable ({e}). Recording continues.",
+                flush=True,
+            )
+            print(
+                "  Fix: pip uninstall -y cupy-cuda11x cupy-cuda12x && pip install cupy-cuda11x",
```
jiwenc-nv: **nit:** this fix message pins `cupy-cuda11x`, but the README setup says `pip install cupy-cuda12x`. A user who followed the README and hits this warning will install the other variant with no explanation. Make the message CUDA-version-neutral ("install the cupy build matching your CUDA driver") or align both on one recommendation.

src/core/schema/fbs/oglo_tactile.fbs
```diff
@@ -0,0 +1,54 @@
...
+  device_time_us: uint32 (id: 1);
+
+  // 80 taxel readings in "finger,row,col" order (5 fingers x 4 rows x 4 cols).
+  // Raw 12-bit ADC counts in [0, 4095]; higher == more force. The anatomical
+  // finger order is side-aware and is reported by the device Config characteristic
+  // (recorded as MCAP metadata at connection time).
+  taxels: [uint16] (id: 2);
+
+  // 6-axis IMU, raw LSB (no on-device fusion in schema 5).
+  // Accelerometer (~0.000244 g/LSB at +/-8 g).
```
jiwenc-nv: **nit:** the accel scale is documented (checks out for ±8 g) but the IMU's mounting frame is not — nothing says which way X/Y/Z point relative to the glove/hand. Without that, the IMU fields are unusable beyond activity detection. One comment block naming the axes (REP-145-style) fixes it; please also state the gyro scale explicitly like the accel (±2000 dps FS ⇒ ≈0.061 dps/LSB).

src/plugins/oglo_tactile/ble/oglo_ble_client.hpp
```diff
@@ -0,0 +1,75 @@
...
+    //! @throws std::runtime_error on scan/connect/read failure or timeout.
+    virtual std::string connect(Side side, std::chrono::milliseconds timeout) = 0;
+
+    //! Subscribe to the notify characteristic. @p cb runs on the BLE thread.
+    virtual void subscribe(NotifyCallback cb) = 0;
+
+    //! Register an optional connection-state observer (connect / drop events).
+    virtual void on_state_change(StateCallback cb) = 0;
+
+    virtual bool is_connected() const = 0;
```
jiwenc-nv: **please fix:** `is_connected()` has zero callers — the plugin tracks connectivity via its own atomic fed by `on_state_change`. Same for `ble_uuids::kCommand` (line 26; only `kNotify`/`kConfig` are ever resolved) and `OgloDeviceConfig::sample_shape` (parsed but never read — `raw_json` already preserves it for provenance). Suggest deleting all three; note the `kCommand` deletion also strengthens the case that the schema-4 fallback path can never be triggered by this plugin.

docs/source/device/oglo.rst
jiwenc-nv: that being said, I don't think you need to hand craft the mcap recorder on your own in the plugin (i.e. mod 1).

You already implemented `examples/oglo_tactile/oglo_teleop_record.py` , so you already have a data recorder there.

but maybe you want the plugin to provide recording feature as well?  happy to discuss your needs as well.

docs/source/device/oglo.rst
```diff
@@ -0,0 +1,94 @@
...
+
+This plugin follows the :doc:`add_device` pattern (FlatBuffer schema → plugin
+``SchemaPusher`` → tracker → MCAP) and is modeled on the OAK camera plugin,
+including its two recording modes.
+
+.. contents:: On this page
+   :local:
+   :depth: 2
+
+Components
```
jiwenc-nv, line 27: **Please fix**: include a gif to show case the capabilities of your device and plugin (less than 5MB, ideally 10~15 seconds long).

src/core/schema/fbs/oglo_tactile.fbs
jiwenc-nv: would be great to add those info in the rst doc and include a link to that from here.

src/core/schema/fbs/oglo_tactile.fbs
```diff
@@ -0,0 +1,54 @@
...
+
+// One tactile + IMU sample from a single OGLO glove (one hand).
+//
+// OGLO is a 5-finger Velostat tactile glove: each finger has a 4x4 taxel matrix
+// (5 x 4 x 4 = 80 taxels). The hand side (left/right) is conveyed by the
+// channel / collection name, mirroring how hand poses use left/right channels.
+//
+// Source device: OGLO-MT-RDR-02 reader, firmware >= 0.7.0 (CONFIG_SCHEMA_VER 5),
+// BLE "packed12 v5" stream at 100 Hz (see the OGLO firmware packet-format spec).
+table OgloGloveSample {
```
jiwenc-nv, line 15: **heads up**: this schema is vendor specific, which is probably ok as you are the first to propose the schema.  We will probably publish a generic tactile schema down the road.  Will keep you in the loop and hope you can provide feedback and help adopt.

jiwenc-nv (changes requested):

src/core/mcap/cpp/inc/mcap/recording_traits.hpp
styu12: Added `ReplayOgloTactileTrackerImpl` and the factory entry, mirroring the pedal impl. record→replay works now.

docs/source/device/oglo.rst
styu12: Added a demo gif!

src/plugins/oglo_tactile/ble/oglo_ble_client.hpp
styu12: Removed all three, plus the BlueZ backend's now write-only `m_connected` that backed `is_connected()`

docs/source/device/oglo.rst
styu12: - fixed the channel names and added a small derivation table so the 5 spellings + mcap info line up.
- (as we discussed) Mode 1 was just a dev convenience, so I dropped it and kept the plugin push-only.
https://github.com/NVIDIA/IsaacTeleop/issues/805

examples/oglo_tactile/oglo_teleop_record.py
styu12: Done, made the hint CUDA-version-neutral!

src/core/schema/fbs/oglo_tactile.fbs
styu12: Added an "IMU frame" section + explicit gyro scale. The raw values are the sensor's own wrist-fixed axes; mapping them to the hand is a separate extrinsic (supplied with the hardware), so I kept the schema to data facts + scales rather than inlining X/Y/Z→hand directions.

src/core/schema/fbs/oglo_tactile.fbs
styu12: Makes sense. Happy to give feedback and help adopt the generic schema when it's ready :)

jiwenc-nv (approved):

Issue #696, opened 2026-06-23
Update all examples to use CloudXRLauncher


jiwenc-nv:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #700, opened 2026-06-24
Add CloudXRLauncher CLI helpers for in-process runtime launch

nv-jakob:
Expose argparse helpers so embedding apps and examples can default to launching the CloudXR runtime and WSS proxy, with --no-launch-cloudxr-runtime for environments where the runtime is already running.

## Description

Partial fix for #696

Fixes #(issue)

## Type of change

## Testing

Ran the added test

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a new command-line option to enable or disable launching the CloudXR runtime.
  * The runtime now supports a “no launch” mode for embedding or external-managed setups, while keeping the default behavior unchanged.

### DIFF
--- a/src/core/cloudxr/python/launcher.py
+++ b/src/core/cloudxr/python/launcher.py
@@ -10,8 +10,10 @@ Isaac Lab Teleop) without requiring a separate terminal.
 
 from __future__ import annotations
 
+import argparse
 import asyncio
 import atexit
+import contextlib
 import logging
 import os
 import signal
@@ -31,6 +33,8 @@ from .runtime import (
 )
 
 logger = logging.getLogger(__name__)
+
+DEFAULT_DEVICE_PROFILE = "Quest3"
 
 _RUNTIME_WORKER_CODE = """\
 import sys, os
@@ -72,6 +76,7 @@ class CloudXRLauncher:
         self,
         install_dir: str = "~/.cloudxr",
         env_config: str | Path | None = None,
+        device_profile: str = DEFAULT_DEVICE_PROFILE,
         accept_eula: bool = False,
         setup_oob: bool = False,
         usb_local: bool = False,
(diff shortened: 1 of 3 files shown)


src/core/cloudxr/python/launcher.py
```diff
@@ -166,6 +168,56 @@ def __init__(
...
+        (for example after sourcing ``~/.cloudxr/run/cloudxr.env``).
+        """
+        parser.add_argument(
+            "--launch-cloudxr-runtime",
+            action=argparse.BooleanOptionalAction,
+            default=True,
+            help=(
+                "Launch the CloudXR runtime and WSS proxy in-process before running "
+                "(default: true). Pass --no-launch-cloudxr-runtime when the runtime is "
+                "already running (e.g. after sourcing ~/.cloudxr/run/cloudxr.env)."
```
jiwenc-nv, line 211: since we are here, maybe add a arg for `--cloudxr-install-dir` as well (default to `~/.cloudxr`)

jiwenc-nv (approved):

src/core/cloudxr/python/launcher.py
nv-jakob: Added

nv-jakob: Hmm, is it actually install dir? We don't install anything there right? It's more like run dir right?

jiwenc-nv: right... the naming is probably not perfect.  I think we can update it.

aristarkhovNV (commented):
CLI helpers being optional ideally should not bloat the launcher api to keep it focused on its one core function. Keeping CLI helpers separate from the core launcher class would keep the door open for adding more such helpers in the future based on different arg parsers, env vars, etc.

patterm be like:
```python
ArgParseLaunchHelper.register(parser)
with ArgParseLaunchHelper(parser): # optionally creates the launcher within the helper context
    ...
```

src/core/cloudxr/python/launcher.py
nv-jakob: Can we do this in a follow up?

aristarkhovNV (approved):

PR #706, opened 2026-06-24
Add Quick Start runtime validation workflow

arussell-nvidia:
## Consolidation status
- The current implementation review path is #773, which consolidates this PR's Quick Start runtime validation baseline onto `main`.
- #773 also adds the hosted no-headset client/OOB streaming probe and an opt-in Isaac Lab validation hook.
- This PR remains open to preserve its review history and should not be merged independently while the consolidated path is under review.

## Scope retained here
- Add containerized WebXR browser smoke coverage to the existing CloudXR web-client build action.
- Build the web client, run Jest, and run a Playwright smoke check inside an isolated container without host networking or published host ports.
- Verify only the desktop browser/IWER path reaches the Connect-ready state.
- Add retry/backoff around wheel artifact URL resolution so transient post-upload lookup timeouts do not fail otherwise successful wheel publishing.

## Validation
- The current PR head is green/skipped in GitHub Actions.
- The consolidated #773 head is also green/skipped.

## Remaining evidence
Provisioned Isaac Lab execution evidence is still required on the consolidated path before it can claim the full opt-in Isaac Lab validation scope.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -519,9 +519,6 @@ jobs:
     # Runs the documented Quick Start workflow against the candidate wheel on a
     # GPU runner: install package, launch CloudXR, source the generated env, run
     # the documented gripper retargeting example, and upload the logs/summary.
-    # The desktop browser/IWER path remains the documented headset-free client
-    # path; this baseline validates the runnable guide steps without scraping
-    # the published documentation site.
     runs-on: [self-hosted, linux, gpu, x64]
     needs: build-ubuntu
     timeout-minutes: 30
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -556,7 +556,11 @@ jobs:
       uses: actions/upload-artifact@v6
       with:
         name: quick-start-e2e-x64-py3.11
-        path: ${{ runner.temp }}/isaacteleop-quick-start-e2e/
+        path: |
+          ${{ runner.temp }}/isaacteleop-quick-start-e2e/*.log
+          ${{ runner.temp }}/isaacteleop-quick-start-e2e/*.json
+          ${{ runner.temp }}/isaacteleop-quick-start-e2e/cloudxr/run/cloudxr.env
+          ${{ runner.temp }}/isaacteleop-quick-start-e2e/cloudxr/logs/**/*.log
         if-no-files-found: warn
         retention-days: 7
 
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -515,6 +515,51 @@ jobs:
         # Run tests with CloudXR runtime
         ./scripts/run_tests_with_cloudxr.sh --python-version ${{ matrix.python_version }}
 
+  test-quick-start-e2e:
+    # Runs the documented Quick Start workflow against the candidate wheel on a
+    # GPU runner: install package, launch CloudXR, source the generated env, run
+    # the documented gripper retargeting example, and upload the logs/summary.
+    # The desktop browser/IWER path remains the documented headset-free client
+    # path; this baseline validates the runnable guide steps without scraping
+    # the published documentation site.
+    runs-on: [self-hosted, linux, gpu, x64]
+    needs: build-ubuntu
+    timeout-minutes: 30
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v6
+      with:
+        submodules: recursive
+
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: '3.11'
+
+    - name: Download isaacteleop wheel (x64, py3.11)
+      uses: actions/download-artifact@v7
+      with:
+        name: isaacteleop-wheels-x64-py3.11
+        path: quick-start-wheels
+
+    - name: Run Quick Start E2E workflow
+      env:
+        CI: true
+        QUICK_START_E2E_REQUIRE_WHEEL: "1"
+        QUICK_START_E2E_WHEEL_DIR: quick-start-wheels
+        QUICK_START_E2E_ARTIFACT_DIR: ${{ runner.temp }}/isaacteleop-quick-start-e2e
+      run: ./scripts/run_quick_start_e2e.sh
+
+    - name: Upload Quick Start E2E artifacts
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v6
+      with:
+        name: quick-start-e2e-x64-py3.11
+        path: ${{ runner.temp }}/isaacteleop-quick-start-e2e/
+        if-no-files-found: warn
+        retention-days: 7
+
   test-teleop-ros2:
     runs-on: [self-hosted, linux, gpu, "${{ matrix.arch }}"]
     needs: build-ubuntu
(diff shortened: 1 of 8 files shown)


yanziz-nvidia:
Can we build and run tests in a container env so that you can truly test e2e without host network (and enable multiple instances on the same host?)

arussell-nvidia:
Thanks, good point. I am checking whether this e2e path can run in a containerized environment without relying on host networking, and what isolation is needed for multiple instances on one host.

arussell-nvidia:
Thanks, I pushed a follow-up commit for this. The web client action now runs the build, Jest tests, and Quick Start Playwright e2e test inside an isolated Docker container, with a per-run Docker network/container name and no host networking or published host ports. I also made the Playwright host/port/base URL configurable so the dev server can bind inside the container while the browser connects locally within that container.

Local verification before pushing:
- `npm test` passed.
- `USE_LOCAL_WEBXR_ASSETS=0 npm run build` passed with the existing bundle-size warnings.
- `npm run test:e2e` could not complete on this Windows workstation because Chromium launch is blocked with `browserType.launch: spawn EPERM`; the containerized path is set up to run in CI with the Playwright image.

arussell-nvidia:
Quick CI update: `Build Teleop Web App` is green on commit `1079d69c`, so the containerized web client build, Jest, and Quick Start Playwright e2e path passed in GitHub Actions. The broader platform matrix is still running.

yanziz-nvidia:
the test doesn't really test e2e?

if you look at quick start https://nvidia.github.io/IsaacTeleop/main/getting_started/quick_start.html it involves many components. We are not testing with just playwright loading the page. We need to actually connect to IT and send data and validate every steps in that quick start.

scripts/run_webxr_smoke_in_container.sh
```diff
@@ -0,0 +1,61 @@
...
+
+cleanup() {
+    docker rm -f "$CONTAINER_NAME" >/dev/null 2>&1 || true
+    docker network rm "$NETWORK_NAME" >/dev/null 2>&1 || true
+}
+trap cleanup EXIT
+
+docker network create "$NETWORK_NAME" >/dev/null
+
+docker run --rm \
```
yanziz-nvidia: see https://nvidia.github.io/IsaacTeleop/main/getting_started/quick_start.html what exactly test e2e of quick start means

yanziz-nvidia (changes requested):

arussell-nvidia:
Thanks, you're right. I pushed `a8aeeffb` to stop presenting this as a Quick Start e2e test.

This PR now scopes the added coverage to a containerized WebXR browser smoke check: it builds the web client, runs Jest, and uses Playwright only to verify that the desktop browser path loads IWER and reaches the Connect-ready state without host networking or published host ports.

A true Quick Start e2e should be a separate system test that starts the CloudXR runtime, connects the WebXR client to it, sends XR/controller input, and validates the teleop example output. I'll keep this PR limited to the smaller smoke coverage rather than claiming full Quick Start validation.

arussell-nvidia:
I updated this PR so it no longer presents the added coverage as a Quick Start E2E test. The current scope is a containerized WebXR browser smoke check: build the web client, run Jest, and use Playwright to verify the desktop browser path loads IWER and reaches the Connect-ready state without host networking or published host ports.

That addresses the review feedback by separating this smoke coverage from the real multi-component Quick Start E2E work, which should remain a separate system test/design. Current checks are green/skipped and the branch is mergeable. Ready for re-review.

.github/workflows/build-ubuntu.yml
```diff
@@ -515,6 +515,55 @@ jobs:
         # Run tests with CloudXR runtime
         ./scripts/run_tests_with_cloudxr.sh --python-version ${{ matrix.python_version }}
 
+  test-quick-start-e2e:
+    # Runs the documented Quick Start workflow against the candidate wheel on a
+    # GPU runner: install package, launch CloudXR, source the generated env, run
+    # the documented gripper retargeting example, and upload the logs/summary.
+    # The desktop browser/IWER path remains the documented headset-free client
+    # path; this baseline validates the runnable guide steps without scraping
```
qingsi-at-nv: nit: the doc scraping comment feels irrelevant.

src/core/cloudxr/python/launcher.py
```diff
@@ -240,11 +240,12 @@ def add_launch_cloudxr_runtime_argument(parser: argparse.ArgumentParser) -> None
         parser.add_argument(
             "--launch-cloudxr-runtime",
             action=argparse.BooleanOptionalAction,
-            default=True,
+            default=None,
```
qingsi-at-nv, line 243: If the default doesn't work for the workflow, can we just provide an explicit value instead?

src/core/cloudxr/python/oob_teleop_env.py
```diff
@@ -104,7 +104,8 @@ def default_web_client_origin() -> str:
 
 USB_HOST = "127.0.0.1"  # serverIP seen by the headset (its own localhost)
 USB_UI_DEFAULT_PORT = 8080  # HTTPS static WebXR UI (loopback)
-USB_BACKEND_DEFAULT_PORT = 49100  # CloudXR backend (webrtc client direct connection)
+CLOUDXR_SERVER_DEFAULT_PORT = 49100  # CloudXR runtime backend
+USB_BACKEND_DEFAULT_PORT = CLOUDXR_SERVER_DEFAULT_PORT  # adb reverse'd backend
```
qingsi-at-nv, line 108: This and code below introduce a new order of precedence order of env vars and also makes USB_BACKEND_DEFAULT_PORT unused except the test.

Earlier it was
USB_BACKEND_PORT > USB_BACKEND_DEFAULT_PORT

and now:
USB_BACKEND_PORT > NV_CXR_SERVER_PORT > CLOUDXR_SERVER_DEFAULT_PORT and USB_BACKEND_DEFAULT_PORT becomes unused.

Why do we need the extra cloudxr server port?

scripts/run_quick_start_e2e.sh
```diff
@@ -0,0 +1,336 @@
...
+        log "===== CloudXR server log ====="
+        cat "${CLOUDXR_LOG}" || true
+        fail "CloudXR WSS proxy did not report running"
+    }
+    assert_clean_log "CloudXR server log" "${CLOUDXR_LOG}"
+
+    log "Step 4: firewall allow-list is a machine configuration step; CI validates the local WSS proxy port instead"
+    check_wss_proxy_port || fail "WSS proxy port ${PROXY_PORT} is not reachable on localhost"
+
+    log "Step 5: desktop browser/IWER is the documented headset-free client path; this CI baseline does not inspect the website"
```
qingsi-at-nv, line 328: We need to connect from the emulated client.

scripts/run_quick_start_e2e.sh
```diff
@@ -0,0 +1,336 @@
...
+
+    run_example "${env_file}"
```
qingsi-at-nv, line 330: We need to not only run the application but also send input from the client and verify the output from the application side.

qingsi-at-nv:
Please also update the PR title to reflect the actual change.

arussell-nvidia:
Updated the PR title to `Add Quick Start E2E validation workflow`.

arussell-nvidia:
Thanks for the review. Agreed on the scope split.

I updated this PR to be the first Quick Start runtime validation baseline rather than the full client-interaction E2E test. The intent for this PR is to prove the automated install/launch/env/example path runs reliably in CI.

I'll handle the richer E2E validation as the next step: launch/connect through an emulated client, send input from that client, and verify the application-side output signal. I also pushed a small cleanup to keep the USB backend port behavior explicit and remove the stale comment wording.

PR #727, opened 2026-07-04
Add optional rate-limiting retargeters for bare position-servo followers (SO-101 class)

johnnynunez:
## What

Adds two optional retargeter nodes under `src/retargeters/` that govern commands before they reach a follower robot:

- **`JointRateLimiter`** — per-joint velocity clamp (with per-joint overrides, e.g. a faster gripper) against the **last emitted** command.
- **`EePoseRateLimiter`** — end-effector pose clamp: linear velocity [m/s] + geodesic angular velocity [rad/s] (double-cover safe).
- **`RateLimiterConfig`** — shared config (limits, `min_dt`/`max_dt` clamping, per-joint overrides, rejection thresholds).

Plus `src/core/retargeting_engine_tests/python/test_rate_limiter.py` (sim-free unit tests, same pattern as `test_so101_retargeters.py`).

Both nodes now implement a **three-band governor**:

1. **Pass-through** — motion under the velocity limits is emitted untouched (no lag, no filtering).
2. **Clamp** — motion over `max_*_velocity` is slewed toward the target at the limit.
3. **Reject** (new, optional) — a frame whose *input* velocity relative to the last **accepted input** exceeds `reject_linear_velocity` / `reject_angular_velocity` / `reject_joint_velocity` is a discontinuity (IK divergence, tracking teleport, stream catch-up step), not motion to follow, and is **not executed at all**: the limiter holds the last emitted command and does not creep toward the anomaly. For the joint limiter, a single joint over the threshold rejects the whole frame — joints move together, and executing the sane joints of an insane frame still produces an insane configuration.

Rejection anti-deadlock: after `max_consecutive_rejections` consecutive anomalous frames (default 30, ~1 s at 30 FPS) the input is treated as a persistent new target — a legitimate regime change, e.g. the leader really did move during a stream outage — and re-accepted, still clamped by band 2. `None` holds indefinitely until the input returns within the envelope or a reset arrives. Rejection is **disabled by default** (`reject_* = None`), so existing clamp-only behavior is unchanged; config validation enforces `reject_* >= max_*` (a rejection threshold below the clamp limit would refuse motion the clamp band is meant to allow).

## Why

Follower arms built on bare position servos (SO-101 Feetech bus and similar) execute whatever position reaches the bus at full torque — there is no controller-side velocity/accel envelope. Any anomalous retargeting frame becomes a full-torque slew:

- IK divergence near singularities / unreachable targets (observed: raw joint commands flipping ±100° at frame rate when the target left the workspace),
- XR tracking glitches (controller pose teleports),
- stream hiccups (a stale-then-recovered leader stream emits a large catch-up step).

This class of failure physically damaged a gear train on one of our SO-101 followers during XR teleop (context: [Slack thread — SO101 harness MR into Isaac Teleop, Lior Ben Horin]).

Clamping alone bounds the damage but still *walks the arm toward the anomaly* at the velocity limit — a ±100° IK flip at a 90 °/s clamp is 90 °/s of confident motion toward a garbage target for as long as the divergence lasts. The rejection tier closes that gap: a frame that asks for e.g. 3000 °/s is not evidence of intent, it is evidence of malfunction, and the correct amount of motion to execute toward it is zero.

## Design notes

- **Clamps against the last emitted command, not the last input** — a far-away target is approached at bounded velocity; it never jumps.
- **Rejects against the last accepted input, not the last emitted command** — the emitted pose intentionally lags a far target during catch-up; measuring input velocity against it would misread bounded catch-up as an anomaly.
- **`dt` is clamped to `[min_dt, max_dt]`** — a multi-second pipeline stall does not authorize a multi-second-sized step on the next frame.
- **Pass-through below the limits** — no added latency or filtering in normal operation.
- **A rejected frame does not advance the time baseline** (mirrors the dropped-frame path), so a glitch cannot inflate the next frame's authorized step.
- **Honest scope:** this bounds *velocity* only. It does not bound torque, acceleration, or collision forces — a collision at legal velocity still pushes at full servo torque. It is a safety net for anomalous frames, not a substitute for workspace/collision safety.

## Testing

- 104 tests pass (`test_rate_limiter.py` incl. 20 new rejection/validation tests + existing `test_so101_retargeters.py` + `test_joint_state_retargeter.py`) against the locally built wheel.
- Mutation-checked: disabling `_is_anomalous` in the EE limiter fails 5 tests; disabling the joint-frame rejection predicate fails 4 tests; removing the position clamp or the joint clip fails the corresponding clamp tests.
- Rejection scenario exercised end-to-end against the installed wheel: a 30 FPS joint stream at ~30 °/s with one injected +100°-in-one-frame IK flip (≈3000 °/s) — the anomalous frame is held (0° motion), and the recovered stream resumes pass-through with no jump. With rejection disabled the same flip is clamped (legacy behavior preserved).
- Previously exercised on real hardware (SO-101 follower + Quest 3 XR teleop at 30 FPS, and the LeRobot `isaac_teleop_to_so101` example wrapping `JointRateLimiter` between `device.compute()` and `robot.send_action()`): anomalous IK frames (±100° raw flips) were clamped to ≤3°/frame steps at 90 °/s; normal teleop is pass-through.
- `SKIP=check-copyright-year pre-commit` clean; DCO signed.

### DIFF
--- a/src/core/retargeting_engine_tests/python/test_rate_limiter.py
+++ b/src/core/retargeting_engine_tests/python/test_rate_limiter.py
@@ -1,0 +1,502 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Sim-free unit tests for the rate-limiting safety harness retargeters.
+
+Covers the safety-harness nodes that bound per-frame command velocity for bare
+position-servo followers (e.g. the SO-101):
+
+* :class:`~isaacteleop.retargeters.EePoseRateLimiter` -- linear/angular velocity
+  bounds on an absolute 7-D ``ee_pose`` stream.
+* :class:`~isaacteleop.retargeters.JointRateLimiter` -- per-joint velocity bounds
+  on a name-keyed ``joint_targets`` group.
+
+Each limiter is exercised at the pure-math level (the module-private clamp
+helpers) and at the ``BaseRetargeter.compute`` level (build inputs/outputs, drive
+frames with controlled ``GraphTime`` stamps, read the emitted tensors), with no
+``gym.make``, USD, GPU, or XR device.
+"""
+
+import math
+
+import numpy as np
+import pytest
+
+from isaacteleop.retargeting_engine.interface import (
+    ComputeContext,
+    ExecutionEvents,
+    ExecutionState,
+    OptionalTensorGroup,
+    TensorGroup,
+)
+from isaacteleop.retargeting_engine.interface.retargeter_core_types import GraphTime
+from isaacteleop.retargeting_engine.interface.tensor_group_type import (
+    OptionalTensorGroupType,
+)
+from isaacteleop.retargeters import (
+    EePoseRateLimiter,
+    JointRateLimiter,
+    RateLimiterConfig,
+)
+from isaacteleop.retargeters.rate_limiter import (
+    EE_POSE_KEY,
+    JOINT_TARGETS_KEY,
+    _clamp_orientation_step,
+    _clamp_position_step,
+    _clamped_dt,
+)
+
+# ---------------------------------------------------------------------------
+# Helpers (mirror the patterns in test_so101_retargeters.py)
+# ---------------------------------------------------------------------------
+
+_ID_QUAT = np.array([0.0, 0.0, 0.0, 1.0], dtype=np.float64)
+_NS = 1_000_000_000
+
+
+def _make_context(
+    *,
+    reset: bool = False,
+    state: ExecutionState = ExecutionState.RUNNING,
+    t_ns: int = 0,
+) -> ComputeContext:
+    """Build a ComputeContext with the given reset flag, state, and timestamp."""
+    return ComputeContext(
+        graph_time=GraphTime(sim_time_ns=t_ns, real_time_ns=t_ns),
+        execution_events=ExecutionEvents(reset=reset, execution_state=state),
+    )
+
+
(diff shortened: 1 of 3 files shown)


jiwenc-nv:
Heads-up on something we hit running these on a headset: `EePoseRateLimiter` emits a non-finite pose after about a second of ordinary teleop.

`_clamp_orientation_step`'s pass-through branch returns `prev ⊗ (prev⁻¹ ⊗ target)` — target, re-derived through `prev` to keep the hemisphere. Its norm is `|prev|²`, and the result becomes the next frame's `prev`, so the deviation from unit **squares every frame**. From float64's 1e-16 that reaches float32's ceiling in ~60 frames. Driving the node directly with a moving target it goes non-finite at frame 62 — 0.86 s at 72 Hz. `astype(np.float32)` then saturates and every downstream quaternion product turns to NaN; on our end the rendered gripper left the room.

Only pass-through diverges — the clamp branch multiplies by a unit step and holds its norm — so it's the band a well-tuned harness spends its whole session in. That's also why the tests here don't catch it: they're a handful of frames and mostly exercise clamping. A *fixed* target stays exactly unit too, because the rounding cancels; it needs a moving one to show.

One renormalization before the value is fed back fixes it, with a regression test: https://github.com/NVIDIA/IsaacTeleop/commit/93a98ed5e7dfa2028bee64ee67ef515afcca5a8e. After it, 20000 frames of jittered pass-through hold `||q|-1|` at 2.2e-16.

We're carrying it in #928, which cherry-picks these nodes — but it reaches any follower driven through the EE limiter, so it belongs here rather than downstream.

jiwenc-nv:
merged this as part of https://github.com/NVIDIA/IsaacTeleop/pull/981

PR #728, opened 2026-07-05
fix(cloudxr): LD_PRELOAD bundled OpenSSL into the runtime worker

johnnynunez:
Fix ARM cloudXR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved launcher startup reliability by ensuring the bundled security libraries are used when available.
  * Resolved a crash scenario that could occur during runtime initialization on affected systems.


nv-mhaselton (approved):
Thanks for providing a patch!

PR #753, opened 2026-07-07
fix(webxr): ship safe Quest profile defaults (72 FPS, 25 Mbps)

asierarranz:
# Description

Out of the box, a Quest connecting to the web client gets the 90 FPS / 150 Mbps profile, which on most Wi-Fi setups means reprojection judder and freezes within a few minutes. This ships Quest defaults that just work.

What changed:

- `DeviceProfiles.ts`: the Quest profile now defaults to 72 FPS and 25 Mbps. 4K per eye at 90 Hz over 150 Mbps saturates a typical 5 GHz link; 72/25 held stable over long real sessions in our testing.
- `index.html`: the bitrate select gains 25 and 50 Mbps options so the new defaults are actually selectable in the UI.
- `CloudXR2DUI.tsx`: on initialize, a persisted non-custom profile is re-applied from the current profile table, so users who stored the old defaults in localStorage pick up the safe ones without clearing storage.
- `DeviceProfiles.test.ts`: new unit test asserting no device profile ships a bitrate above 100 Mbps.
- `package.json`: the jest roots now include `helpers/` so the new test runs with `npm test`.

How tested: jest unit tests, plus real Quest 3 sessions over CloudXR with the new defaults (72 FPS negotiated end to end, no freezes at 25 Mbps where 150 Mbps froze within minutes).

## Summary by CodeRabbit

* **New Features**
  * Added automated coverage to verify streaming bitrate settings stay within supported limits.

* **Bug Fixes**
  * Updated default device streaming settings to use 72 FPS and 25 Mbps for more stable behavior.
  * Restored saved device profile selections more reliably after initialization, helping prevent outdated settings from sticking.
  * Updated the UI defaults so the frame rate and streaming bitrate controls now start at the new recommended values.

### DIFF
--- a/deps/cloudxr/webxr_client/src/index.html
+++ b/deps/cloudxr/webxr_client/src/index.html
@@ -905,8 +905,8 @@ limitations under the License.
                             <label class="input-label">Frame Rate Configuration</label>
                             <label for="deviceFrameRate" class="input-label">Device Frame Rate</label>
                             <select id="deviceFrameRate" class="ui-input config-input">
-                                <option value="72">72 FPS</option>
-                                <option value="90" selected>90 FPS</option>
+                                <option value="72" selected>72 FPS</option>
+                                <option value="90">90 FPS</option>
                                 <option value="120">120 FPS</option>
                             </select>
                             <div class="config-text">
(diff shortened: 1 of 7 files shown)


deps/cloudxr/webxr_client/src/index.html
```diff
@@ -905,8 +905,8 @@ <h4 class="advanced-divider">Advanced settings</h4>
                             <label class="input-label">Frame Rate Configuration</label>
                             <label for="deviceFrameRate" class="input-label">Device Frame Rate</label>
                             <select id="deviceFrameRate" class="ui-input config-input">
-                                <option value="72">72 FPS</option>
-                                <option value="90" selected>90 FPS</option>
+                                <option value="72" selected>72 FPS</option>
```
yanziz-nvidia, line 908: This is probably ok here for now. We should instead detect supported device frame rate and make the list dynamic

yanziz-nvidia: could be a separate PR

yanziz-nvidia (approved):
It changes the default behavior. Please add to documentation.

asierarranz:
Thanks for the review! Docs updated in 8ceb91e: the new defaults are now in the quick start (Quest connect section), plus a note on how saved profiles pick up the new values on load.

Agreed on the dynamic frame rate list, will send that as a separate PR.

PR #755, opened 2026-07-08
ROS: Standardize hand and EE pose topics with NamedPoseArray

### LINKED ISSUE #645: Standardize interfaces for EE poses and hand commands in ROS or introduce custom messages
Currently EE poses and hand commands in ROS are published using Pose Array that is hard for subscribers to handle - it provides no frame ids for poses. Either standardize a new ROS message that is a Pose Array with frame ids, or use a custom message with frame ids when publishing hand commands

sgrizan-nv:
## Description

Fixes #645

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Testing

Manual & integration testing

## Checklist

- [X] I have read and understood the [contribution guidelines](../CONTRIBUTING.md)
- [X] I have run the linter and formatter with `SKIP=check-copyright-year pre-commit run --all-files`
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix/feature works (or explained why not)
- [X] I have signed off all my commits (`git commit -s`) per the [DCO](../CONTRIBUTING.md#signing-your-work)

## Summary by CodeRabbit

- **New Features**
  - Added `NamedPoseArray` message support (names alongside poses and per-entry validity).
  - Unified end-effector publishing on `xr_teleop/ee_pose` using `NamedPoseArray`.
  - Updated hand publishing to named left/right joint poses on `xr_teleop/hand` using `NamedPoseArray`.
  - Added the `teleop_ros2_interfaces` ROS 2 package and installation of its interfaces for `examples/teleop_ros2`.

- **Bug Fixes**
  - Strengthened runtime topic verification (frame, name ordering/uniqueness, array consistency, and validity/non-zero pose expectations).

- **Documentation**
  - Updated README topics, remapping examples, and echo guidance to match the new message format and topic names.

### DIFF
--- a/examples/teleop_ros2/Dockerfile
+++ b/examples/teleop_ros2/Dockerfile
@@ -114,6 +114,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     git \
     patchelf \
     pkg-config \
+    python3-colcon-common-extensions \
+    ros-${ROS_DISTRO}-ament-cmake-auto \
     ros-${ROS_DISTRO}-rosidl-default-generators \
     && rm -rf /var/lib/apt/lists/*
 
@@ -146,17 +148,18 @@ RUN --mount=type=cache,target=/ccache,id=isaacteleop-teleop-ros2-ccache-${ROS_DI
     && cmake --build build --config Release -j$(nproc) \
     && cmake --install build --config Release \
     && bash -lc "source /opt/ros/${ROS_DISTRO}/setup.bash \
-        && cmake -B build/teleop_ros2_interfaces -S examples/teleop_ros2/teleop_ros2_interfaces \
-            -DCMAKE_BUILD_TYPE=Release \
-            -DCMAKE_INSTALL_PREFIX=/opt/isaacteleop/install \
-        && cmake --build build/teleop_ros2_interfaces --config Release -j$(nproc) \
-        && cmake --install build/teleop_ros2_interfaces --config Release" \
+        && colcon build \
+            --base-paths examples/teleop_ros2/teleop_ros2_interfaces \
+            --build-base build/teleop_ros2_interfaces \
+            --install-base /opt/isaacteleop/install \
+            --merge-install \
+            --cmake-args -DCMAKE_BUILD_TYPE=Release" \
     && ccache --show-stats
 
 COPY --chmod=644 <<EOF /usr/local/bin/teleop-env-setup
 source /opt/ros/${ROS_DISTRO}/setup.bash
-if [[ -f /opt/isaacteleop/install/share/teleop_ros2_interfaces/local_setup.bash ]]; then
-    source /opt/isaacteleop/install/share/teleop_ros2_interfaces/local_setup.bash
+if [[ -f /opt/isaacteleop/install/local_setup.bash ]]; then
+    source /opt/isaacteleop/install/local_setup.bash
 fi
 EOF
 
--- a/examples/teleop_ros2/Dockerfile
+++ b/examples/teleop_ros2/Dockerfile
@@ -128,7 +128,8 @@ COPY . /opt/isaacteleop
 # produces; a stale build/ would skip a clean CMake reconfigure. The build tree
 # is rebuilt from scratch every time -- ccache keeps recompiles cheap -- so each
 # build reconfigures against the current image rather than trusting cached
-# configure results.
+# configure results. Ament's Python tools come from apt, so the interface
+# package build below pins the ROS distro's system Python instead of uv's.
 RUN --mount=type=cache,target=/ccache,id=isaacteleop-teleop-ros2-ccache-${ROS_DISTRO}-${TARGETARCH} \
     rm -rf /opt/isaacteleop/install /opt/isaacteleop/build \
     && cmake -B build -S . \
@@ -153,7 +154,9 @@ RUN --mount=type=cache,target=/ccache,id=isaacteleop-teleop-ros2-ccache-${ROS_DI
             --build-base build/teleop_ros2_interfaces \
             --install-base /opt/isaacteleop/install \
             --merge-install \
-            --cmake-args -DCMAKE_BUILD_TYPE=Release" \
+            --cmake-args \
+                -DCMAKE_BUILD_TYPE=Release \
+                -DPython3_EXECUTABLE=/usr/bin/python3" \
     && ccache --show-stats
 
 COPY --chmod=644 <<EOF /usr/local/bin/teleop-env-setup
--- a/examples/teleop_ros2/Dockerfile
+++ b/examples/teleop_ros2/Dockerfile
@@ -46,6 +46,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libxxf86vm-dev \
     swig \
     ros-${ROS_DISTRO}-rmw-cyclonedds-cpp \
+    ros-${ROS_DISTRO}-rosidl-default-runtime \
     && rm -rf /var/lib/apt/lists/*
 
 # Install uv via standalone script (avoids PEP 668 "externally managed" pip on Ubuntu 24.04 / Jazzy)
@@ -113,6 +114,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     git \
     patchelf \
     pkg-config \
+    ros-${ROS_DISTRO}-rosidl-default-generators \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /opt/isaacteleop
@@ -143,10 +145,19 @@ RUN --mount=type=cache,target=/ccache,id=isaacteleop-teleop-ros2-ccache-${ROS_DI
         -DISAAC_TELEOP_PYTHON_VERSION=${PYTHON_VERSION} \
     && cmake --build build --config Release -j$(nproc) \
     && cmake --install build --config Release \
+    && bash -lc "source /opt/ros/${ROS_DISTRO}/setup.bash \
+        && cmake -B build/teleop_ros2_interfaces -S examples/teleop_ros2/teleop_ros2_interfaces \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX=/opt/isaacteleop/install \
+        && cmake --build build/teleop_ros2_interfaces --config Release -j$(nproc) \
+        && cmake --install build/teleop_ros2_interfaces --config Release" \
     && ccache --show-stats
 
 COPY --chmod=644 <<EOF /usr/local/bin/teleop-env-setup
 source /opt/ros/${ROS_DISTRO}/setup.bash
+if [[ -f /opt/isaacteleop/install/share/teleop_ros2_interfaces/local_setup.bash ]]; then
+    source /opt/isaacteleop/install/share/teleop_ros2_interfaces/local_setup.bash
+fi
 EOF
 
 COPY --chmod=755 <<'EOF' /usr/local/bin/teleop-entrypoint
--- a/examples/teleop_ros2/python/messages.py
+++ b/examples/teleop_ros2/python/messages.py
@@ -11,6 +11,7 @@ import numpy as np
 from geometry_msgs.msg import Pose, PoseArray, PoseStamped
 from scipy.spatial.transform import Rotation
 from sensor_msgs.msg import JointState
+from teleop_ros2_interfaces.msg import HandJointPose, HandJointPoses
 
 from isaacteleop.retargeting_engine.interface import OptionalTensorGroup
 from isaacteleop.retargeting_engine.tensor_types.indices import (
@@ -21,9 +22,8 @@ from isaacteleop.retargeting_engine.tensor_types.indices import (
     HeadPoseIndex,
 )
 
-from constants import BODY_JOINT_NAMES
+from constants import BODY_JOINT_NAMES, HAND_POSE_JOINT_INDICES, HAND_POSE_NAMES
 from geometry import (
-    append_hand_poses,
     apply_manus_controller_to_hand_pose,
     apply_transform_to_pose,
     to_pose,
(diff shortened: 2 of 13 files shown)


sgrizan-nv:
@coderabbitai help

sgrizan-nv:
@coderabbitai full review

examples/teleop_ros2/teleop_ros2_interfaces/msg/NamedPoseArray.msg
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```
kajananchinniahNV, line 1: Nit: we should consider renaming this

Some options are:
- PoseStampedArray  (although this might not be exactly right)
- LinkState (although not confident in this one but this message is analogous to: https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/JointState.html)

Not sure what else, but we should look at what other ROS Msgs are available and trying to have a similar name to that basically

examples/teleop_ros2/teleop_ros2_interfaces/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+<package format="3">
+  <name>teleop_ros2_interfaces</name>
+  <version>0.1.0</version>
+  <description>Custom ROS 2 interfaces for the Isaac Teleop ROS 2 example.</description>
+  <maintainer email="isaacteleop-maintainers@nvidia.com">Isaac Teleop Maintainers</maintainer>
```
kajananchinniahNV: Does the email actually work?

examples/teleop_ros2/teleop_ros2_interfaces/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+<package format="3">
+  <name>teleop_ros2_interfaces</name>
+  <version>0.1.0</version>
+  <description>Custom ROS 2 interfaces for the Isaac Teleop ROS 2 example.</description>
```
kajananchinniahNV: Maybe call this Custom ROS 2 Interfaces for the Isaac Teleop or something instead

examples/teleop_ros2/teleop_ros2_interfaces/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.8)
+project(teleop_ros2_interfaces)
+
+find_package(ament_cmake REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
+find_package(std_msgs REQUIRED)
```
kajananchinniahNV: Using ament_auto might make this simpler - we use a similar pattern here for example: https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_teleop/blob/main/isaac_ros_teleop/CMakeLists.txt#L9

examples/teleop_ros2/teleop_ros2_interfaces/msg/NamedPoseArray.msg
sgrizan-nv: Yeah, I'm not happy with the current name either, but it doesn't feel like either PoseStampedArray or LinkState describe this message better

examples/teleop_ros2/Dockerfile
```diff
@@ -143,10 +145,19 @@ RUN --mount=type=cache,target=/ccache,id=isaacteleop-teleop-ros2-ccache-${ROS_DI
         -DISAAC_TELEOP_PYTHON_VERSION=${PYTHON_VERSION} \
     && cmake --build build --config Release -j$(nproc) \
     && cmake --install build --config Release \
+    && bash -lc "source /opt/ros/${ROS_DISTRO}/setup.bash \
+        && cmake -B build/teleop_ros2_interfaces -S examples/teleop_ros2/teleop_ros2_interfaces \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX=/opt/isaacteleop/install \
+        && cmake --build build/teleop_ros2_interfaces --config Release -j$(nproc) \
+        && cmake --install build/teleop_ros2_interfaces --config Release" \
```
kajananchinniahNV: Should we use colcon here instead?

examples/teleop_ros2/teleop_ros2_interfaces/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+<package format="3">
+  <name>teleop_ros2_interfaces</name>
```
kajananchinniahNV, line 7: Would we want to rename this package into something else? The current name is fine, but another name could be:
- isaac_teleop_ros2_interfaces

examples/teleop_ros2/python/messages.py
```diff
@@ -36,6 +36,70 @@
 )
 
 
+def _compose_ee_msg(
+    left_pose: Pose | None,
+    right_pose: Pose | None,
+    now,
+    frame_id: str,
```
kajananchinniahNV, line 43: Nit: could we make the interface of this into:
```
def _compose_ee_msg(pose, name, frame_id) 
```

instead and then just call it twice:
_compose_ee_msg(left_pose, "left", now, frame_id)
_compose_ee_msg(right_pose, "right", now, frame_id)

examples/teleop_ros2/python/messages.py
```diff
@@ -119,50 +183,53 @@ def _as_float(ctrl, index):
     }
 
 
-def build_ee_msg_from_controller(
-    ctrl: OptionalTensorGroup,
-    side: str,
+def build_ee_msg_from_controllers(
+    left_ctrl: OptionalTensorGroup,
+    right_ctrl: OptionalTensorGroup,
```
kajananchinniahNV, line 188: Could we keep the call here the same as before and just call it twice?

examples/teleop_ros2/python/messages.py
```diff
@@ -36,6 +36,70 @@
...
+            pose_is_valid = False
+        else:
+            pose_is_valid = True
+        msg.name.append(side)
+        msg.pose.append(pose)
+        msg.is_valid.append(pose_is_valid)
+    return msg
+
+
+def _compute_ee_pose_from_controller(
```
kajananchinniahNV, line 61: Could we merge _compute_ee_pose_from_controller and _compute_ee_pose_from_hand? Their core logic seems to be pretty similar looking back (the main difference being if we need manus or not)

examples/teleop_ros2/python/messages.py
sgrizan-nv: build_ee_msg_from_controllers and build_ee_msg_from_hands package left and right poses into the same NamedPoseArray, just as it was with PoseArray. Do you suggest splitting both EEs and hand poses into two topics (left and right)?

examples/teleop_ros2/python/messages.py
```diff
@@ -119,50 +183,53 @@ def _as_float(ctrl, index):
...
+        transform_rot,
+        transform_trans,
+        controller_uses_hands_source,
+    )
+    return _compose_ee_msg(left_pose, right_pose, now, frame_id)
 
 
-def build_ee_msg_from_hand(
-    hand: OptionalTensorGroup,
+def build_ee_msg_from_hands(
```
kajananchinniahNV, line 213: Similarly, do we have any opportunties of reusing code b/w this and the controller path?

examples/teleop_ros2/python/messages.py
sgrizan-nv: _compose_ee_msg produces one NamedPoseArray. Calling it twice will produce two NamedPoseArrays

examples/teleop_ros2/teleop_ros2_interfaces/package.xml
sgrizan-nv: Good catch, we don't have an alias. Changed it to my email address

examples/teleop_ros2/teleop_ros2_interfaces/package.xml
sgrizan-nv: Done

examples/teleop_ros2/teleop_ros2_interfaces/CMakeLists.txt
sgrizan-nv: Done

examples/teleop_ros2/Dockerfile
sgrizan-nv: Done

sgrizan-nv:
@coderabbitai review

examples/teleop_ros2/python/messages.py
kajananchinniahNV: I think keeping them in the same namedposearray makes sense; although we can have the hand poses in a separate topic (similar to how it was before)

The proposal is just changing it in code so it looks structually like:
```
named_pose_msg = NamedPoseMessage()
left_pose = build_left_msg()
right_pose = build_right_msg()
named_pose_msg.append([left_pose, right_pose])
```

Admittedly since we are doing the message as an structure of arrays it might not fit as nicely but I'd prefer trying to make things generic until we can't - since I feel like it'd support non-bimanual cases in the future easier

examples/teleop_ros2/python/messages.py
kajananchinniahNV: Clarified what I meant here: https://github.com/NVIDIA/IsaacTeleop/pull/755#discussion_r3575095256

We can discuss it offline if it's still not clear; it's mostly just an idea for restructuring the code not changing the final output

examples/teleop_ros2/python/messages.py
sgrizan-nv: They have similarities, but they differ in validation, tensor layout, and controller-only side correction. A single function would need source flags or callbacks, reducing readability. How do you propose to cleanly merge them?

examples/teleop_ros2/python/messages.py
sgrizan-nv: What code reuse do you have in mind? Their common responsibility is already factored into _compose_ee_msg

examples/teleop_ros2/python/messages.py
sgrizan-nv: The current approach is to isolate all message building logic in messages.py, so that downstream functions just send it over without modifying. Any common logic between different builders is extracted into private functions within messages.py

examples/teleop_ros2/python/messages.py
kajananchinniahNV: Fair, I think it might reduce readability unless we extracted the specific things earlier but that would also be messy 

We can likely resolve this. I don't have an idea currently for merging it unless we had some sort of class served as an abstraction b/w the two modes, and extracted whatever is common + some metadata earlier, then we did ducktyping for class specific things

examples/teleop_ros2/python/messages.py
kajananchinniahNV: Responded in the other thread - I think we can resolve this. My idea isn't too solid currently

examples/teleop_ros2/python/messages.py
kajananchinniahNV: Would accepting an array instead make sense here or do you think we should just resolve this for now until we explicitly want to support non-bimanual cases? That case is the main reason I'm worried 

I think we can resolve this for now however

examples/teleop_ros2/teleop_ros2_interfaces/msg/NamedPoseArray.msg
kajananchinniahNV: I'm curious what Lionel thinks a better name is since I agree that those two message names aren't ideal either

kajananchinniahNV (approved):
LGTM beyond the minor nitpicks!

examples/teleop_ros2/python/messages.py
sgrizan-nv: I'm also not 100% happy with the current structure, so let me know if you have ideas of how to make it more structured and reduce code repetition, without sacrificing readability. Resolving this for now

examples/teleop_ros2/teleop_ros2_interfaces/msg/NamedPoseArray.msg
lgulich: All nit, but I don't like PoseStampedArray since it suggests that the message contains PoseStamped elements. And `PoseArrayStamped` is also not a good option since this suggests a message with a Pose array and a header, which is exactly what PoseArray already is.

Personally, I like LinkState since it's analogous to JointState (but I also see that it's confusing since Link suggests a real physical link whereas this message could also represent virtual frames like the ee target).

Actually while writing this I realized that we may have a third option where we can use an existing message: [TFMessage](https://docs.ros2.org/foxy/api/tf2_msgs/msg/TFMessage.html). This message allows to send a list of transforms with parent and child frame name. It does not have an "is_valid" field but we could send this by sending NaN values for the transform values. Would this actually be a better solution that would not require any custom messages? Sorry for only coming up with this so late

sgrizan-nv: Discussed with @kajananchinniahNV offline that we will land the current version and iterate on it later if needed

PR #774, opened 2026-07-15
ROS: Centralize teleop runtime profiles

sgrizan-nv:
## Description

Make mode-specific session and publication contracts explicit, and add focused unit coverage to prevent wiring drift.

## Summary by CodeRabbit

* **New Features**
  * Added profile-based teleoperation handling to select required outputs and published telemetry for each operating mode.
  * Added clearer validation for session results, including actionable missing or unexpected output errors.
  * Standardized teleoperation mode handling with named mode values.

* **Bug Fixes**
  * Improved consistency of hand, controller, full-body, head, and root-command output publishing across modes.

* **Tests**
  * Added coverage for geometry transformations, message construction, profile selection, mode validation, and malformed session results.

### DIFF
--- a/examples/teleop_ros2/python/constants.py
+++ b/examples/teleop_ros2/python/constants.py
@@ -27,6 +27,13 @@ class HandRetargeter(StrEnum):
     DEXPILOT = "dexpilot"
 
 
+class TeleopMode(StrEnum):
+    CONTROLLER_TELEOP = "controller_teleop"
+    HAND_TELEOP = "hand_teleop"
+    CONTROLLER_RAW = "controller_raw"
+    FULL_BODY = "full_body"
+
+
 BODY_JOINT_NAMES = [e.name for e in BodyJointPicoIndex]
 HAND_POSE_JOINT_INDICES = tuple(
     HandJointIndex(i)
@@ -35,7 +42,7 @@ HAND_POSE_JOINT_INDICES = tuple(
 HAND_POSE_NAMES = [joint.name for joint in HAND_POSE_JOINT_INDICES]
 HAND_RETARGETERS = tuple(retargeter.value for retargeter in HandRetargeter)
 SHARPA_HAND_RETARGETERS = (HandRetargeter.PINK_IK, HandRetargeter.DEXPILOT)
-TELEOP_MODES = ("controller_teleop", "hand_teleop", "controller_raw", "full_body")
+TELEOP_MODES = tuple(mode.value for mode in TeleopMode)
 
 TRIHAND_JOINT_NAMES = [
     "thumb_rotation",
@@ -81,16 +88,16 @@ DEX_HANDTRACKING_TO_BASELINK_FRAME_TRANSFORM = (0, -1, 0, -1, 0, 0, 0, 0, -1)
 
 
 def resolve_hand_retargeter(
-    mode: str, hand_retargeter: HandRetargeter
+    mode: TeleopMode, hand_retargeter: HandRetargeter
 ) -> HandRetargeter:
     if hand_retargeter == HandRetargeter.MODE_DEFAULT:
-        if mode == "controller_teleop":
+        if mode == TeleopMode.CONTROLLER_TELEOP:
             return HandRetargeter.TRIHAND
-        if mode == "hand_teleop":
+        if mode == TeleopMode.HAND_TELEOP:
             return HandRetargeter.DEXPILOT
         return hand_retargeter
 
-    if mode == "hand_teleop" and hand_retargeter == HandRetargeter.TRIHAND:
+    if mode == TeleopMode.HAND_TELEOP and hand_retargeter == HandRetargeter.TRIHAND:
         raise ValueError(
             "Parameter 'hand_retargeter:=trihand' is only valid with "
             "mode:=controller_teleop"
@@ -100,6 +107,9 @@ def resolve_hand_retargeter(
 
 
 def uses_hands_source_for_controller(
-    mode: str, hand_retargeter: HandRetargeter
+    mode: TeleopMode, hand_retargeter: HandRetargeter
 ) -> bool:
-    return mode == "controller_teleop" and hand_retargeter in SHARPA_HAND_RETARGETERS
+    return (
+        mode == TeleopMode.CONTROLLER_TELEOP
+        and hand_retargeter in SHARPA_HAND_RETARGETERS
+    )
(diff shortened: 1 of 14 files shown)


jiwenc-nv (approved):

examples/teleop_ros2/python/geometry.py
```diff
@@ -10,26 +10,6 @@
 from geometry_msgs.msg import Pose, TransformStamped
 from scipy.spatial.transform import Rotation
 
-from isaacteleop.retargeting_engine.tensor_types.indices import HandJointIndex
-
-
-def append_hand_poses(
```
kajananchinniahNV, line 16: Was this deletion intentional?

examples/teleop_ros2/python/teleop_profiles.py
```diff
@@ -0,0 +1,178 @@
...
+
+from isaacteleop.retargeting_engine.interface import OptionalTensorGroup
+
+from constants import SHARPA_HAND_RETARGETERS, HandRetargeter, StrEnum, TeleopMode
+
+
+class FramePublication(StrEnum):
+    CONTROLLER_PAYLOAD = "controller_payload"
+    EE_FROM_CONTROLLERS = "ee_from_controllers"
+    EE_FROM_HANDS = "ee_from_hands"
```
kajananchinniahNV, line 18: Do we need to separate these two cases?

examples/teleop_ros2/python/teleop_profiles.py
```diff
@@ -0,0 +1,178 @@
...
+    FINGER_JOINTS = "finger_joints"
+    FULL_BODY_PAYLOAD = "full_body_payload"
+    HAND_POSES = "hand_poses"
+    HEAD = "head"
+    ROOT_COMMAND = "root_command"
+
+
+class TeleopProfile(StrEnum):
+    CONTROLLER_TELEOP = "controller_teleop"
+    CONTROLLER_TELEOP_WITH_HANDS = "controller_teleop_with_hands"
```
kajananchinniahNV, line 28: I also wonder if we need to separate these two cases too

examples/teleop_ros2/python/constants.py
```diff
@@ -27,6 +27,13 @@ class HandRetargeter(StrEnum):
     DEXPILOT = "dexpilot"
 
 
+class TeleopMode(StrEnum):
```
kajananchinniahNV, line 30: Do we need to worry about compatibility with older python versions?

examples/teleop_ros2/python/teleop_profiles.py
```diff
@@ -0,0 +1,178 @@
...
+
+from dataclasses import dataclass
+from typing import Mapping, TypedDict, cast
+
+from isaacteleop.retargeting_engine.interface import OptionalTensorGroup
+
+from constants import SHARPA_HAND_RETARGETERS, HandRetargeter, StrEnum, TeleopMode
+
+
+class FramePublication(StrEnum):
```
kajananchinniahNV: Is FramePublication the best name for this too? Would message publication be a better name?

kajananchinniahNV (approved):

examples/teleop_ros2/python/geometry.py
sgrizan-nv: Yes, the function is not used anywhere

examples/teleop_ros2/python/teleop_profiles.py
sgrizan-nv: Both publish xr_teleop/ee_pose, but they use different session inputs and builders: controller aim poses versus tracked hand wrist poses. Keeping separate publication actions lets the resolved profile select the correct path without reintroducing a mode/source conditional in the node. We could combine them only by adding a separate ee_pose_source field, which would add more complexity.

examples/teleop_ros2/python/teleop_profiles.py
sgrizan-nv: They share the user-facing controller_teleop mode, but resolve to different runtime contracts. The Sharpa path adds hand_left/hand_right session outputs and publishes HAND_POSES; the TriHand path does neither. Separate profiles keep each contract complete and immutable instead of adding conditional fields or capability booleans to one profile.

examples/teleop_ros2/python/constants.py
sgrizan-nv: constants.py already provides a enum.StrEnum fallback when the import is unavailable

examples/teleop_ros2/python/teleop_profiles.py
kajananchinniahNV: Got it, I was thinking we keep the conditional from before. I don't mind this instead, resolving it

examples/teleop_ros2/python/teleop_profiles.py
kajananchinniahNV: Got it, I was also thinking keeping the conditional here was better since we have two variables to track this, but I guess making it into one would be easier

Another option could be to have each mode have it's own class with a set of functions etc that it performs (so the controller teleop mode could take and store both variables).

Roughly it'd be:
```
class ControllerTeleopClass:
   def __init__(self, use_hands):
        self.use_hands = use_hands

    def example_function():
        if self._use_hands:
               return HAND_POSES
        else:
              return CONTROL_POSES
```

examples/teleop_ros2/python/teleop_profiles.py
sgrizan-nv: What do you think about PublishAction?

examples/teleop_ros2/python/constants.py
kajananchinniahNV: Ack

examples/teleop_ros2/python/teleop_profiles.py
sgrizan-nv: A per-mode strategy class would be a reasonable next step if modes acquire substantial state or mode-specific execution logic. Right now they only select static required-result keys and publish actions, while the builders/publishers remain shared

examples/teleop_ros2/python/teleop_profiles.py
kajananchinniahNV: Action is a specific thing within ROS so against that name - what about PublishType?

examples/teleop_ros2/python/teleop_profiles.py
kajananchinniahNV: Ack - resolving this. Let's keep it as is then unless things get more complicated

examples/teleop_ros2/python/teleop_profiles.py
sgrizan-nv: Great idea, done

PR #779, opened 2026-07-15
ROS: Extract teleop ROS publication builders

sgrizan-nv:
## Description

Move root-command, msgpack, and TF construction out of TeleopRos2Node so publisher methods only perform ROS side effects.

## Summary by CodeRabbit

* **New Features**
  * Improved ROS 2 teleoperation publishing for controller, full-body, head, hand, and root-command data.
  * Added coordinated publication of pose messages and related transforms.
  * Added support for encoded controller and full-body message payloads.
  * Root commands now publish both movement and height information.

* **Tests**
  * Expanded coverage for message formats, transforms, absent inputs, and combined hand/controller outputs.

* **Documentation**
  * Added guidance for structuring ROS publisher methods and output mapping logic.

### DIFF
--- a/examples/teleop_ros2/python/messages.py
+++ b/examples/teleop_ros2/python/messages.py
@@ -2,15 +2,18 @@
 # All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""ROS message and msgpack payload builders for teleop_ros2_node."""
+"""Pure publication-output builders for teleop_ros2_node."""
 
 import time
 from typing import Dict, Sequence
 
+import msgpack
+import msgpack_numpy as mnp
 import numpy as np
-from geometry_msgs.msg import Pose, PoseStamped
+from geometry_msgs.msg import Pose, PoseStamped, TransformStamped, TwistStamped
 from scipy.spatial.transform import Rotation
 from sensor_msgs.msg import JointState
+from std_msgs.msg import ByteMultiArray
 from teleop_ros2_interfaces.msg import NamedPoseArray
 
 from isaacteleop.retargeting_engine.interface import OptionalTensorGroup
@@ -26,6 +29,7 @@ from constants import BODY_JOINT_NAMES, HAND_POSE_JOINT_INDICES, HAND_POSE_NAMES
 from geometry import (
     apply_manus_controller_to_hand_pose,
     apply_transform_to_pose,
+    make_transform,
     to_pose,
 )
 from tensor_group_helpers import (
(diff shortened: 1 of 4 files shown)


examples/teleop_ros2/python/teleop_ros2_node.py
```diff
@@ -154,33 +112,28 @@ def _create_publishers(self) -> None:
...
+        ee_pose_msg, wrist_tfs = build_ee_output_from_controllers(
+            result["controller_left"],
+            result["controller_right"],
             now,
             self._params.world_frame,
+            self._params.left_wrist_frame,
+            self._params.right_wrist_frame,
             self._params.transform_rotation,
             self._params.transform_translation,
             self._params.controller_uses_hands_source,
```
kajananchinniahNV, line 124: Nit: should we consider a param dict or a better way to keep together at least the frames?

I can see pros and cons for having one or not so I'm fine with not making this change

examples/teleop_ros2/python/messages.py
```diff
@@ -100,17 +104,51 @@ def _compute_ee_pose_from_hand(
...
+                    pose.orientation.y,
+                    pose.orientation.z,
+                    pose.orientation.w,
+                ],
+            )
+        )
+    return transforms
+
+
+def build_controller_msg(
```
kajananchinniahNV, line 144: Nit: is there a better way to make it more clear from function name that this may not return anything?

examples/teleop_ros2/python/teleop_ros2_node.py
```diff
@@ -154,33 +112,28 @@ def _create_publishers(self) -> None:
...
-        right_ctrl = result["controller_right"]
-        if left_ctrl.is_none and right_ctrl.is_none:
-            return
-
-        controller_payload = build_controller_payload(left_ctrl, right_ctrl)
-        payload = msgpack.packb(controller_payload, default=mnp.encode)
-        controller_msg = ByteMultiArray()
-        controller_msg.data = tuple(bytes([a]) for a in payload)
-        self._pub_controller.publish(controller_msg)
+        controller_msg = build_controller_msg(
```
kajananchinniahNV: Call this variable maybe_controller_msg?

examples/teleop_ros2/python/teleop_ros2_node.py
```diff
@@ -193,15 +146,9 @@ def _publish_finger_joints(self, result: SessionResult, now) -> None:
...
-        full_body_data = result["full_body"]
-        if full_body_data.is_none:
-            return
-
-        body_payload = build_full_body_payload(full_body_data)
-        payload = msgpack.packb(body_payload, default=mnp.encode)
-        body_msg = ByteMultiArray()
-        body_msg.data = tuple(bytes([a]) for a in payload)
-        self._pub_full_body.publish(body_msg)
+        body_msg = build_full_body_msg(result["full_body"])
```
kajananchinniahNV: maybe_body_msg instead of body_msg?

examples/teleop_ros2/python/teleop_ros2_node.py
```diff
@@ -215,72 +162,47 @@ def _publish_hand_poses(self, result: SessionResult, now) -> None:
...
             self._tf_broadcaster.sendTransform(wrist_tfs)
 
     def _publish_head(self, result: SessionResult, now) -> None:
-        head = result["head"]
-        if not head_is_valid(head):
-            return
-
-        head_msg = build_head_msg(
-            head,
+        head_output = build_head_output(
```
kajananchinniahNV: maybe_head_output - there are probably a few other places we need it (haven't commented on it)

kajananchinniahNV (approved):
LGTM! Just minor nits

examples/teleop_ros2/python/teleop_ros2_node.py
sgrizan-nv: If we later introduce a shared publish/frame context object, bundling them there would be natural — but that feels like a separate cleanup rather than something this PR needs.

examples/teleop_ros2/python/messages.py
sgrizan-nv: The `| None` return type and docstring are intentional for that. Renaming to something like `try_build_controller_msg` would make absence louder, but then we’d want the same convention for all functions that may return None. Prefer keeping the current build_* names unless we want a broader rename pass.

examples/teleop_ros2/python/teleop_ros2_node.py
sgrizan-nv: Sure, done

examples/teleop_ros2/python/teleop_ros2_node.py
sgrizan-nv: Done

examples/teleop_ros2/python/teleop_ros2_node.py
sgrizan-nv: Done

PR #781, opened 2026-07-16
refactor(deviceio): generic FullBodyTracker + session-level vendor selection

aristarkhovNV:
## Description

Make the full-body tracker agnostic to both vendor and live/replay. The tracker becomes a bare marker; the vendor is chosen at session construction, mirroring how live-vs-replay is already chosen.

- **Generic rename.** `FullBodyTrackerPico` → `FullBodyTracker` and the matching schema/impl/traits types drop their `Pico` suffix. The native XR impl stays `LiveFullBodyTrackerPicoImpl`, now registered as the `body.pico-xr` vendor (the default).
- **Vendor selection API.** New `TrackerVendor{id, params}` + `VendorConfig`, threaded through the live `DeviceIOSession` and exposed per-source in the teleop session. The live factory owns a single vendor dispatch table (id → extensions + build); unknown ids, vendor selections on non-vendored trackers, and replay-mode selections are rejected up front.
- **Replay stays vendor-neutral** — it reads the recorded full-body channel regardless of which live vendor produced it.
- **Backward compatible.** Deprecated aliases keep the old `Pico` names (marker, schema types, retargeting pose types) working with a deprecation warning.
- **Tests.** Added live vendor-validation coverage and updated schema/teleop-session tests for the renames and vendor plumbing.

## Type of change

- [ ] Documentation update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added vendor selection for full-body trackers in live sessions.
  * Added vendor-aware OpenXR extension discovery and validation.
  * Introduced the vendor-neutral `FullBodyTracker` and body-joint APIs.
  * Added vendor selection support to teleoperation sources and sessions.
  * Replay sessions remain vendor-neutral.

* **Documentation**
  * Documented vendor selection, defaults, validation, live-session behavior, and replay compatibility.

* **Bug Fixes**
  * Invalid, unknown, duplicate, or unsupported vendor selections now raise clear errors.
  * Replay mode rejects live-only vendor selections.

* **Compatibility**
  * Existing Pico-prefixed names remain available as deprecated aliases with warnings.


src/core/teleop_session_manager/python/teleop_session.py
```diff
@@ -969,9 +969,27 @@ def _add_tracker(tracker: Any) -> None:
...
+                            f"DeviceIO source '{source_name}'. Known sources: "
+                            f"{sorted(source_by_name)}"
+                        )
+                    vendor_entries.append((source.get_tracker(), vendor))
+                vendor_config = deviceio.VendorConfig(vendor_entries)
+
             # Get required extensions from all trackers
             required_extensions = deviceio.DeviceIOSession.get_required_extensions(
-                trackers
+                trackers, vendor_config
```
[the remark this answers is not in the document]
aristarkhovNV: fixed

src/core/deviceio_base/cpp/inc/deviceio_base/tracker_vendor.hpp
```diff
@@ -0,0 +1,29 @@
...
+// single tracker type stays vendor-agnostic.
+//
+// `id` is a string so new pre-built plugin vendors can be selected without
+// recompiling core (e.g. "body.pico-xr"). `params` carries vendor-specific
+// settings as free-form strings, mirroring plugin CLI arguments
+// (e.g. {"max_flatbuffer_size": "16384"}).
+struct TrackerVendor
+{
+    std::string id;
+    std::map<std::string, std::string> params;
```
jiwenc-nv, line 29: **must-fix:** `params` is a dead knob — it's documented with a worked example (`max_flatbuffer_size`) and bound to Python, but nothing reads it: `create_full_body_tracker_pico_impl` never consults `find_vendor()`, and `validate_vendor_selections` checks only `id`. Any params dict is silently accepted and dropped, and tightening later is a behavior break for configs that "worked".

Please either **delete the field + binding** until the first consumer lands (restoring it is additive and kwarg-compatible), or **reject non-empty params** in `validate_vendor_selections` so the contract starts strict.

src/core/teleop_session_manager/python/teleop_session.py
```diff
@@ -969,9 +969,27 @@ def _add_tracker(tracker: Any) -> None:
             for tracker in self.config.trackers:
                 _add_tracker(tracker)
 
+            # Resolve per-source vendor selections (keyed by source name) into a
```
jiwenc-nv: **must-fix:** this vendor-resolution block only runs on the live path — in `SessionMode.REPLAY` a non-empty `config.tracker_vendors` is silently ignored. That's inconsistent with the fail-fast convention this PR establishes everywhere else (unknown ids, non-vendored trackers, and the unknown-source `ValueError` just below all raise). A `mode` typo currently masks a vendor selection without a word.

Please raise when `tracker_vendors` is non-empty in REPLAY mode — the `__post_init__` mode-consistency block in `config.py` (next to the REPLAY-without-`mcap_config` check) is the natural home. A test pinning this would be great.

src/core/teleop_session_manager/python/config.py
```diff
@@ -386,6 +396,7 @@ class TeleopSessionConfig:
     verbose: bool = True
     oxr_handles: Optional[OpenXRSessionHandles] = None
     mcap_config: Optional[Union[McapRecordingConfig, McapReplayConfig]] = None
+    tracker_vendors: Dict[str, TrackerVendor] = field(default_factory=dict)
```
jiwenc-nv: **design question (your call):** is the manager-level `tracker_vendors` surface needed in this PR at all? Nothing in-repo sets it, and its only legal value today is the default — a no-op. Deleting it until a second vendor exists would moot the replay-mode question and the discoverability work at this layer, and re-adding it later is purely additive (default-empty dict, no callers). Totally fine to keep it with the fixes requested elsewhere — I'd just like the choice to be deliberate rather than incidental.

src/core/deviceio_trackers/cpp/inc/deviceio_trackers/full_body_tracker.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+private:
+    static constexpr const char* TRACKER_NAME = "FullBodyTracker";
+};
+
+// Deprecated alias for the renamed FullBodyTracker (was FullBodyTrackerPico before the
+// vendor-agnostic rename). Retained so existing C++ consumers keep compiling; prefer
+// FullBodyTracker. Mirrors the Python alias and the ReplayFullBodyTrackerPicoImpl /
+// FullBodyPicoRecordingTraits C++ aliases added in the same change.
+using FullBodyTrackerPico = FullBodyTracker;
```
jiwenc-nv: **should-fix (deprecation):** the aliases here and in `recording_traits.hpp`, `replay_full_body_tracker_impl.hpp`, and the Python inits are comment-only — no `[[deprecated]]`, no `DeprecationWarning`, and `FullBodyTrackerPico` keeps equal billing in `__all__`. Since this PR sets the repo's deprecation precedent, make it machine-readable: `using FullBodyTrackerPico [[deprecated("use FullBodyTracker")]] = FullBodyTracker;` and a module-`__getattr__` warning on the Python side.

Also, the "existing C++ consumers keep compiling" claim is only partially true: `full_body_tracker_pico.hpp` and `full_body_tracker_pico_base.hpp` are deleted (include-path break) and there's no `IFullBodyTrackerPicoImpl` alias — either add shim headers + that alias, or drop the claim from the comments/PR description.

A bigger idea that may worth discussing: do we want to make this PR backwards compatible? Or just mark it as a breaking change?

src/core/deviceio_base/cpp/inc/deviceio_base/full_body_tracker_base.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+struct FullBodyPosePicoTrackedT;
+
+// Abstract base interface for full body tracker implementations.
+// Vendor-agnostic: every live/replay backend (native XR, pushed tensor, ...)
+// implements this and produces the same FullBodyPosePicoTrackedT payload.
+class IFullBodyTrackerImpl : public ITrackerImpl
+{
+public:
+    virtual const FullBodyPosePicoTrackedT& get_body_pose() const = 0;
```
jiwenc-nv: **should-fix (vendor contract):** this interface pins the joint *layout* (24-joint XR_BD) but not the *frame*. The native impl expresses joints in the session's `OpenXRSessionHandles::space` reference space; a future pushed-tensor vendor sourcing e.g. a mocap suit has no documented obligation to match that space, handedness, or meters. One sentence here — "poses are expressed in the session reference space, meters, OpenXR conventions" — turns an implicit assumption into a testable vendor contract, so the second vendor's author doesn't reverse-engineer conventions from the Pico impl.

src/core/mcap/cpp/inc/mcap/recording_traits.hpp
```diff
@@ -38,13 +38,18 @@ struct ControllerRecordingTraits
     static constexpr std::array replay_channels = { "left_controller", "right_controller" };
 };
 
-struct FullBodyPicoRecordingTraits
+struct FullBodyRecordingTraits
 {
     static constexpr std::string_view schema_name = "core.FullBodyPosePicoRecord";
```
jiwenc-nv: **proposal — finish the de-Pico rename at the schema level, preferably in this PR (your call):** the payload types (`FullBodyPosePico*` in `full_body.fbs`) and this recorded schema name still leak Pico through the vendor-agnostic facade, so the rename is currently skin-deep at the data layer.

I checked that the rename is safe for replay: `McapTrackerViewers` matches by **topic only** (see the topic filter in `tracker_channels.hpp`) and never compares the schema name, and FlatBuffers wire format is name-independent — renaming the tables (keeping field ids) and writing `core.FullBodyPoseRecord` into new recordings breaks nothing in-repo, in either direction, for old or new files.

I'd prefer it in this PR so the whole rename lands as one deprecation cycle — the Python alias pattern you've established here covers the schema exports verbatim. If you'd rather keep this PR's scope as-is, an **immediate** follow-up is fine too, with one constraint: it should land before the next dataset-producing release, so recordings only ever cross a single schema-name epoch. Either way, two guardrails: (1) a quick audit of external consumers keyed on `core.FullBodyPosePicoRecord` (Foxglove/Lichtblick layouts, dataset scripts) first; (2) keep `LiveFullBodyTrackerPicoImpl` Pico-named — it genuinely *is* the `body.pico-xr` vendor — and document that the "generic" payload is the canonical XR_BD 24-joint skeleton so the new name doesn't overclaim.

jiwenc-nv (commented):
The architecture is right: vendor-as-session-concern mirrors the existing live/replay split, the dispatch-table change is behavior-preserving for the other tracker types, and keeping the recorded schema/channel byte-identical preserves replay of existing recordings. The rename half is load-bearing on its own. My blocking items are two small contract defects on the new public surface (inline comments); both are same-day fixes.

**Should-fix not anchored below:** `helpers.py::get_required_oxr_extensions_from_pipeline` takes no vendor argument while required extensions are now vendor-dependent, so the documented external-`oxr_handles` flow computes wrong extensions for any future non-default vendor. Please add the optional parameter now, while the surface is fresh.

**Nits (take or leave):** `tracker_supports_vendors` type-probes via a side-effectful `collect_extensions` call with a discarded scratch set — an explicit type predicate would be cleaner; vendor ids are validated globally rather than per tracker type (a cross-type pairing fails later with a vaguer error); the same not-in-list failure is worded differently in `deviceio_session.cpp` vs `live_deviceio_factory.cpp`; error messages identify trackers only by vendor id — include `get_name()`; the `[FullBodyTrackerPico]` log prefix in `live_full_body_tracker_pico_impl.cpp` now names the deprecated alias of a different class; a `get_tracker_vendors` accessor on the `VendorConfig` binding would match its `McapRecordingConfig` sibling; "selected without recompiling core" in `tracker_vendor.hpp` is aspirational while the registry is a compile-time table.

src/core/deviceio_base/cpp/inc/deviceio_base/tracker_vendor.hpp
aristarkhovNV: enforcing empty for now

src/core/teleop_session_manager/python/config.py
aristarkhovNV: the field was removed

src/core/deviceio_trackers/cpp/inc/deviceio_trackers/full_body_tracker.hpp
aristarkhovNV: added machine readable deprecation. @sgrizan-nv @rwiltz please comment on whether to maintain backward compatibility and if so for how long?

src/core/deviceio_base/cpp/inc/deviceio_base/full_body_tracker_base.hpp
aristarkhovNV: we can no longer refer to xr session as the tracker is more generic now. With that said we could consider borowing the wording from the original extension spec itself and standartize it for this tracker.

src/core/mcap/cpp/inc/mcap/recording_traits.hpp
aristarkhovNV: renamed pico schema

src/core/teleop_session_manager/python/teleop_session.py
aristarkhovNV: fixed

jiwenc-nv (approved):

src/core/deviceio_trackers/cpp/inc/deviceio_trackers/full_body_tracker.hpp
aristarkhovNV: keeping back compat aliases for now, will drop later if necessary

src/core/deviceio_session/cpp/inc/deviceio_session/deviceio_session.hpp
```diff
@@ -47,13 +63,17 @@ class DeviceIOSession : public ITrackerSession
...
 
     // Static factory - Create and initialize a session with trackers.
-    // Optionally pass a McapRecordingConfig to enable automatic MCAP recording.
+    // Optionally pass a McapRecordingConfig to enable automatic MCAP recording, and a
+    // VendorConfig to select the vendor for any vendored trackers.
     static std::unique_ptr<DeviceIOSession> run(const std::vector<std::shared_ptr<ITracker>>& trackers,
                                                 const OpenXRSessionHandles& handles,
-                                                std::optional<McapRecordingConfig> recording_config = std::nullopt);
+                                                std::optional<McapRecordingConfig> recording_config = std::nullopt,
+                                                VendorConfig vendor_config = {});
```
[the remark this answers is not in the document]
aristarkhovNV: fixed

PR #782, opened 2026-07-16
Fix resource leaks when TeleopSession setup fails

fallintoplace:
## Description

`TeleopSession.__enter__` acquired OpenXR, DeviceIO, and plugin contexts onto the session-owned `ExitStack`. If a later setup step raised, the `with` body never started and Python did not call `TeleopSession.__exit__`, leaving the resources acquired so far alive.

This change makes session startup transactional:

- acquire resources on a temporary `ExitStack`
- on failure, unwind the temporary stack in reverse order and clear the stale OpenXR reference
- keep propagating the setup error even if a child suppresses it or raises during rollback; rollback failures are logged
- transfer stack ownership to the session only after every setup step succeeds
- reject nested entry so an active session stack cannot be overwritten

There is no public API change.

The parameterized regression test fails OpenXR, DeviceIO, the first plugin, and a later plugin in turn, checking exact-once reverse-order cleanup. Additional tests cover rollback cleanup failure and nested entry while preserving normal reuse after exit.

## Type of change

- [ ] Documentation update

## Testing

- `ctest --test-dir build -R "^teleop_session_manager_" --output-on-failure --parallel 4` (4/4 passed, macOS arm64, Release, Python 3.11)

## Checklist

- [x] Documentation is not required because the public API is unchanged
- [x] I have added tests that prove the fix works
- [x] I have signed off all commits (`git commit -s`) per the [DCO](../CONTRIBUTING.md#signing-your-work)

### DIFF
--- a/src/core/teleop_session_manager/python/teleop_session.py
+++ b/src/core/teleop_session_manager/python/teleop_session.py
@@ -208,8 +208,8 @@ class TeleopSession:
         self._active_retargeting_execution_mode: Optional[RetargetingExecutionMode] = (
             None
         )
-        self._is_active: bool = False
-        self._setup_complete: bool = False
+        # True from the start of __enter__ until __exit__ or rollback finishes.
+        self._in_context: bool = False
         # Discover sources and external leaves from pipeline
         self._discover_sources()
 
@@ -919,12 +919,15 @@ class TeleopSession:
 
         Returns:
             self for context manager protocol
+
+        Raises:
+            RuntimeError: If the session is already active.
         """
-        if self._is_active:
+        # Detect nested context-manager use; this does not make the session thread-safe.
+        if self._in_context:
             raise RuntimeError("TeleopSession is already active")
 
-        self._is_active = True
-        self._setup_complete = False
+        self._in_context = True
         stack = ExitStack()
         try:
             self._enter_resources(stack)
@@ -937,10 +940,9 @@ class TeleopSession:
                 logger.exception("Failed to roll back TeleopSession setup")
             finally:
                 self._oxr_session = None
-                self._is_active = False
+                self._in_context = False
             raise
 
-        self._setup_complete = True
         return self
 
     def _enter_resources(self, stack: ExitStack) -> None:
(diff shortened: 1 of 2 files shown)


aristarkhovNV (approved):

src/core/teleop_session_manager/python/teleop_session.py
```diff
@@ -208,6 +208,7 @@ def __init__(self, config: TeleopSessionConfig):
         self._active_retargeting_execution_mode: Optional[RetargetingExecutionMode] = (
             None
         )
+        self._is_active: bool = False
```
jiwenc-nv: Now that `_is_active` exists, `_setup_complete` looks fully subsumed: its only production reader is the `__exit__` guard, and the two flags only diverge after a clean exit (`_is_active` goes back to False, `_setup_complete` stays True forever). In that one divergent state the `_is_active` behavior is strictly better — a stray direct `__exit__()` call becomes a no-op instead of re-running the teardown body against the spent stack.

Could we consolidate to a single flag? Concretely: guard `__exit__` with `if not self._is_active:`, drop `_setup_complete` (declaration + the two writes in `__enter__`), and update the handful of test asserts on it — ideally replacing them with the public contract (e.g. after a failed `__enter__`, `with session: pass` proves the session is reusable). With that guard in place, the defensive `self._exit_stack = ExitStack()` reassignment in `__exit__` also becomes removable if you want.

Naming thought while here: `_is_active` can read as "setup succeeded", but it's True from the top of `__enter__`, before setup completes. Something like `_in_context` with a one-line lifecycle comment ("True from the start of `__enter__` until `__exit__`/rollback finishes; rejects nested entry and gates teardown") would be unambiguous.

src/core/teleop_session_manager/python/teleop_session.py
```diff
@@ -919,6 +920,31 @@ def __enter__(self):
         Returns:
             self for context manager protocol
         """
+        if self._is_active:
```
jiwenc-nv: The docstring should mention the new behavior: `Raises: RuntimeError: If the session is already active.` — nested entry is an observable change (previously a double `__enter__` "worked" by stacking onto the same `ExitStack`). Might also be worth a one-liner that this check-then-set is misuse detection, not thread safety.

jiwenc-nv (changes requested):

jiwenc-nv (approved):

PR #791, opened 2026-07-17
Add Wuji glove input plugin and Wuji hand retargeter

wuji-tech-dev:
## Summary

This PR adds Wuji glove and dexterous-hand support to IsaacTeleop:

- Add an optional out-of-process Wuji glove plugin based on Wuji SDK
  v2026.7.14.
- Map the glove's 21-joint skeleton to the existing OpenXR 26-joint
  `HandPose` representation. Unavailable joints are marked invalid.
- Reuse the existing `HandTracker` → `HandsSource` → `HandInput`
  pipeline without introducing a new schema.
- Support left/right gloves, stale-data handling, disconnection and
  automatic reconnection.
- Add `WujiHandRetargeter` to convert hand-tracking input into 20 Wuji
  hand joint commands.
- Add a teleoperation example supporting both Wuji Hand and Wuji Hand 2,
  along with synthetic and replay modes.

## Testing

- Verified plugin installation and hand skeleton streaming.
- Verified glove disconnection and automatic reconnection.
- Verified end-to-end glove-to-hand teleoperation.
- Verified synthetic and replay modes without physical devices.
- Passed pre-commit and C++ formatting checks.

## Known limitation

The glove skeleton is currently wrist-relative and provides 21 of the 26
OpenXR hand joints. A future update will add wrist-pose fusion and complete
26-joint hand tracking.

## Summary by CodeRabbit

* **New Features**
  * Added Wuji Glove support for OpenXR hand tracking and teleoperation.
  * Added wrist pose sourcing from optical hand tracking or controllers, with fallback handling.
  * Added Wuji hand retargeting for Wuji Hand and Wuji Hand 2 models.
  * Added a demo supporting synthetic, replay, and live hardware modes.
  * Added optional build and installation support for the Wuji plugin.

* **Documentation**
  * Added setup, usage, troubleshooting, and retargeting guides for Wuji hardware.

* **Bug Fixes**
  * Improved static web responses with explicit content length and MIME headers.


src/plugins/wuji_glove/README.md
```diff
@@ -0,0 +1,65 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 Wuji Technology. All rights reserved.
```
jiwenc-nv, line 2: @wuji-tech-dev please turn this md file into a rst under docs/source/devices

src/plugins/wuji_glove/wuji_glove_plugin.cpp
```diff
@@ -0,0 +1,429 @@
...
+    XR_HAND_JOINT_RING_INTERMEDIATE_EXT,
+    XR_HAND_JOINT_RING_DISTAL_EXT,
+    XR_HAND_JOINT_RING_TIP_EXT,
+    XR_HAND_JOINT_LITTLE_PROXIMAL_EXT,
+    XR_HAND_JOINT_LITTLE_INTERMEDIATE_EXT,
+    XR_HAND_JOINT_LITTLE_DISTAL_EXT,
+    XR_HAND_JOINT_LITTLE_TIP_EXT,
+};
+
+constexpr XrSpaceLocationFlags kValidFlags =
```
jiwenc-nv: Please fix: IIUC, this PR doesn't provide 6DoF pose for the hand posision?  Can you do something similar to what Manus has?
That make the pose fushion for all the hand joints?

A bit more details: Because the plugin injects through the typed push-device hand channel (`HandInjector` → `xrPushDevicePushHandTrackingNV`), the injected hand is indistinguishable from a native `XR_EXT_hand_tracking` hand to every consumer of `/hand/left|right` — recorders, visualizers, and wrist-consuming retargeters will trust these bits. Please fuse a device wrist pose, following the manus plugin's pattern: wrist source from XDEV hand tracking with controller-aim fallback (`src/plugins/manus/core/manus_hand_tracking_plugin.cpp:800-830`), skeleton placed at the fused wrist, and the TRACKED bits set only while the wrist source is actually tracked (`:886-892`). Ideally factor the wrist-source machinery into `plugin_utils`, since two glove plugins now need it. If no wrist source is available at runtime, degrade to honest flags: VALID without TRACKED.

src/plugins/wuji_glove/wuji_glove_plugin.cpp
```diff
@@ -0,0 +1,429 @@
...
+        j = XrHandJointLocationEXT{};
+        j.locationFlags = 0;
+        j.radius = kDefaultJointRadius;
+        j.pose.orientation = XrQuaternionf{ 0.0f, 0.0f, 0.0f, 1.0f };
+    }
+
+    for (int mp = 0; mp < 21; ++mp)
+    {
+        const WujiSkeletonJoint& src = frame->joints[mp];
+        XrHandJointLocationEXT& dst = out[kMpToXr[mp]];
```
jiwenc-nv, line 145: Please fix: the SDK joint orientations are copied verbatim and stamped `ORIENTATION_VALID`, but nothing states the Wuji skeleton's per-joint axis convention. `XR_EXT_hand_tracking` prescribes a specific basis (−Z along the bone toward the wrist, +Y through the back of the hand), and `SharpaHandRetargeter` consumes `JOINT_ORIENTATIONS` through the same `HandsSource` path, so it will trust whatever basis these quaternions are in. Please verify the SDK basis against the OpenXR convention, remap at conversion if it differs, and document it here. This also gates the wrist-fusion request above: composing `wrist_pose × wrist-relative joint pose` requires knowing the SDK's local frame. Note that clearing `ORIENTATION_VALID` instead won't work — `live_hand_tracker_impl.cpp:430` requires position *and* orientation valid per joint, which would zero out your own pipeline.

examples/retargeting/python/wuji_hand_retargeter_demo.py
```diff
@@ -0,0 +1,325 @@
...
+
+        # Finish the current step before handling Ctrl+C so device cleanup completes.
+        previous_sigint_handler = signal.signal(signal.SIGINT, request_stop)
+        try:
+            with TeleopSession(session_config) as session:
+                print("Teleoperating (Ctrl+C to stop)...")
+                while not stop_requested:
+                    t0 = time.monotonic()
+                    result = session.step()
+                    send(list(result["hand_joints"]))  # 20 values, firmware order
```
jiwenc-nv: Please fix: on any tracking dropout (hand absent, or a single flickering joint tripping the retargeter's all-or-nothing validity gate), `WujiHandRetargeter` emits all-zero joint angles, and this loop forwards them straight to the hardware at 120 Hz. 

Am I understand this correctly?

On the Wuji Hand 2 path that's a raw MIT-mode `JointCommand` with kp=3.0 and no filtering (the Hand 1 path at least has the 5 Hz `LowPass`), so a 200 ms glove dropout mid-grasp becomes an instantaneous snap-to-zero on a real hand. Please hold the last command (or ramp) on the loss transition, and run Hand 2 through the same low-pass as Hand 1.

src/plugins/wuji_glove/install.sh
```diff
@@ -0,0 +1,75 @@
...
+        aarch64 | arm64) arch="aarch64-linux-gnu" ;;
+        x86_64) arch="x86_64-linux-gnu" ;;
+        *)
+            echo "unsupported arch: $(uname -m)" >&2
+            exit 1
+            ;;
+    esac
+    name="wuji-sdk-c-${WUJI_SDK_C_VERSION}-${arch}"
+    url="${WUJI_SDK_C_URL_BASE}/v${WUJI_SDK_C_VERSION}/${name}.tar.gz"
+    work="$(mktemp -d)"
```
jiwenc-nv: Please fix: the SDK extracts into this `mktemp -d` directory, which the EXIT trap deletes — but `WUJI_SDK_INCLUDE_DIR`/`WUJI_SDK_LIB` pointing into it are baked into the persistent CMake cache. The first build works; every later reconfigure/rebuild fails on a missing `wuji_sdk.h` (and `cmake --install` would try to install the vendor libs from the deleted dir), and since the cache variables are still set, the friendly `FATAL_ERROR` guard never fires. Please extract to a persistent location instead (e.g. under the build tree — `install_manus.sh` sets the precedent by extracting next to the script).

src/plugins/wuji_glove/install.sh
```diff
@@ -0,0 +1,75 @@
...
+    echo "==> Downloading ${name}.tar.gz"
+    curl -fL --retry 3 "$url" -o "$work/${name}.tar.gz" \
```
jiwenc-nv: Please fix: this downloads a binary SDK over the network and links it into the build with no integrity check. `install_manus.sh:184-195` already sets the house pattern: pin per-arch SHA-256s and hard-fail on mismatch. Please do the same for the wuji-sdk-c tarball.

src/plugins/wuji_glove/wuji_glove_plugin.cpp
```diff
@@ -0,0 +1,429 @@
...
+void WujiGlovePlugin::on_skeleton(const WujiHandSkeleton* frame, bool is_left)
+{
+    std::array<XrHandJointLocationEXT, XR_HAND_JOINT_COUNT_EXT> joints{};
+    if (!convert_skeleton(frame, joints))
+    {
+        return;
+    }
+
+    std::lock_guard<std::mutex> lock(m_frame_mutex);
+    HandFrame& slot = is_left ? m_left : m_right;
```
jiwenc-nv, line 536: Please fix: hand-slot selection is by side only, so if two same-side gloves are ever visible (a spare powered on, or a second pair on the same LAN), their frames interleave last-writer-wins into one injected hand — silently. Assuming gloves operate as a single left/right pair is fine for this plugin, but please state that assumption in a comment here (and/or in `discover_gloves`), ideally with a log line when a second same-side glove is discovered so the failure mode isn't silent.

examples/retargeting/python/wuji_hand_retargeter_demo.py
```diff
@@ -0,0 +1,325 @@
...
+A retargeter is device-agnostic; the device binding lives only in the opt-in
+``drive`` mode, whose hardware/session imports are loaded lazily so the
+device-free modes run without them.
+
+Prereqs:
+    pip install 'isaacteleop[wuji]'    # brings in wuji-sdk[retarget]
+
+Usage:
+    python wuji_hand_retargeter_demo.py
+    python wuji_hand_retargeter_demo.py --mode replay --replay data/avp1.pkl
```
jiwenc-nv: Please fix: this usage example references `data/avp1.pkl`, but no such file ships and nothing in the tree can produce one — the replay pickle format (`{"left_fingers": (21,3)|None, ...}`) has no producer or sample here. Please update the docstring and the `--replay` help to say where these recordings come from and how to obtain or produce one (a link to Wuji's official docs or recorder tooling is fine), plus the keypoint provenance and convention (source device, keypoint order, units).

src/retargeters/wuji_hand_retargeter.py
```diff
@@ -0,0 +1,207 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 Wuji Technology. All rights reserved.
```
jiwenc-nv, line 1: Please fix: every comparable retargeter has a unit test under `src/core/retargeting_engine_tests/python/` (`test_sharpa_hand_retargeter.py` is the closest template). Please add one for `WujiHandRetargeter`, gated with `pytest.importorskip("wuji_sdk")` — spec shapes, zeros-on-invalid, reset-on-gap — and have it cross-assert that the C++ `kMpToXr` table and the Python `OPENXR_TO_MEDIAPIPE_INDICES` table agree, since a silent divergence between them would be a joint-permutation bug nothing at runtime can catch.

examples/retargeting/python/wuji_hand_retargeter_demo.py
```diff
@@ -0,0 +1,325 @@
...
+    if hand_dev is None:
+        print("No Wuji Hand / Wuji Hand 2 found")
+        manager.disconnect_all()
+        return 1
+
+    hand = manager.connect(sn=hand_dev.sn, device_name=hand_dev.sn)
+    is_hand2 = hand_dev.device_type == DeviceType.WujiHand2
+
+    # The retarget model must match the connected hardware.
+    model = "wuji_hand_2" if is_hand2 else "wuji_hand"
```
jiwenc-nv: Please fix: this silently overwrites the user's `--model` with the detected hardware, so `--model` is dead in drive mode (and the flag has no help text at all). Please either warn/error when an explicitly-passed value disagrees with the detected hand, or drop `model` from `run_drive`'s signature — and add a `--model` help string noting it's auto-detected in drive mode.

src/retargeters/wuji_hand_retargeter.py
```diff
@@ -0,0 +1,207 @@
...
+            output_group[i] = 0.0
+        # Drop warm-start / smoothing state so we restart cleanly next valid frame.
+        self._session.reset()
+
+    def _compute_fn(self, inputs: RetargeterIO, outputs: RetargeterIO, context) -> None:
+        output_group = outputs["hand_joints"]
+        hand_group = inputs[self._input_key]
+
+        # Honor an explicit reset event (e.g. session restart): drop session state.
+        events = getattr(context, "execution_events", None)
```
jiwenc-nv: Nit: `ComputeContext` always carries `execution_events` (default-constructed in `retargeter_core_types.py`), and the sibling retargeters access it directly (e.g. `locomotion_retargeter.py:123`). Prefer `if context.execution_events.reset:` over the double `getattr` — the defensiveness hides the real contract.

src/retargeters/wuji_hand_retargeter.py
```diff
@@ -0,0 +1,207 @@
...
+there is no config path or URDF to manage. The session keeps warm-start +
+smoothing state across frames, so it is built once (a heavy, multi-second call)
+and reused; ``reset()`` is called after a tracking gap.
+
+Requires ``isaacteleop[wuji]`` → ``wuji-sdk[retarget]``.
+"""
+
+import logging
+from dataclasses import dataclass
+from typing import Optional
```
jiwenc-nv: Nit: the repo targets Python ≥ 3.10 and the neighboring retargeters use PEP 604 unions (`sharpa_hand_retargeter.py:103`: `list[str] | None`). Prefer `hand_joint_names: list[str] | None = None` and drop this import. Same theme in the demo: the quoted `"str | None"` annotations at `wuji_hand_retargeter_demo.py:167-168` can be unquoted, and the demo is missing the `#!/usr/bin/env python3` shebang its sharpa counterpart has.

src/plugins/wuji_glove/wuji_glove_plugin.cpp
```diff
@@ -0,0 +1,429 @@
...
+        {
+            std::lock_guard<std::mutex> lock(m_frame_mutex);
+            left_copy = m_left;
+            right_copy = m_right;
+        }
+
+        pump_hand(m_left_injector, XR_HAND_LEFT_EXT, left_copy, time);
+        pump_hand(m_right_injector, XR_HAND_RIGHT_EXT, right_copy, time);
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(16));
```
jiwenc-nv: Nit: this file names its constants (`kShutdownPollInterval`, `kDefaultJointRadius`) except the timing literals — the 16 ms pump period here, the 200 ms staleness window in `pump_hand`, and the 1 s discovery interval (`i < 10`). Hoisting `kFramePeriod` / `kStaleThreshold` / `kDiscoveryInterval` next to the existing constants (per `so101_leader`) would also fix `kShutdownPollInterval` doing double duty as the generic sleep quantum. Same for the magic `21` (a `kNumMediaPipeJoints` or `std::size(kMpToXr)`). And with two threads, `m_thread` reads better as `m_worker_thread` next to `m_connection_thread`.

src/plugins/wuji_glove/wuji_glove_plugin.cpp
```diff
@@ -0,0 +1,429 @@
...
+        std::cerr << "WujiGlovePlugin: could not determine hand_side for " << connection.serial << std::endl;
+        wuji_dev_disconnect(device);
+        wuji_dev_release(device);
+        return false;
+    }
+
+    auto context = std::make_unique<SubContext>();
+    context->self = this;
+    context->is_left = *is_left;
+    wuji_glove_set_emf_poses_rate_divider(device, 2);
```
jiwenc-nv: Nit: `wuji_glove_set_emf_poses_rate_divider(device, 2)` is the one unchecked, uncommented SDK call in an otherwise well-commented file. What does the divider gate, what's the glove's native rate, and why 2? A comment tying it to the 16 ms pump and the 200 ms staleness window would make the latency budget auditable (below ~60 Hz effective rate, the pump re-injects identical samples under fresh timestamps), plus a log if the call fails.

src/retargeters/wuji_hand_retargeter.py
```diff
@@ -0,0 +1,207 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+# OpenXR (26 joints) -> MediaPipe (21 joints) index mapping.
+# Skips OpenXR palm(0) and the non-thumb metacarpal joints. The MediaPipe
+# 21-landmark order (wrist, thumb 1-4, index 1-4, middle 1-4, ring 1-4,
+# pinky 1-4) is identical to the MANO 21-joint order, so this is the same
+# table the Sharpa retargeter uses.
+OPENXR_TO_MEDIAPIPE_INDICES: list[int] = [
```
jiwenc-nv, line 51: Nit: this table and the C++ `kMpToXr` in `wuji_glove_plugin.cpp` are the same 21-entry mapping in opposite directions (and sharpa's `_OPENXR_TO_MANO_INDICES` is a third copy). A divergence would be a silent joint-permutation bug, so a one-line cross-reference comment in each pointing at the other would help them not drift. Hosting the table once (near `HandJointIndex` in `tensor_types`) is fine as a follow-up.

src/retargeters/wuji_hand_retargeter.py
```diff
@@ -0,0 +1,207 @@
...
+            ``"wuji_hand"`` or ``"wuji_hand_2"``. Selects the builtin tuning
+            config internally.
+        hand_side: ``"left"`` or ``"right"`` (the retargeter's handedness basis).
+        hand_joint_names: Optional output joint-name override (length 20, same
+            firmware order). If ``None``, uses ``_default_joint_names()``.
+    """
+
+    model: str = "wuji_hand_2"
+    hand_side: str = "right"
+    hand_joint_names: Optional[list[str]] = None
```
jiwenc-nv: Nit: `hand_joint_names` has no caller that passes a non-default value, and unlike Sharpa (where MJCF-derived names make an override plausible) the defaults here are synthetic labels. Either drop the knob (and its length-20 validation branch), or keep it with a note that it exists for parity with `SharpaHandRetargeterConfig`.

src/plugins/wuji_glove/wuji_glove.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <chrono>
+#include <csignal>
+#include <iostream>
+#include <memory>
+#include <string>
+#include <thread>
+
+using namespace plugins::wuji_glove;
+
+static_assert(ATOMIC_BOOL_LOCK_FREE, "lock-free atomic bool is required for signal safety");
```
jiwenc-nv: Nit: `static_assert(ATOMIC_BOOL_LOCK_FREE, ...)` passes when the macro is 1 ("sometimes lock-free"), so it doesn't guarantee what its message claims. Either compare `== 2` or drop it (`haptikos/main.cpp:16` carries the identical assert — if you keep a corrected form here, the sibling deserves the same fix in a follow-up).

src/plugins/wuji_glove/wuji_glove_plugin.cpp
```diff
@@ -0,0 +1,429 @@
...
+}
+
+void WujiGlovePlugin::discover_gloves()
+{
+    WujiDiscovered* list = nullptr;
+    size_t count = 0;
+    if (wuji_scan(&list, &count) != WUJI_STATUS_OK)
+    {
+        std::cerr << "WujiGlovePlugin: wuji_scan failed: " << safe_err() << std::endl;
+        wuji_discovered_free(list, count);
```
jiwenc-nv: Nit: two small teardown hardenings. (1) `wuji_discovered_free(list, count)` is also reached on the `wuji_scan` failure path with whatever the failed call left in `list`/`count` — guard with `if (list)` unless the SDK documents zero-init on failure. (2) In `disconnect_glove`, `invalidate_hand` runs before `wuji_sub_close` joins the callback thread, so an in-flight callback can re-validate the slot; invalidating again after the close would make teardown airtight (the 200 ms staleness gate bounds the damage meanwhile).

examples/retargeting/python/wuji_hand_retargeter_demo.py
```diff
@@ -0,0 +1,325 @@
...
+    if is_hand2:
+        hand.effort_limit().set(1.5)
+        hand.mit_params().set((3.0, 0.05))
```
jiwenc-nv, line 292: Nit: `effort_limit(1.5)` and `mit_params((3.0, 0.05))` are unitless magic gains in an example people will copy verbatim — please state units and provenance ("vendor-recommended safe defaults" would do). Related: drive mode never checks that the connected hand's side matches `--hand`/the glove's side, so a left glove driving a right hand mirrors silently; a one-line warning on mismatch would help.

src/plugins/wuji_glove/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+
+install(TARGETS wuji_glove_plugin
+    RUNTIME DESTINATION plugins/wuji_glove
+)
+
+install(FILES plugin.yaml README.md
+    DESTINATION plugins/wuji_glove
+)
+
+install(FILES "${WUJI_SDK_LIB}" "${WUJI_HAND_CPP_LIB}"
```
jiwenc-nv, line 62: Nit: installing the vendor `.so`s into the shared `${CMAKE_INSTALL_LIBDIR}` follows the haptikos precedent, so no change requested on location — but flagging for an explicit decision: these are proprietary Wuji binaries landing in the common libdir next to Apache-2.0 artifacts (redistribution question + soname-collision surface). If that's not intended, installing them under `plugins/wuji_glove/` next to the binary (the `$ORIGIN` build RPATH already proves the layout) is the alternative, with `INSTALL_RPATH` adjusted to match.

src/core/python/requirements-wuji.txt
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 Wuji Technology. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Runtime deps for the Wuji hand retargeter (WujiHandRetargeter, which wraps
+# wuji_sdk.retargeting). The wuji-sdk[retarget] extra pulls the retargeting
+# engine's own deps (numpy, scipy, nlopt, pin, pyyaml).
+#
+# Heads-up for aarch64: wuji-sdk[retarget] drags in nlopt. PyPI ships modern nlopt
```
jiwenc-nv: Please fix: This just happens recently:  nlopt now has a ARM64 wheel prebuilt.  So the comments where about nlopt is no longer needed.

Can you help validate and update the docs if that's the case already? Thanks!

jiwenc-nv (changes requested):

src/plugins/wuji_glove/wuji_glove_plugin.cpp
wuji-tech-dev: Done, following the Manus pattern: optical (XR_MNDX_xdev_space) preferred, controller aim plus a calibrated per-hand offset as fallback, skeleton placed at the fused wrist, and VALID-without-TRACKED when no source is available. Configurable via WUJI_GLOVE_WRIST_SOURCE. The machinery moved to plugin_utils as you suggested; migrating Manus onto it needs Manus hardware, so that is a follow-up.
On the offsets: the built-in kLeftAimToWrist / kRightAimToWrist values are calibrated against one specific controller mount and are only meaningful with that adapter, which we will publish in https://github.com/wuji-technology/wuji-description. Any other mount can be accommodated at runtime through WUJI_GLOVE_AIM_TO_WRIST_{LEFT,RIGHT} without rebuilding.
Known limitation as discussed: on the controller path TRACKED follows is_valid, because ControllerSnapshot does not carry the TRACKED bit — same as controller_synthetic_hands and haptikos today.

src/plugins/wuji_glove/README.md
wuji-tech-dev: Done, at docs/source/device/wuji_glove.rst (the directory is device, singular). The retargeter got its own page under references/retargeting/, since it needs neither this plugin nor CloudXR.

src/plugins/wuji_glove/wuji_glove_plugin.cpp
wuji-tech-dev: Verified and remapped. Quoting the spec's joint conventions — "The backward (+Z) direction is parallel to the corresponding bone and points away from the finger tip" and "(+Y) ... points away from the palm, i.e. towards the back of the hand" — so +Z runs toward the wrist rather than −Z as in your note; it does not change the outcome, since the remap is a rotation about Z and leaves that axis alone.

The SDK skeleton matches on Z but has the palmar side at +Y on the right hand, so the right hand needs a 180° rotation about Z and the left is identity. The derivation is in the comment on remap_to_openxr_basis, pinned by a static_assert, and checked on hardware for both hands.

examples/retargeting/python/wuji_hand_retargeter_demo.py
wuji-tech-dev: Your reading was right. The retargeter now publishes a hand_valid output, and the demo holds the last valid command while it is low (and sends nothing before the first valid frame).

One deviation: WujiHand2 does not expose realtime_controller, so the Hand 2 path cannot reuse the SDK's LowPass. The demo applies an equivalent single-pole filter at the same 5 Hz cutoff instead.

The retargeter's validity contract is covered by unit tests; the demo-side hold and filter were verified by hand.

src/plugins/wuji_glove/install.sh
wuji-tech-dev: Fixed — extracts to a persistent gitignored directory next to the script, re-runs reuse it, and the trap only removes the download temp dir.

src/plugins/wuji_glove/install.sh
wuji-tech-dev: Fixed — pinned per-arch SHA-256 with a hard failure, following install_manus.sh. Both digests were cross-checked against the release asset metadata, and the mismatch path was tested with a tampered tarball

src/plugins/wuji_glove/wuji_glove_plugin.cpp
wuji-tech-dev: Fixed — the duplicate check runs before subscribing, so the second glove never touches the shared slot; it is logged with both serials and skipped until restart. Verified on hardware with two same-side gloves, and there is a troubleshooting row for it.

examples/retargeting/python/wuji_hand_retargeter_demo.py
wuji-tech-dev: Documented: the format comes from the open-source wuji-retargeting tooling (calibrate_offset.py --dump-pkl), and the docstring and --replay help now state the landmark order, units and frame.

src/retargeters/wuji_hand_retargeter.py
wuji-tech-dev: Done. Added test_wuji_hand_retargeter.py (7 tests: spec shapes, zeros-on-invalid, single-flickering-joint, reset-on-gap and explicit reset) plus test_wuji_mapping_tables.py, which parses both mapping tables from source and asserts they agree with no SDK dependency.

Two deviations worth calling out. The gate is a try-import plus class-level skipif rather than a module-level importorskip, which would leave pytest with zero collected tests and exit code 5 — CTest reports that as a failure. And since there was no wuji extra for the test environment, the SDK was never installed and the test could only ever skip; a [wuji] extra now installs it so the test actually runs, with -DTEST_WUJI_RETARGETER=OFF to keep the vendor package out.

examples/retargeting/python/wuji_hand_retargeter_demo.py
wuji-tech-dev: Fixed — defaults to None, drive mode auto-detects and errors on a conflicting explicit value instead of overriding it. Help text added.

src/retargeters/wuji_hand_retargeter.py
wuji-tech-dev: Fixed — accessed directly, like the sibling retargeters.

src/retargeters/wuji_hand_retargeter.py
wuji-tech-dev: Fixed — the demo's quoted annotations are unquoted now (str | None) and the shebang is added. On the retargeter side hand_joint_names was dropped entirely per your next comment, so the Optional import went with it rather than becoming a union.

src/plugins/wuji_glove/wuji_glove_plugin.cpp
wuji-tech-dev: Fixed — kFramePeriod, kStaleThreshold and kDiscoveryInterval are hoisted next to the existing constants with a comment tying them into one timing budget, the magic 21 is now kNumMediaPipeJoints = std::size(kMpToXr), and m_thread is m_worker_thread. The discovery loop now says "wait kDiscoveryInterval, in kShutdownPollInterval slices" instead of i < 10.

src/plugins/wuji_glove/wuji_glove_plugin.cpp
wuji-tech-dev: Removed the divider instead of documenting it. Its only justification was halving transport load, and it was what put the 16 ms pump (~62.5 Hz) above the data rate in the first place — exactly the re-injection you flagged in your parenthetical. At the native 120 Hz the pump always finds a sample newer than the one it injected last, so identical samples are never re-published under fresh timestamps.

The constants now carry a timing-budget comment giving the native rate and noting that the 200 ms staleness window is ~24 missed frames.

src/retargeters/wuji_hand_retargeter.py
wuji-tech-dev: Both tables now cross-reference each other. On top of that, test_wuji_mapping_tables.py parses both straight from source and asserts they agree entry for entry — no SDK needed, so it runs in the plain dev environment. Consolidating to a single source near HandJointIndex is a follow-up.

src/retargeters/wuji_hand_retargeter.py
wuji-tech-dev: Dropped it, along with the length-20 validation branch — your read was right, the labels are synthetic and nothing overrides them. The comment above _FINGERS now just says the names label the firmware order in the output spec.

src/plugins/wuji_glove/wuji_glove.cpp
wuji-tech-dev: Fixed to == 2. haptikos/main.cpp:16 has the same issue — left alone to keep this PR surgical, happy to file a follow-up.

src/plugins/wuji_glove/wuji_glove_plugin.cpp
wuji-tech-dev: Both applied: a null guard on the scan-failure path, and a second invalidate_hand after wuji_sub_close joins the callback thread.

examples/retargeting/python/wuji_hand_retargeter_demo.py
wuji-tech-dev: Units documented — effort_limit is the per-joint current limit in amperes (the SDK's Kt=1.0 placeholder makes N·m numerically equal to A for now) and mit_params is the MIT impedance (kp, kd) pair, both labelled as vendor-recommended safe defaults.

Drive mode now warns when the connected hand's handedness disagrees with --hand. It reads handedness() after connecting rather than from scan metadata, which does not carry the side; anything other than 'left'/'right' (older firmware, Hand 1's integer encoding) skips the check quietly. Verified on a real Hand 2 both ways.

src/plugins/wuji_glove/CMakeLists.txt
wuji-tech-dev: Took the isolation option: both vendor libraries now install under plugins/wuji_glove/ next to the plugin binary, with INSTALL_RPATH extended to $ORIGIN;$ORIGIN/../../<libdir> so the plugin still resolves the project's own shared libraries. The SDK's license text installs alongside as LICENSE.wuji-sdk-c.

One correction to the premise: the Wuji C SDK is MIT-licensed — the release tarball ships a LICENSE file — so there is no redistribution restriction. We took the isolation option anyway, since it also scopes the soname-resolution surface to this plugin as you noted.

src/core/python/requirements-wuji.txt
wuji-tech-dev: Validated — nlopt 2.11.0 (uploaded 2026-07-17) is the first release shipping manylinux aarch64 wheels; 2.10.0 and earlier have none. On an aarch64 host the install now resolves a prebuilt wheel with no source build.

The note is removed. requirements-wuji.txt was the only mention left in this PR — the rest of the tree was already cleaned up upstream in #798 and #801.

examples/retargeting/python/wuji_hand_retargeter_demo.py
```diff
@@ -0,0 +1,426 @@
...
+
+def run_replay(model: str, hand_side: str, path: str, loop: bool) -> int:
+    print(f"[mode] replay {path} (device-free)\n")
+    retargeter = _build_retargeter(model, hand_side)
+    # Only load trusted, self-generated files: pickle can execute code while loading.
+    with open(path, "rb") as fh:
+        frames = pickle.load(fh)
+    key = f"{hand_side}_fingers"
+    n = 0
+    t0 = time.time()
```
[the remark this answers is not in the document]
wuji-tech-dev: Fixed. --replay now loads through a restricted Unpickler whose find_class allows only the NumPy globals the frame format needs (ndarray, dtype, and _reconstruct under both the numpy<2 and numpy>=2 module paths); anything else raises UnpicklingError. Verified that a normal recording round-trips unchanged and that os.system, builtins.eval and subprocess.run payloads are all rejected.

Keeping the .pkl format matters here — recordings are produced by the upstream wuji-retargeting tooling, so switching to a safer container would make the demo unable to read real takes.

This is the first restricted unpickler in the tree, but it follows the same disposition as the existing YAML loads, all four of which use yaml.safe_load.

jiwenc-nv (approved):

PR #808, opened 2026-07-21
feat(cloudxr): add optional experimental runtime packaging

nv-mhaselton:
## Description

Opt-in packaging for a sibling experimental CloudXR runtime package
(`isaacteleop.cloudxr_exp` with its own `native/`) in the same wheel as the
default `isaacteleop.cloudxr` bundle.

- CMake: `-DENABLE_CLOUDXR_EXP_BUNDLE=ON` (default OFF) downloads/packages the
  exp tarball when available; `ENABLE_CLOUDXR_BUNDLE_CHECK` still owns hard-fail
  vs warn for missing artifacts.
- Runtime: `ISAAC_TELEOP_CLOUDXR_EXP` selects stable vs exp via
  `resolve_cloudxr_runtime_module()` / `get_sdk_path()` (used by
  `CloudXRLauncher`). On Tegra T234, exp is preferred automatically when the
  env is unset; missing exp fails (set `ISAAC_TELEOP_CLOUDXR_EXP=0` to force
  stable). Explicit `ISAAC_TELEOP_CLOUDXR_EXP=1` also fails if exp is not in
  the wheel.
- Docs: CloudXR reference covers runtime selection; build-from-source covers
  the cmake/tarball side (including Orin / TeleViz motivation for exp).

Public CI remains stable-only (flag stays OFF). Layout is ready for a later
dedicated-wheel split if needed.

## Type of change

- [x] Documentation update

## Testing

- Unit: `test_runtime.py` (selection helpers, `resolve_cloudxr_runtime_module`,
  `get_sdk_path`).
- Local dual-bundle smoke:
  - `CXR_RUNTIME_SDK_VERSION=222d1f93 cmake -B build -DENABLE_CLOUDXR_EXP_BUNDLE=ON`
  - Installed `cp311` wheel; `ISAAC_TELEOP_CLOUDXR_EXP=0` →
    `…/cloudxr/native` (StreamSDK `gcomp/rel/gs_04_86`);
    `ISAAC_TELEOP_CLOUDXR_EXP=1` → `…/cloudxr_exp/native` (StreamSDK
    `gcomp/dev`); both report CloudXR `6.3.0` via `runtime_version()` in
    separate processes.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional support for packaging and using an experimental CloudXR runtime bundle.
  * Added environment-based runtime selection between standard and experimental CloudXR libraries.
  * Improved CloudXR SDK download support, including optional experimental releases.

* **Documentation**
  * Added setup and usage instructions for downloading, packaging, and selecting the experimental runtime.

* **Tests**
  * Added coverage for experimental runtime selection and missing-library handling.

### DIFF
--- a/docs/source/getting_started/build_from_source/index.rst
+++ b/docs/source/getting_started/build_from_source/index.rst
@@ -82,10 +82,13 @@ Pre-download CloudXR SDK (Optional)
    :code-file:`download_cloudxr_runtime_sdk.sh <scripts/download_cloudxr_runtime_sdk.sh>`
    script.
 
-Sometimes NVIDIA might share early access CloudXR SDKs with you. In that case, you may get one of
-the two tarballs:
+Sometimes NVIDIA might share early access CloudXR SDKs with you. In that case, you may get
+tarballs such as:
 
 - ``CloudXR-<version-for-runtime-sdk>-Linux-<arch>-sdk.tar.gz`` (CloudXR Runtime SDK)
+- ``CloudXR-exp-<version-for-runtime-sdk>-Linux-<arch>-sdk.tar.gz`` (optional experimental
+  runtime). Needed for Jetson Orin support (for example :doc:`/getting_started/televiz`)
+  until the default runtime covers those platforms.
 - ``nvidia-cloudxr-<version-for-web-sdk>.tgz`` (CloudXR Web SDK)
 
 You can place them in the :code-file:`deps/cloudxr/` directory and update the ``deps/cloudxr/.env``
@@ -96,6 +99,10 @@ like this:
 
    CXR_RUNTIME_SDK_VERSION=<version-for-runtime-sdk>
    CXR_WEB_SDK_VERSION=<version-for-web-sdk>
+
+To package the experimental runtime into the wheel as ``isaacteleop.cloudxr_exp``, configure with
+``-DENABLE_CLOUDXR_EXP_BUNDLE=ON``. Select it at runtime with ``ISAAC_TELEOP_CLOUDXR_EXP``.
+See :ref:`dedicated-cloudxr-runtime`.
 
 2. CMake: Configure and build
 -----------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,9 @@ option(ENABLE_EXPERIMENTAL_WINDOWS_BUILD "Enable experimental Windows build supp
 # TODO: Make this default to ON for Linux builds when the CloudXR runtime SDK is publicly available.
 option(ENABLE_CLOUDXR_BUNDLE_CHECK "Enable CloudXR bundle check" OFF)
 
+# When ON, also package CloudXR Experimental Runtime SDK as isaacteleop.cloudxr_exp.
+option(ENABLE_CLOUDXR_EXP_BUNDLE "Also package CloudXR Experimental Runtime SDK as isaacteleop.cloudxr_exp" OFF)
+
 ## Enforce clang-format on Linux builds by default
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     set(_clang_default ON)
--- a/docs/source/references/cloudxr.rst
+++ b/docs/source/references/cloudxr.rst
@@ -133,7 +133,7 @@ uses the running runtime instead of starting another one:
 Configuration
 -------------
 
-The standalone launcher accepts the same configuration flags as the embedded
+The standalone launcher accepts the same configuration options as the embedded
 one:
 
 - ``--cloudxr-env-config <PATH>`` — a ``KEY=value`` env file of CloudXR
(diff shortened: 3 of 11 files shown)


docs/source/getting_started/build_from_source/index.rst
```diff
@@ -97,6 +98,11 @@ like this:
    CXR_RUNTIME_SDK_VERSION=<version-for-runtime-sdk>
    CXR_WEB_SDK_VERSION=<version-for-web-sdk>
 
+To package the alternate runtime tree into the wheel as ``native_exp/``, configure with
```
sgrizan-nv: Could you help me understand the expected use case for switching between the stable and experimental runtimes within a single installed wheel?

Since CI does not produce the dual-runtime wheel, would it be simpler to produce a separate wheel containing only the experimental runtime when needed? Are there drawbacks to that approach that make bundling both runtimes preferable?

docs/source/getting_started/build_from_source/index.rst
```diff
@@ -82,10 +82,11 @@ Pre-download CloudXR SDK (Optional)
    :code-file:`download_cloudxr_runtime_sdk.sh <scripts/download_cloudxr_runtime_sdk.sh>`
    script.
 
-Sometimes NVIDIA might share early access CloudXR SDKs with you. In that case, you may get one of
-the two tarballs:
+Sometimes NVIDIA might share early access CloudXR SDKs with you. In that case, you may get
+tarballs such as:
 
 - ``CloudXR-<version-for-runtime-sdk>-Linux-<arch>-sdk.tar.gz`` (CloudXR Runtime SDK)
+- ``CloudXR-exp-<version-for-runtime-sdk>-Linux-<arch>-sdk.tar.gz`` (optional alternate runtime tree)
```
jiwenc-nv: @nv-mhaselton maybe we need to include some wording about why is this needed. i.e. Orin support for TeleViz

docs/source/getting_started/build_from_source/index.rst
jiwenc-nv: @nv-mhaselton made some comments during a meeting today that we actually wanna to auto-detect orin, without user need to turn it on explicitly.

sgrizan-nv: Is automatic Orin detection intended to be part of this PR? The current implementation appears to require ISAAC_TELEOP_CLOUDXR_EXP to be set explicitly.

nv-mhaselton: Correct, this PR is only the packaging path and the knobs for future CI once we have an artifact ready to release, with a followup PR for the automation.

nv-mhaselton: As for single vs multiple wheels, it was requested to keep this as a flag on the existing wheel to lessen user friction. For `ISAAC_TELEOP_CLOUDXR_EXP`, the plan is to configure it automatically in the follow-up PR.

If we go with multiple wheels, we could likely update the shared `launcher.py` to detect the platform and switch to the matching installed wheel / namespace (and guide the user if it is missing), for example `cloudxr.runtime.exp` vs `cloudxr.runtime`? It does make the packages smaller by about 40MB and the pattern would be more opt-in which I do prefer.

@jiwenc-nv do you have a preference on the approach?

docs/source/getting_started/build_from_source/index.rst
nv-mhaselton: Sure, added some more clarification, though I expect we'll want to adjust this further when we land the automation component in a follow-up PR.

docs/source/getting_started/build_from_source/index.rst
nv-mhaselton: In the meantime I went ahead and moved `cloudxr/native_exp` to a sibling package (`isaacteleop.cloudxr_exp` with `native/`) in the same wheel. That keeps the single-wheel / flag workflow for now, and lines up better if we later split stable vs exp into dedicated wheels.

sgrizan-nv: Thanks. I see a tradeoff between the two approaches.

With both runtimes in one wheel, installation is simpler, but the same Isaac Teleop wheel may behave differently across platforms because it selects different underlying CloudXR/StreamSDK implementations—not only platform-specific configurations of the same runtime.

One other possible approach would be two complete Isaac Teleop wheel variants built from the same source, each containing one runtime. That would make the selected runtime explicit and behavior easier to reproduce across platforms. On the other hand, maintaining, distributing, installing, and switching between two wheel variants could add significant complexity.

I’m not necessarily advocating for either approach, but I think it would be useful to consider this tradeoff when choosing the packaging model.

jiwenc-nv: Let's try keep the same wheel for Isaac Teleop, the goal is to have the downstream users just pip install and can get the right streamsdk version automatically.

sgrizan-nv (approved):

jiwenc-nv (approved):

sgrizan-nv (approved):

PR #809, opened 2026-07-21
fix(webxr-recorder): clear traces on mode change, dedup joint map, guard unmount

yanziz-nvidia:
Rebases `yanziz/xr-input-recording` onto `origin/main` (one conflict: dropped the now-superseded `store.setFrameRate` call that main moved into `CloudXRComponent` in #0d6e4ba), then applies three targeted fixes:

## Changes

- **TraceVisualization.tsx** — add `useEffect([recorder.mode])` that clears all trace buffers when the recorder mode changes. Prevents dots accumulated during a recording from ghosting under the next live trace or replay.

- **xrInputRecorder.ts** — add `_lastJointMapSession: XRSession | null` sentinel. `_ensureJointMap` now returns immediately if called again for the same session object this frame, eliminating 2–3 redundant joint-map iterations per frame (called from `beginFrame`, `_onGetJointPose`×recording, and `_onGetJointPose`×replay paths).

- **RecorderComponent.tsx** — add `useEffect` with a cleanup that calls `stopRecording()` or `stopReplay()` on unmount. Guarantees prototype patches on `XRFrame.prototype.getPose / getJointPose` and `XRSession.prototype.inputSources` are always removed even when the component is torn down unexpectedly.

## Rebase conflict resolution

`App.tsx` conflict at the `store.setFrameRate(...)` call: kept HEAD (main). The feature branch had added this call to set the frame rate after `enterVR()`; main subsequently moved frame-rate negotiation into `CloudXRComponent` before CloudXR session creation (`fix(webxr): apply effective target frame rate before CloudXR session creation`). Keeping the new approach is correct.


yanziz-nvidia:
Closing — changes belong in #765 on the original branch.

PR #810, opened 2026-07-21
fix(mcap-examples): added launcher arguments for 1.4+local

jsepulveda-nvidia:
Installed isaacteleop 1.3.9rc1 lacks add_launcher_arguments() and launch_context() added in 1.4+local. Replace CloudXRLauncher usage with contextlib.nullcontext so all six mcap_record_replay example scripts connect to an already-running CloudXR instance instead of trying to launch a new one.

This initial diff also tries to  bypass CloudXRLauncher for existing CloudXR instances. However I DON"T want this to happen all the time. I only want this to happen if the switch is specified: 
  --no-launch-cloudxr-runtime
I plan to repair this before checking in. If you see the hook to the launcher in any of the files feel free to mark it in error. I will get around to fixing it ASAP.

## Description

Fixes #(issue)

## Type of change

- [ ] Documentation update

## Checklist

## Summary by CodeRabbit

* **Changes**
  * Updated live visualization and recording examples to connect to an already-running CloudXR instance instead of starting one automatically.
  * Removed CloudXR-specific startup options from these example commands.
  * Existing controller, hand, full-body visualization, and MCAP recording workflows remain available.
  * Added connection status messaging when attaching to CloudXR.

### DIFF
--- a/examples/mcap_record_replay/python/common.py
+++ b/examples/mcap_record_replay/python/common.py
@@ -12,9 +12,14 @@ Viz classes (used by replay_* and live_* scripts): ``HandViz``,
 Rendering helpers: ``HandJoints``, ``HAND_BONES``, ``BODY_BONES``.
 """
 
+import argparse
+import contextlib
+import os
+
 import numpy as np
 import viser
 
+from isaacteleop.cloudxr import CloudXRLauncher
 from isaacteleop.retargeting_engine.deviceio_source_nodes import (
     ControllersSource,
     FullBodySource,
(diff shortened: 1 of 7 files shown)


jiwenc-nv:
thanks for actively thinking about the healthiness of the examples we have.  But I don't think we need to support runnings 1.4 examples on top of 1.3 wheel: https://github.com/NVIDIA/IsaacTeleop/tree/release/1.3.x/examples/mcap_record_replay

jsepulveda-nvidia:
Yes, I don't think this needs to be checked in anymore if our users can just use the 1.4 wheel. I will close the review but leave the branch open just for organizational purposes over here.

PR #816, opened 2026-07-23
fix(cloudxr): tear down on signal and join main on Orin

nv-mhaselton:
## Description

Signalled shutdown of an embedding process previously left the CloudXR runtime running. The runtime is spawned with `start_new_session=True`, and Python does not run `atexit` handlers on SIGTERM/SIGINT, so the next start could fail with `PORT_UNAVAILABLE`. This change installs SIGTERM/SIGINT handlers (and `atexit`) after `Popen` so the session-scoped runtime is torn down cleanly, then invokes any previously installed handler so the embedding app keeps its normal shutdown behavior. Teardown keeps those handlers installed until stop finishes, guards re-entrant `stop()`, and restores a prior handler only when the launcher wrapper is still installed.

Separately, joining `nv_cxr_service_join` on a worker thread can abort on Orin with `_PyGILState_NoteThreadState: Couldn't create autoTSSkey mapping` after V4L2 encoder teardown. Main-thread join is selected automatically on Orin. Because Python handlers cannot run while that join blocks, SIGINT/SIGTERM are handled via `sigwait` so `nv_cxr_service_stop` can still run. Override with `ISAAC_TELEOP_CLOUDXR_JOIN_MAIN=0|1` (unset keeps the platform default). Launcher overrides (`ISAAC_TELEOP_CLOUDXR_EXP`, `ISAAC_TELEOP_CLOUDXR_JOIN_MAIN`) are documented as Orin auto / override-only in the CloudXR reference, with a short note under the quick-start env table.

## Type of change

- [x] Documentation update

## Testing

- Unit: `ctest --test-dir build --output-on-failure -R 'cloudxr'` (launcher + runtime, including join-main sigwait coverage).
- Smoke on Orin: join auto-selected; start → SIGTERM → restart with no orphans / no `PORT_UNAVAILABLE`.
- Smoke on Thor: join not auto-selected; same teardown/restart path ok.

## Checklist


docs/source/references/cloudxr.rst
```diff
@@ -148,6 +148,10 @@ one:
   platforms. On Tegra T234 it is selected automatically; if the package is
   missing the launcher fails. Set ``ISAAC_TELEOP_CLOUDXR_EXP=0`` to force the
   stable runtime.
+- ``ISAAC_TELEOP_CLOUDXR_JOIN_MAIN=1`` — join the CloudXR service on the main
```
jiwenc-nv: From a quick glance at this section of the doc, it's not very clear where to put `ISAAC_TELEOP_CLOUDXR_EXP` and `ISAAC_TELEOP_CLOUDXR_EXP`.  I think we want to consolidate configurations into the env file?

jiwenc-nv: @nv-mhaselton can you add the new configs to https://nvidia.github.io/IsaacTeleop/main/getting_started/quick_start.html#id1

nv-mhaselton: Added some clarification as to their purpose as they are mostly escape hatches, and documented them as a table in the cloudxr launcher reference. 🤞

sgrizan-nv (approved):

aristarkhovNV:
Is it possible to converge on the same Main-thread join code path on all platforms?

nv-mhaselton:
> Is it possible to converge on the same Main-thread join code path on all platforms?

The thought was to keep this workaround Orin-specific. On other platforms the worker-thread join is useful so the main thread can still run Python signal handlers (Ctrl+C) while blocked in native `nv_cxr_service_join`. Converging everyone onto main-thread join would give up that behavior and rely on launcher process teardown everywhere. I figured we'd rather keep the worker-thread join as the long-term path and move Orin over to it once the abort is gone.

On Orin, starting any new Python thread after the native service is up aborts with `_PyGILState_NoteThreadState: Couldn't create autoTSSkey mapping` (including a join worker and a `sigwait` helper we tried), so join stays on the main thread there.

I've added more clarification around the workaround in `runtime.py`. Happy to revisit a single (worker-thread) path once the Orin abort is gone.

aristarkhovNV (approved):

PR #819, opened 2026-07-24
[TeleViz] Opt-in native OpenXR quad layers for QuadLayer

farbod-nv:
## Description
  
  Adds an opt-in native OpenXR quad path to Televiz: `QuadLayer::Config::use_openxr_quad_layer`.
  
  When set on a layer in a `kXr` session, the quad is submitted to the runtime as a native
  `XrCompositionLayerQuad` — the backend copies the CUDA-fed image into a per-layer, lazily-created
  color swapchain and lets the runtime place and sample the quad directly — instead of compositing it
  into the shared render target. Stereo layers emit one quad per eye via `eyeVisibility`, with the
  per-eye baseline offset matching the compositor path's convention. Frames whose visible layers are
  all native quads drop the shared projection layer entirely, so the runtime sees a quad-only
  submission and can engage its quad fast path / client-reconstructed streaming. Window/offscreen
  modes ignore the flag and keep the compositor draw path.
  
  Trade-off (documented on the config field): a native quad carries no depth — OpenXR quad layers
  have none — so it composites in submission order rather than z-testing against `ProjectionLayer`
  content, and `generate_mipmaps` is unused in native mode. In non-opaque (passthrough) blend modes
  the source-alpha layer flag keeps quads compositing correctly but excludes them from the runtime's
  client-reconstructed optimization; opaque VR sessions get the fast path.
  
  Implementation notes:
  
  - `layer_base`/`display_backend`: `NativeQuadView` descriptor plus `is_native_quad()` /
    `acquire_native_quad()` and `supports_native_quad()` / `record_native_quads()` extension points
  - `quad_layer`: extracted `promote_slot()` shared by the render-pass and native-quad consumers;
    `cuda_done_writing` waits at TRANSFER stage in native mode (the backend copy is the first GPU read)
  - `xr_backend`: per-quad swapchain cache keyed by layer identity (exception-safe partial-creation
    cleanup), `projection_active_` gating in `end_frame`, abort/release/destroy coverage
  - `viz_compositor`: partitions native vs composited layers; skips the shared render pass on
    quad-only frames
  - Python: exposes the flag on `QuadLayerConfig` (stub regenerated); `camera_viz` gains `--native-quad`
  
  ## Type of change 
  
  - [ ] Documentation update

  ## Testing

  Linux x86_64, 2× RTX 6000 Ada, CUDA 13.x.

  - **Unit/GPU (Release)**: full viz suites pass — `viz_core_tests` (186 assertions),
    `viz_layers_tests` (351), `viz_session_tests` (386). New `[native]` tests cover the config
    default, the native gate (`is_native_quad()` false without a kXr session), acquire preconditions
    (pre-publish leniency, placement required once content is published, out-of-range in-flight
    slot), and the offscreen fallback (flag set → still composites through `record()`, pixel-verified
    via readback).
  - **ASAN+UBSAN**: same suites pass with zero reports (`ASAN_OPTIONS=protect_shadow_gap=0` for CUDA).
  - **Python**: all 4 viz binding test files pass; flag round-trips through `QuadLayerConfig`.
  - **Live XR** against a local CloudXR runtime 6.2.0 (no client attached): mono native, stereo
    native, mixed native+composited (projection + quad in one `xrEndFrame`), and per-frame visibility
    toggling (projection layer drops and re-engages mid-session) — 300+ frames total, no XR errors,
    ~33 fps runtime pacing. Visual placement/disparity and streaming-fast-path engagement with a
    headset client remain to be verified manually.
  - `SKIP=check-copyright-year pre-commit run` passes on all changed files; clang-format clean.

  ## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional native OpenXR quad-layer rendering for XR sessions.
  * Added `--native-quad` to the camera visualization example.
  * Native quad layers now support stereo placement, sizing, and per-eye imagery.
  * Quad-only XR frames can be presented without an intermediate projection layer.
  * Non-XR sessions continue using the existing compositor path.

* **Bug Fixes**
  * Improved frame synchronization for native quad rendering.

* **Tests**
  * Added coverage for configuration, prerequisites, fallback behavior, and quad-only presentation.

### DIFF
--- a/examples/camera_viz/camera_viz.py
+++ b/examples/camera_viz/camera_viz.py
@@ -234,6 +234,15 @@ def main(argv: Optional[list[str]] = None) -> int:
         help="Override display.mode from the config "
         "(default: the config's value, or xr when the config omits it).",
     )
+    parser.add_argument(
+        "--native-quad",
+        action="store_true",
+        help="Submit each camera as a native OpenXR quad layer "
+        "(XrCompositionLayerQuad) instead of the built-in compositor. "
+        "kXr only; ignored in window mode. When every layer is a native "
+        "quad the projection layer is dropped, letting the runtime engage "
+        "its quad fast path / client-reconstructed streaming.",
+    )
     args = parser.parse_args(argv)
 
     with open(args.config) as f:
@@ -274,6 +283,9 @@ def main(argv: Optional[list[str]] = None) -> int:
         if entry.stereo:
             layer_cfg.stereo = True
             layer_cfg.stereo_baseline_mm = entry.stereo_baseline_mm
+        # Native OpenXR quad path (kXr only; a no-op fallback in window mode).
+        # Requires a placement, which the placement strategy applies below.
+        layer_cfg.use_openxr_quad_layer = args.native_quad
         layers.append(session.add_quad_layer(layer_cfg))
         strategies.append(entry.placement)
 
--- a/src/viz/layers/cpp/inc/viz/layers/quad_layer.hpp
+++ b/src/viz/layers/cpp/inc/viz/layers/quad_layer.hpp
@@ -119,6 +119,23 @@ public:
         // false or outside kXr. mm-scale chosen because typical real-
         // world IPDs and camera baselines are 50–80 mm.
         float stereo_baseline_mm = 0.0f;
+
+        // Submit this quad as a native OpenXR quad layer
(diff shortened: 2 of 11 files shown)


tengw-nv (approved):

nv-jakob:
What is the default here? Before 6.3 we should keep the default to be composited in the renderer here.

PR #833, opened 2026-07-28
Support argus camera for camera_viz

tengw-nv:
## Description

Implement argus camera driver support in the camera_viz example

Fixes #(issue)

## Type of change

- [ ] Documentation update

## Testing

Tested with camera that supports argus driver

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Jetson Argus camera support with mono and stereo capture.
  * Added CUDA-accelerated YUV-to-RGBA frame conversion.
  * Added SHW5G stereo camera configuration.
  * Added optional `--with-argus` setup support and improved setup guidance.
  * Added stereo emission mode that waits for fresh frames from both eyes.

* **Bug Fixes**
  * Prevented partial startup when one stereo camera fails to start.
  * Added clearer guidance when the native camera component is unavailable.

* **Tests**
  * Added coverage for synchronized stereo frame emission.

### DIFF
--- a/examples/camera_viz/argus/yuv_to_rgba.cu
+++ b/examples/camera_viz/argus/yuv_to_rgba.cu
@@ -1,0 +1,214 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Argus YUV420 -> RGBA8 conversion for CUDA EGLStream consumers.
+
+#include "yuv_to_rgba.cuh"
+
+namespace
+{
+
+__device__ __forceinline__ unsigned char clamp_u8(float v)
+{
+    return static_cast<unsigned char>(v < 0.f ? 0.f : (v > 255.f ? 255.f : v));
+}
+
+__device__ __forceinline__ void yuv_to_rgb(
+    int Y, int Cb, int Cr, bool full_range, unsigned char& r, unsigned char& g, unsigned char& b)
+{
+    float R, G, B;
+    if (full_range)
+    {
+        const float yf = static_cast<float>(Y);
+        const float u = static_cast<float>(Cb) - 128.f;
+        const float v = static_cast<float>(Cr) - 128.f;
+        R = yf + 1.402f * v;
+        G = yf - 0.344136f * u - 0.714136f * v;
+        B = yf + 1.772f * u;
+    }
+    else
+    {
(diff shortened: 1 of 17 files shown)


examples/camera_viz/argus/yuv_to_rgba.cu
```diff
@@ -0,0 +1,214 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Argus YUV420 -> RGBA8 conversion for CUDA EGLStream consumers.
+
+#include "yuv_to_rgba.cuh"
+
```
jiwenc-nv, line 7: it's a bit funny what we till need to hand roll rgh/yuv conversion... maybe we can move this to viz lib in the future.

jiwenc-nv (approved):

farbod-nv:
```
  1. deploy deletes the Argus module and never rebuilds it — camera_viz.sh:283

  cmd_deploy hardcodes bash scripts/_install_deps.sh --sender-only --jetson with no --with-argus, and there's no flag to
  inject one. Worse, rsync_to_remote (camera_viz.sh:110-113) excludes argus/build/ but the built artifact lands in the source
  dir (argus/CMakeLists.txt:114 LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}), and argus/_camera_viz_argus*.so is not
  excluded. So rsync --delete removes a working .so from the Jetson on every deploy, and the deps step doesn't build it back.
  The Jetson deploy path is exactly the Argus target. Needs --exclude='argus/*.so' plus a way to pass --with-argus through
  deploy.

  2. Missing sync between EGL acquire and the conversion kernel — argus_camera.cpp:672

  AcquiredFrame::stream is default-initialized to nullptr, so cuEGLStreamConsumerAcquireFrame(..., &out.stream, ...) orders
  the frame's availability/migration on the legacy default stream. The kernel then launches on convert_stream_, created with
  CU_STREAM_NON_BLOCKING (:349) — which by definition does not implicitly synchronize with the legacy default stream. Set
  out.stream = convert_stream_ before the acquire (and pass the same stream to release), or drop CU_STREAM_NON_BLOCKING.
```
^ just an agent review. ignore if already addressed/irrelevent.

Let's also add argus to the public docs too. Otherwise LGTM

farbod-nv (approved):

PR #838, opened 2026-07-29
docs: derive install version specifiers from VERSION

jiwenc-nv:
## Description

The quick start hardcoded `pip install 'isaacteleop[cloudxr,retargeters]~=1.0.0' --extra-index-url https://pypi.nvidia.com`, so every published docs version — `main` included — sends readers to the 1.0 series no matter what the page around it describes.

The install step now offers three commands:

| Intent | Command |
| --- | --- |
| Latest stable | `pip install 'isaacteleop[cloudxr,retargeters]~=1.0'` |
| Latest pre-release | `uv pip install 'isaacteleop[cloudxr,retargeters]~=1.0' --extra-index-url https://pypi.nvidia.com --prerelease=allow` |
| The series this page documents | `uv pip install 'isaacteleop[cloudxr,retargeters]~=1.5.0' --extra-index-url https://pypi.nvidia.com --prerelease=allow` |

The third one is not written by hand. `conf.py` reads the `VERSION` file it already reads for `version`/`release`, turns `MAJOR.MINOR.x` into `~=MAJOR.MINOR.0`, and exposes it as the `|pip_version_pin|` substitution. Each sphinx-multiversion build therefore advertises its own series: `main` renders `~=1.5.0` today, and every release branch and tag cut from this commit onward renders its own (`release/1.6.x` -> `~=1.6.0`) with no follow-up edit at release time. Already-published versions are unaffected — sphinx-multiversion builds each ref from that ref's own `docs/`, which is exactly why the site keeps serving old docs for old versions.

Two things that came out of that:

- `docs/source/getting_started/lerobot/data_collection_real.rst` carried its own hardcoded `~=1.3.131`. It now uses the same substitution — same defect, and leaving a stale literal pin next to the new mechanism would just recreate the issue on the next branch cut. Shout if you would rather keep that page pinned to a known-good build.
- Substitutions only expand in `parsed-literal`, which Sphinx renders as a bare `pre.literal-block` rather than `div.highlight pre` — sphinx-copybutton's default selector, so the version-pinned command would have silently lost its copy button. `copybutton_selector` is widened to cover both. Across a full build the only pages containing `pre.literal-block` are the two touched here plus the two pre-existing blocks in `references/oob_teleop_control.rst`, which gain a copy button they should have had already.

Fixes #837

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
[...]

PR #839, opened 2026-07-29
fix: align CloudXR quick start requirements

arussell-nvidia:
## Summary

- clarify single-runtime, certificate-acceptance, and mode-specific `adb`/`coturn` requirements
- declare the MuJoCo CloudXR example's WebSockets dependency
- document Windows and sparse-checkout prerequisites for running the repository hook set

## Validation


arussell-nvidia:
Addressed the package-installation review in `c3e30a9c`: the normal Wi-Fi path now has no runtime-package command, `--setup-oob` installs only `adb`, and `--setup-oob --usb-local` installs `adb` plus `coturn`. I also merged current `main` normally, preserving the version-aware install flow from #838 without force-pushing. The full pre-commit hook set passes on the merged result.

PR #850, opened 2026-07-30
fix(web): upgrade IWER to 2.3.0, force-install over unsupported native XR, and prevent duplicate installs

yanziz-nvidia:
## Description

Updates the Isaac Teleop web client IWER dependency from 2.2.1 to 2.3.0 (DevUI 2.2.0 → 2.3.0) and fixes two issues:

1. **Force-install over unsupported native XR** — IWER 2.3 preserves an existing `navigator.xr` by default, even when it reports no immersive support. This PR passes `forceInstall: true` on the fallback path so the emulated runtime replaces the unusable native one.

2. **Prevent duplicate device installs from concurrent callers** — React's `useEffect` (which calls `loadIWERIfNeeded`) can be invoked concurrently in development (Strict Mode double-invocation). Without a guard, two overlapping calls each install a separate IWER device, leaving `window.xrDevice` pointing at an inactive one. A single-flight `pendingIWERLoad` promise ensures all concurrent callers share one installation.

Also updates the CDN SHA-384 integrity hashes for both IWER and DevUI.

## Type of change

- [ ] Documentation update

## Testing

- `npm test -- --runInBand` — 63 tests passed (including regression tests for the force-install path and the concurrent-callers guard)
- `npm run build`

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -100,9 +100,6 @@ pre-commit install --hook-type commit-msg
   SKIP=check-copyright-year pre-commit run --all-files
   ```
 
-- In sandboxed macOS environments, REUSE multiprocessing can fail while reading
-  `SC_SEM_NSEMS_MAX`; rerun the full hook set outside the sandbox before treating
-  that `PermissionError` as a source failure.
 - **REUSE:** files covered by the REUSE hook need **`SPDX-FileCopyrightText`** and **`SPDX-License-Identifier`** in the form the repo already uses (for example the HTML comment block at the top of `README.md` also applies to **`AGENTS.md`** and similar docs).
 - **C++ formatting is enforced by CI, not pre-commit.** The hook set runs `ruff` for Python but does **not** run `clang-format`; CI (`build-ubuntu.yml`) installs **`clang-format-14`** and rejects unformatted C++ as `-Wclang-format-violations`. Before pushing, format touched C++ with the system `clang-format` (match CI's version 14) and verify:
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -100,6 +100,9 @@ pre-commit install --hook-type commit-msg
   SKIP=check-copyright-year pre-commit run --all-files
   ```
 
+- In sandboxed macOS environments, REUSE multiprocessing can fail while reading
+  `SC_SEM_NSEMS_MAX`; rerun the full hook set outside the sandbox before treating
+  that `PermissionError` as a source failure.
 - **REUSE:** files covered by the REUSE hook need **`SPDX-FileCopyrightText`** and **`SPDX-License-Identifier`** in the form the repo already uses (for example the HTML comment block at the top of `README.md` also applies to **`AGENTS.md`** and similar docs).
 - **C++ formatting is enforced by CI, not pre-commit.** The hook set runs `ruff` for Python but does **not** run `clang-format`; CI (`build-ubuntu.yml`) installs **`clang-format-14`** and rejects unformatted C++ as `-Wclang-format-violations`. Before pushing, format touched C++ with the system `clang-format` (match CI's version 14) and verify:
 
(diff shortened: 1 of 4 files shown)


jiwenc-nv (approved):

yanziz-nvidia (commented):
Reviewed by yanziz-reviewer-bot

## Summary

Bumps the web client's IWER pin from 2.2.1 to 2.3.0 (CDN script tags + SRI hashes, DevUI constant), passes `forceInstall: true` on the fallback `installRuntime` so IWER replaces a native `navigator.xr` that reports no immersive support, and adds an in-flight single-flight guard plus a new jsdom test file. Dependency change is one devDependency (`iwer` `^2.2.1` → `^2.3.0`); `package-lock.json` is gitignored repo-wide so no lockfile update applies, and there is no `DEPENDENCIES.md` in this repo. CI: `Build Teleop Web App`, `pre-commit`, `Check repo`, CodeQL, and both DCO checks pass; the `build-ubuntu` matrix and `pip-editable-install` are still pending (C++/Python jobs, unrelated to this diff).

> Legend: 🚫 Blocker · 💡 Suggestion · 🔍 Nit

|     | Finding |
|-----|---------|
| 🚫 | None |
| 💡 | [`deps/cloudxr/webxr_client/helpers/LoadIWER.ts:35-45`](https://github.com/NVIDIA/IsaacTeleop/blob/40069a784618a4e82915fa673a645963d9265d77/deps/cloudxr/webxr_client/helpers/LoadIWER.ts#L35-L45) — **The guard dedups only in-flight calls, so a caller arriving after the load resolves re-runs `loadIWER` and gets `iwerLoaded: false`.** React 19 StrictMode remounts sequentially; if the scripts are already cached the first call can settle before the second effect fires. Not harmful (the now-emulated `navigator.xr` reports support, so no second install), but the returned `iwerLoaded` flips to `false` for the same page state. |
| 💡 | [`deps/cloudxr/webxr_client/helpers/LoadIWER.test.ts:33`](https://github.com/NVIDIA/IsaacTeleop/blob/40069a784618a4e82915fa673a645963d9265d77/deps/cloudxr/webxr_client/helpers/LoadIWER.test.ts#L33) — **Both tests exercise the fallback path; nothing covers the native-supported path.** `forceInstall: true` is destructive by design, and the only thing stopping it from clobbering a working native runtime is the `if (!supportsImmersive)` guard at [`LoadIWER.ts:59`](https://github.com/NVIDIA/IsaacTeleop/blob/40069a784618a4e82915fa673a645963d9265d77/deps/cloudxr/webxr_client/helpers/LoadIWER.ts#L59). |
| 🔍 | [`deps/cloudxr/webxr_client/helpers/LoadIWER.ts:27`](https://github.com/NVIDIA/IsaacTeleop/blob/40069a784618a4e82915fa673a645963d9265d77/deps/cloudxr/webxr_client/helpers/LoadIWER.ts#L27) — **`@iwer/devui` is CDN-loaded but not declared in `package.json`, so its version is tracked only by this constant.** `iwer` gets dependency-bot and audit coverage; DevUI does not, and the two must stay version-compatible. |

## Actionables (for bots — copy-paste-ready for AI)
> Fix if it makes sense in context — these are agent-generated suggestions, not human-vetted obligations. Skip anything that's wrong, already addressed, or not worth the churn.

- `deps/cloudxr/webxr_client/helpers/LoadIWER.ts:35-45` — Cache the resolved `IWERLoadResult` when `iwerLoaded` is true instead of always clearing `pendingIWERLoad` in `finally`, so post-resolution callers get a stable result; keep clearing on failure so a failed load can be retried.
- `deps/cloudxr/webxr_client/helpers/LoadIWER.test.ts:33` — Add a test where `isSessionSupported` resolves true, asserting no script is appended, `installRuntime` is never called, and the result is `{ supportsImmersive: true, iwerLoaded: false }`.
- `deps/cloudxr/webxr_client/helpers/LoadIWER.ts:27` — Declare `@iwer/devui` in `devDependencies` alongside `iwer` (or add a comment noting the two CDN versions must be bumped together) so the DevUI version is tracked.

PR #853, opened 2026-07-30
feat(deviceio): generate passthrough trackers from a TOML manifest

nv-jakob:
Six trackers were pure FlatBuffer passthrough over the OpenXR tensor extensions, and after normalizing type names their live and replay impls were the same file. The duplication had already caused drift: the se3 replay impl logged a missing-data gap once behind a warned_no_data_ flag while the oglo one wrote to std::cerr every frame -- same intent, only one copy got the fix. Each new tracker also cost ~24 files of mechanical code plus edits to six shared registration points.

Declare passthrough trackers in deviceio_trackers/trackers.toml instead, with %placeholder% defaults in defaults.toml, and generate the facade, base interface, and live/replay impls at CMake configure time via codegen/generate_trackers.py. Registration points (factory dispatch rows and methods, MCAP recording traits, pybind blocks, Python __all__) are emitted as .inc fragments that the hand-written files include, so hand-written trackers keep their rows. Adding a passthrough tracker is now a .fbs schema plus ~10 lines of TOML.

Migrate se3_tracker, oglo_tactile, joint_state, generic_3axis_pedal and haptic_command_reader, and delete their hand-written stacks. All generated replay impls adopt the warn-once-per-gap logging, so oglo's per-frame stderr spam is gone -- a deliberate behavior change, not an accident.

Add a typed producer direction: direction = "out" generates a <Table>PushTracker wrapping SchemaPusher, so PushTensorHapticDevice now pushes a typed HapticCommand instead of hand-rolling pack_haptic_command over the untyped TensorPushTracker, which stays as the bytes escape hatch.

head, hand, controller, full_body, message_channel and FrameMetadataTrackerOak remain hand-written; the .fbs schemas and their pybind bindings are out of scope. Docs and the AGENTS.md files for the affected packages are updated to describe the manifest workflow.

## Description

Fixes #(issue)

## Type of change

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added manifest-driven generation for tracker facades, live/replay implementations, factory registration, recording traits, and Python bindings.
  * Added configurable inbound trackers and outbound haptic command push trackers.
  * Generated tracker classes are available through standard Python device-tracker exports.

* **Documentation**
  * Added guidance for configuring, regenerating, troubleshooting, and inspecting generated trackers.
  * Documented supported tracker types and integration boundaries.

* **Tests**
  * Added coverage for manifest validation, defaults, generation, stale-file cleanup, and haptic command recording.


src/core/codegen/generate_trackers.py
```diff
@@ -0,0 +1,956 @@
...
+
+def _copyright_lines() -> Iterator[str]:
+    yield from COPYRIGHT.splitlines()
+
+
+def _join_lines(lines: Iterator[str]) -> str:
+    return "\n".join(lines) + "\n"
+
+
+def _gen_single_in_deviceio_base(ctx: TrackerGenContext) -> Iterator[str]:
```
jiwenc-nv: as discussed during the workshop today, let's replace all those python yields with something like:

https://github.com/NVIDIA/IsaacTeleop/blob/main/src/core/python/pyproject.toml.in

docs/source/references/generated_trackers.rst
```diff
@@ -0,0 +1,145 @@
...
+
+**Vendored shape (Noitom).** ``LiveFullBodyTrackerNoitomImpl`` reads
+``SchemaTracker<FullBodyPoseRecord, FullBodyPose>``, so it is passthrough by mechanism, but it is a
+second *vendor* of the hand-written ``FullBodyTracker`` rather than its own tracker type. Generating
+it would need a shape that emits **only** a live impl and a vendor-keyed dispatch row — no facade,
+base interface, recording traits, replay half, or pybind block, since it shares all of those with the
+PICO vendor. Its ``collection_id`` and ``max_flatbuffer_size`` also arrive through
+``TrackerVendor::params`` at runtime instead of being fixed at generation time.
+
+**Schema pybind bindings.** The field-by-field ``src/core/schema/python/*_bindings.h`` files are
```
jiwenc-nv, line 149: @aristarkhovNV, I think you can follow up here for using flatbuffer objs rather than the manually coded bindings as the next step.

jiwenc-nv (changes requested):
let's use template files, rather than writing c++ code in python strings :)

src/core/codegen/generate_trackers.py
nv-jakob: Replaced most of them with `*.[cpp|hpp].template` files.

src/core/codegen/generate_trackers.py
```diff
@@ -0,0 +1,956 @@
...
+    )
+    yield "{"
+    if ctx.direction == "in" and ctx.record:
+        yield f"    std::unique_ptr<{ctx.mcap_channels_type}> channels;"
+        yield "    if (should_record(tracker)) {"
+        yield f"        channels = {ctx.live_impl}::create_mcap_channels(*writer_, get_name(tracker));"
+        yield "    }"
+        yield f"    return std::make_unique<{ctx.live_impl}>(handles_, tracker, std::move(channels));"
+    else:
+        yield f"    return std::make_unique<{ctx.live_impl}>(handles_, tracker);"
```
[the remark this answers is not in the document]
nv-jakob: Removed `multi_endpoint` so resolving.

src/core/deviceio_trackers/trackers.toml
```diff
@@ -0,0 +1,37 @@
...
+table = "Generic3AxisPedalOutput"
+schema = "pedals"
+channel = "pedals"
+traits = "PedalRecordingTraits"
+max_flatbuffer_size = 256
+python_accessor = "get_pedal_data"
+
+[[tracker]]
+name = "haptic_command"
+direction = "out"
```
jiwenc-nv: is that right to think that all other tracker to be "in"

src/core/codegen/templates/push/facade.hpp.template
```diff
@@ -0,0 +1,44 @@
...
+#pragma once
+#include <deviceio_base/@BASE_HEADER@.hpp>
+#include <schema/@SCHEMA@_generated.h>
+
+#include <cstddef>
+#include <string>
+#include <string_view>
+namespace core
+{
+class @CLASS@ : public ITracker
```
jiwenc-nv, line 13: why don't we just call this `IPusher`

jiwenc-nv: definitely in a follow up

nv-jakob: You mean instead `ITracker`, they are still trackers because everything has to be a tracker. Changing that would be a much larger change. :)

src/core/deviceio_trackers/trackers.toml
nv-jakob: Changed to pull/push.

cmake/ClangFormat.cmake
```diff
@@ -36,6 +36,7 @@ set(_cf_patterns
 )
 
 set(_cf_exclude_patterns
+  ".*\\.template$"
```
jiwenc-nv: just curious: why we need this?

docs/source/device/add_device.rst
```diff
@@ -107,11 +106,55 @@ Reference implementation: :code-dir:`src/plugins/generic_3axis_pedal`. The plugi
...
+     --emit-cmake /tmp/isaac-teleop-trackers.cmake \\
+     --print-resolved
+
+(``--out-dir`` and ``--emit-cmake`` are required by the CLI even when ``--print-resolved``
+skips writing generated sources; temporary paths are fine.)
+
+Hand-written trackers
+~~~~~~~~~~~~~~~~~~~~~
+
+Non-passthrough trackers (OpenXR ``xrLocate*``, multi-stream camera metadata, opaque message
```
jiwenc-nv: iirc, `camera metadata` can be a pure passthrough as well.   I saw you removed [frame_metadata_tracker_oak_base.hpp](https://github.com/NVIDIA/IsaacTeleop/pull/853/changes#diff-ae25ef2c2bad49f25baa16e299a77838fe46557d278274f2e5ebd31814628fea) already.

docs/source/device/add_device.rst
```diff
@@ -107,11 +106,55 @@ Reference implementation: :code-dir:`src/plugins/generic_3axis_pedal`. The plugi
 Step 3: Implement a tracker
 ----------------------------
 
+.. _passthrough-tracker-manifest:
+
+Passthrough trackers (manifest)
```
jiwenc-nv: tbh, "Passthrough" sounds a bit weird, as it sort of suggesting it's a no-op.

I think Schema-based trackers makes a bit more sense.

The benefit of this naming is that it reuses the term we already have. esp the generated code literally uses `SchemaTracker` classes.  No need to coin a new name "Passthrough" here, where we will now argue about that it means.

Since this is doc-only change, I hope we can land it quickly.

jiwenc-nv (approved):

docs/source/device/add_device.rst
nv-jakob: Done

docs/source/device/add_device.rst
nv-jakob: Done

cmake/ClangFormat.cmake
nv-jakob: Clang format would be run on the templates and then the `@KEY@` would get formatted as `@KEY @` , which would then break when running the template code.

nv-jakob: Actually, it was only needed when the files ended on `.cpp` / `.hpp` but with the `.template` suffix it's not needed anymore. Removed it.

docs/source/references/generated_trackers.rst
nv-jakob: Resolving so I can merge.

PR #844, opened 2026-07-29
feat(client): show host workstation notices in the teleop UI

rwiltz:
## Description
The client already ran a receive loop on the teleop_command message channel, but the handler was an empty stub, so nothing the host sent was ever surfaced. Dispatch received messages by type and render a system_notice as a dismissible warning banner.

Isaac Lab uses this to report that its workstation is below the recommended spec for teleoperation (CPU single-thread throughput, frequency governor, GPU, memory). Without it the operator sees only a low frame rate and no explanation, and the terminal warning is not visible while wearing a headset.

The banner renders in XR via uikit and is mirrored to the 2D status box so it is also visible when testing from a desktop browser. Text is carried in signals so updates bypass React re-renders, matching the existing performance metrics. Unknown message types are logged and ignored, keeping this client compatible with hosts that send message kinds it does not know about.

Also set white-space: pre-line on the 2D status box. It is populated with textContent and existing callers already separate entries with newlines, which were being collapsed onto one line.

Screenshot from 2026-07-29 17-19-55

Fixes #(issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read and understood the [contribution guidelines](../CONTRIBUTING.md)
- [x] I have run the linter and formatter with `SKIP=check-copyright-year pre-commit run --all-files`
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix/feature works (or explained why not)
- [x] I have signed off all my commits (`git commit -s`) per the [DCO](../CONTRIBUTING.md#signing-your-work)

## Summary by CodeRabbit

* **New Features**
  * Added system notices delivered from the server.
  * Displays notice titles, details, requirements, and documentation links in the 2D and in-headset interfaces.
  * Added a dismiss button and automatic dismissal after 20 seconds.
  * Clears active notices when disconnecting from XR.

* **Bug Fixes**
  * Invalid or unsupported server notices are safely ignored.
  * Notices no longer persist across XR connections.

* **Tests**
  * Added coverage for notice validation and formatted display content.

PR #860, opened 2026-07-30
fix(client): repair status box regressions from the workstation notice

rwiltz:
## Description

  Two regressions in the shared 2D status box, both introduced by #844.

  **CloudXR errors were erased on disconnect.** The box is a single shared slot,
  but `dismissSystemNotice()` cleared it unconditionally. On a failed session
  CloudXR reports the error via `onError` → `showError()`, then immediately calls
  `onExitImmersiveXR` → `handleDisconnect()` → `dismissSystemNotice()`, which
  blanked the error that had just been written. Because that error is reported
  imperatively rather than through React state, nothing re-rendered it and the
  operator saw a disconnect with no explanation.

  The notice now records the exact text it wrote and retracts it only while the 
  box still shows it, via a new `CloudXR2DUI.hideStatusIfShowing()`. Messages from
  any other source are left alone. The auto-dismiss timeout is routed through the
  same path, which also fixes it clearing only the XR banner and leaving the 2D
  mirror up.

  **Every message rendered below the warning icon.** `white-space: pre-line` was
  set on `.error-message-box` to preserve newlines in multi-line notices, but the
  box holds an icon span and a text span on separate source lines, so that markup
  newline rendered as a real line break. The rule is now scoped to
  `#errorMessageText`, the element that actually receives the newline-separated
  text.

  Thanks to @yanziz-nvidia for tracing the error-clearing path.

Fixes #(issue)

## Type of change

- [ ] Documentation update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * System notices now dismiss cleanly without removing newer, unrelated status or error messages.
  * Automatic notice dismissal now follows the same consistent behavior as manual dismissal.

* **Style**
  * Improved system notice formatting in the desktop status banner, preserving intended line breaks without disrupting icons or layout.

### DIFF
--- a/deps/cloudxr/webxr_client/src/App.tsx
+++ b/deps/cloudxr/webxr_client/src/App.tsx
@@ -685,14 +685,23 @@ function AppContent() {
   // static colors at render time rather than subscribing.
   const [systemNoticeLevel, setSystemNoticeLevel] = useState<'warning' | 'info'>('warning');
 
+  /** Exact text this component last wrote to the shared 2D status box. */
+  const systemNoticeTextRef = useRef<string | null>(null);
+
   /** Take the notice down, in both surfaces, and cancel any pending auto-dismiss. */
   const dismissSystemNotice = () => {
     systemNotice.value = null;
     setSystemNoticeVisible(false);
-    // showStatus() sets .show and never clears itself, so without this the 2D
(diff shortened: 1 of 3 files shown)


yanziz-nvidia (approved):

yanziz-nvidia:
comments are very verbose.. but other than that all good!

PR #881, opened 2026-08-04
build: add project C++ warning set and configure ruff lint rules

maxwbuckley:
## Description

Two related build-hygiene changes that were previously unenforced.

**1. `build:` project C++ warning set with opt-in warnings-as-errors**

Adds `cmake/CompilerWarnings.cmake`, included from the root `CMakeLists.txt` *after*
`add_subdirectory(deps)`. Directory-scope compile options are only inherited by
subdirectories added after the call, so first-party targets get the flags while the
third-party trees (OpenXR SDK, yaml-cpp, pybind11, mcap, flatbuffers, Catch2) keep
their own.

Flags on GNU/Clang: `-Wall -Wextra -Wno-missing-field-initializers -Wnon-virtual-dtor
-Woverloaded-virtual -Wimplicit-fallthrough -Wextra-semi`. MSVC gets `/W4 /permissive-`.
Two options: `ISAAC_TELEOP_ENABLE_WARNINGS` (ON) and `ISAAC_TELEOP_WARNINGS_AS_ERRORS`
(OFF, opt in per build/CI).

Fixes everything the set surfaced:

- `properties.serial` is a fixed `char[256]`, never a pointer, so the
  `properties.serial ? ... : ""` guard was always-true dead code
  (`-Wpointer-bool-conversion`, 3 sites). Replaced with a `strnlen`-bounded
  `std::string` construction, which additionally guards against a runtime that fills
  the array without a terminator.
- Removed the empty, unused `print_xdev_info` (`-Wunused-function`).
- Commented out the unused `argc` parameter name in six `main()` definitions.

`robstride_bus`'s private members are used only inside `#ifdef __linux__`, so Clang
reports them unused when the file compiles to its throwing stub. That warning is
correct but unactionable off Linux (GCC has no equivalent), so it is suppressed for
that one target on non-Linux only.

**2. `chore:` ruff lint rules, and the fixes they surfaced**

The `ruff` and `ruff-format` pre-commit hooks were running with no configuration at
all, so only ruff's built-in defaults (E4/E7/E9 + F) were ever enforced. Adds a
`[tool.ruff]` section selecting the correctness-oriented groups: B, C4, PIE, PERF,
PLE, PLW, LOG, G, ASYNC and RUF. Formatting is unchanged (ruff-format defaults, which
the tree already matches). Everything in `select` is currently clean, so the hook is
enforceable as-is. Each entry in `ignore` is a deliberate call with its reason inline.

Bug-shaped findings fixed:

- `B023`: a closure captured the loop variable `phase` in the synthetic camera source.
  Hoisting it above the loop and passing `phase` explicitly also stops rebuilding the
  closure on every generated frame.
- `RUF043`: pytest `match="libcloudxr.so"` treated `.` as a regex wildcard where a
  literal filename was meant; the remaining patterns are intentional regexes and are
  now raw strings.
- `RUF012`: three mutable class attributes annotated `ClassVar`, one of them on the
  `EnvConfig` singleton.
- `LOG014`: pass the caught exception via `exc_info=exc` rather than relying on ambient
  `sys.exc_info()`.
- `RUF013` implicit Optional, `B904` exception chaining, `B011` `assert False` (removed
  by `python -O`), `B017` blind `pytest.raises(Exception)`, `G004` eager f-string log
  formatting, `PLW1510` implicit `subprocess` check, plus safe C4/PIE/PERF/RUF autofixes.

`TRY004` (`ValueError` -> `TypeError` in `TeleopSessionConfig` validation) was left
alone: it is a public API behaviour change, not a lint fix.

## Type of change

- [ ] Documentation update

## Testing

Validated on two toolchains.

**Apple Clang 21 / arm64 (macOS)** — 88 first-party TUs configure on that host. Clean
build with `ISAAC_TELEOP_WARNINGS_AS_ERRORS=ON`. `ctest` 158/160, the two failures being
environment gaps (no wuji-sdk wheel, no CloudXR SDK tarball) confirmed pre-existing by
re-running with these changes stashed.

**GCC 13.3 / x86_64, Ubuntu 24.04, RTX PRO 6000 Blackwell, CUDA 13.0** —

```
cmake --preset py3.12 -DCMAKE_BUILD_TYPE=Release -DISAAC_TELEOP_WARNINGS_AS_ERRORS=ON
cmake --build --preset py3.12 --parallel 24 --clean-first
```

- 328 first-party TUs, **zero warnings**, clean under `-Werror`. This covers the targets
  macOS cannot reach: `rebot_devarm_leader` on its real SocketCAN path (so the
  `-Wno-unused-private-field` suppression above is confirmed to be needed only off
  Linux), and the whole Televiz tree with `BUILD_VIZ` auto-ON (Vulkan + CUDA + glslang
  all present).
- `BUILD_PLUGIN_OGLO=ON` and `BUILD_PLUGIN_NOITOM_MOCAP=ON` also build clean under
  `-Werror`.
- `ctest`: 308/309. The one failure, `cloudxr_test_launcher`, is the missing CloudXR SDK
  (no NGC key on this host, so the download 404s and `get_sdk_path()` raises) — the same
  environment gap as on macOS, not a code failure.

**Known limitation — `BUILD_PLUGIN_OAK_CAMERA=ON` does not build with
`ISAAC_TELEOP_WARNINGS_AS_ERRORS=ON`.** DepthAI pulls XLink via FetchContent from inside
`src/plugins/oak/`, so that third-party tree *does* inherit the flags — 41 errors, all in
`xlink-src` (16 `unused-parameter`, 11 `stringop-truncation`, 5 `unused-variable`, 3
`parentheses`, 2 `unused-but-set-variable`, 2 `implicit-fallthrough`, 1
`unused-but-set-parameter`, 1 `switch`). Note `stringop-truncation` is a GCC default
rather than part of this warning set, so plain `-Werror` would break XLink regardless.
This is why `ISAAC_TELEOP_WARNINGS_AS_ERRORS` defaults to OFF and CI is left at the
default; enabling it in CI needs OAK excluded or a suppression scoped to the fetched
tree. The header comment in `cmake/CompilerWarnings.cmake` records the gap.

MSVC is unvalidated.

## Checklist

Documentation: no user-facing behaviour changes, so no doc updates. The two new CMake
options are documented in the header comment of `cmake/CompilerWarnings.cmake`.

Tests: no new tests — this is a build-configuration and lint-configuration change, and
it is exercised by the existing suite building and passing under `-Werror` on both
toolchains above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,13 @@ endif()
 # Build dependencies (OpenXR SDK, etc.)
 add_subdirectory(deps)
 
+# Project warning set. Deliberately after add_subdirectory(deps): directory-scope
+# compile options are inherited only by subdirectories added *after* this call, so
+# third-party trees keep their own flags while everything below (src/, examples/,
+# plugins) is held to ours. See cmake/CompilerWarnings.cmake.
+include(cmake/CompilerWarnings.cmake)
+isaac_teleop_enable_compiler_warnings()
+
 # Enable CTest at top level so tests from subdirectories are discoverable
 if(BUILD_TESTING)
     # Make sure to call this after `deps` is added so that Catch2 is available
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -1,0 +1,67 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# ==============================================================================
+# Compiler warnings for first-party code
+# ==============================================================================
+# isaac_teleop_enable_compiler_warnings() applies the project's warning set to the
+# CALLING directory scope, which CMake then inherits into every subdirectory added
+# after the call. The top-level CMakeLists.txt therefore calls it *after*
+# add_subdirectory(deps) so third-party trees (OpenXR SDK, yaml-cpp, pybind11,
+# mcap, flatbuffers, Catch2, ...) keep building with their own flags and are never
+# held to warning levels we do not control.
+#
+# Known gap: the OAK plugin (OFF by default) fetches DepthAI, and transitively XLink,
+# via FetchContent from inside src/plugins/, so that tree DOES inherit these flags. It
+# builds as long as warnings are not errors; with ISAAC_TELEOP_WARNINGS_AS_ERRORS=ON it
+# fails with 41 errors in xlink-src on GCC 13 (unused-parameter, stringop-truncation,
+# parentheses, ...). Note stringop-truncation is a GCC default rather than part of this
+# set, so plain -Werror would break XLink even with ISAAC_TELEOP_ENABLE_WARNINGS=OFF.
+# Turning warnings-as-errors on in CI therefore needs OAK excluded, or a -Wno-
+# suppression scoped to the fetched tree.
+#
+# The OGLO and Noitom plugins also fetch from inside src/plugins/, but neither
+# contributes third-party translation units — OGLO uses header-only nlohmann/json plus
+# the system libdbus, and Noitom links a prebuilt libMocapApi.so — so both build clean
+# under -Werror.
+
+option(ISAAC_TELEOP_ENABLE_WARNINGS "Enable the project warning set on first-party C++ targets" ON)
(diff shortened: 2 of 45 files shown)


maxwbuckley:
Superseded by a split into two independent PRs, so the C++ and Python changes can be reviewed separately:

- #882 — `build:` project C++ warning set (CMake + .cpp only)
- #883 — `chore:` ruff lint rules (pyproject.toml + .py only)

The two commits touch disjoint file sets, so the split is one commit per PR with no rebasing. Closing this one.

PR #882, opened 2026-08-04
build: add project C++ warning set with opt-in warnings-as-errors

maxwbuckley:
## Description

Adds `cmake/CompilerWarnings.cmake`, included from the root `CMakeLists.txt` *after*
`add_subdirectory(deps)`. Directory-scope compile options are only inherited by
subdirectories added after the call, so first-party targets get the flags while the
third-party trees (OpenXR SDK, yaml-cpp, pybind11, mcap, flatbuffers, Catch2) keep
their own.

Flags on GNU/Clang: `-Wall -Wextra -Wno-missing-field-initializers -Wnon-virtual-dtor
-Woverloaded-virtual -Wimplicit-fallthrough -Wextra-semi`. MSVC gets `/W4 /permissive-`.
Two options: `ISAAC_TELEOP_ENABLE_WARNINGS` (ON) and `ISAAC_TELEOP_WARNINGS_AS_ERRORS`
(OFF, opt in per build/CI).

Ordering handles `deps/`, but not a tree fetched from *inside* an already-flagged
directory: a subdirectory snapshots `COMPILE_OPTIONS` at the point it is added, so what
the plugins pull in via FetchContent would inherit our flags. The OAK plugin fetches
DepthAI, which fetches XLink, and XLink does not build at `-Wall -Wextra -Werror` — it
sets `-Wno-unused-parameter` for itself precisely because it does not hold to that. Two
mechanisms keep our flags out:

- `isaac_teleop_third_party_scope_begin()`/`_end()` clear the calling directory's
  `COMPILE_OPTIONS` across the `add_subdirectory()`/`FetchContent_MakeAvailable()` and
  restore them afterwards, so the fetched tree builds with its own flags while
  first-party targets later in the same file still get ours. Used for DepthAI and SDL2
  in the OAK plugin, and nlohmann/json in OGLO.
- `isaac_teleop_mark_include_dirs_system()` moves a dependency's interface includes to
  `-isystem`, because a warning raised inside a header is attributed to the first-party
  TU that included it — DepthAI's headers alone accounted for 128 `-Wextra-semi` hits in
  our own sources. (`add_subdirectory(... SYSTEM)` does this in one step but needs CMake
  3.25; the project floor is 3.20.)

Fixes everything the set surfaced on first-party code:

- `properties.serial` is a fixed `char[256]`, never a pointer, so the
  `properties.serial ? ... : ""` guard was always-true dead code
  (`-Wpointer-bool-conversion`, 3 sites). Replaced with a `strnlen`-bounded
  `std::string` construction, which additionally guards against a runtime that fills
  the array without a terminator.
- Removed the empty, unused `print_xdev_info` (`-Wunused-function`).
- Commented out the unused `argc` parameter name in six `main()` definitions.
- `oak_camera.cpp` called the deprecated `dai::DeviceInfo::getMxId()`
  (`-Wdeprecated-declarations`); the same file already used `getDeviceId()` in the two
  neighbouring call sites.

`robstride_bus`'s private members are used only inside `#ifdef __linux__`, so Clang
reports them unused when the file compiles to its throwing stub. That warning is
correct but unactionable off Linux (GCC has no equivalent), so it is suppressed for
that one target on non-Linux only.

C++/CMake only — no Python is touched. A companion PR configures the ruff lint rules
for the Python side; the two are independent and can land in either order.

## Type of change

- [ ] Documentation update

## Testing

**GCC 13.3 / x86_64, Ubuntu 24.04, RTX PRO 6000 Blackwell, CUDA 13.0** —

```
cmake --preset py3.12 -DCMAKE_BUILD_TYPE=Release -DISAAC_TELEOP_WARNINGS_AS_ERRORS=ON
cmake --build --preset py3.12 --parallel 24 --clean-first
```

- 339 first-party TUs, **zero warnings**, clean under `-Werror`, with
  `BUILD_PLUGIN_OGLO=ON` and `BUILD_PLUGIN_NOITOM_MOCAP=ON`. This includes
  `rebot_devarm_leader` on its real SocketCAN path (so the `-Wno-unused-private-field`
  suppression above is confirmed to be needed only off Linux), and the whole Televiz
  tree with `BUILD_VIZ` auto-ON (Vulkan + CUDA + glslang all present).
- `BUILD_PLUGIN_OAK_CAMERA=ON` (vcpkg toolchain) also builds clean under `-Werror`.
  Verified from the generated `flags.make` that the containment cuts in one direction
  only: `camera_plugin_oak` gets the full project set including `-Werror` and reaches
  DepthAI's headers via `-isystem`, while `depthai-core` and `XLink` compile with their
  own upstream warning flags and none of ours. Before the containment this configuration
  failed with 41 errors inside `xlink-src`.
- `ctest`: 309/310. The one failure, `cloudxr_test_launcher`, is the missing CloudXR SDK
  (no NGC key on this host, so the download 404s and `get_sdk_path()` raises) — an
  environment gap, not a code failure.

MSVC is unvalidated, so CI is deliberately left at the `ISAAC_TELEOP_WARNINGS_AS_ERRORS`
OFF default.

## Checklist

Documentation: no user-facing behaviour changes, so no doc updates. The two new CMake
options are documented in the header comment of `cmake/CompilerWarnings.cmake`.

Tests: no new tests — this is a build-configuration change, and it is exercised by the
existing suite building and passing under `-Werror` on both toolchains above.

## Summary by CodeRabbit

- **Build Improvements**
  - Added configurable compiler warnings, including optional warnings-as-errors support.
  - Simplified build configuration and improved conditional handling for examples, visualization, plugins, and ROS 2 components.
  - Improved compatibility across supported compilers and platforms.
  - Reduced noise from third-party dependency warnings.

- **Bug Fixes**
  - Improved device identification in camera connection logs.
  - Prevented unsafe reads when processing fixed-length device serial numbers.
  - Corrected skipped-device reporting to display the parsed serial information.


cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,120 @@
...
+# add_subdirectory(deps) so third-party trees (OpenXR SDK, yaml-cpp, pybind11,
+# mcap, flatbuffers, Catch2, ...) keep building with their own flags and are never
+# held to warning levels we do not control.
+#
+# Ordering alone does not cover code fetched from inside src/plugins/, which is added
+# after the call and would otherwise inherit these flags. See the third-party
+# containment scope at the bottom of this file for how those trees opt out.
+
+option(ISAAC_TELEOP_ENABLE_WARNINGS "Enable the project warning set on first-party C++ targets" ON)
+option(ISAAC_TELEOP_WARNINGS_AS_ERRORS "Promote the project warning set to errors (-Werror / /WX)" OFF)
```
aristarkhovNV, line 20: what is the plan for enabling warnings as errors by defailt?

CMakeLists.txt
```diff
@@ -133,6 +133,13 @@ endif()
 # Build dependencies (OpenXR SDK, etc.)
 add_subdirectory(deps)
 
+# Project warning set. Deliberately after add_subdirectory(deps): directory-scope
+# compile options are inherited only by subdirectories added *after* this call, so
+# third-party trees keep their own flags while everything below (src/, examples/,
+# plugins) is held to ours. See cmake/CompilerWarnings.cmake.
+include(cmake/CompilerWarnings.cmake)
+isaac_teleop_enable_compiler_warnings()
```
aristarkhovNV: This is a good opportunity to split off `src/` and `examples/` portions into their own cmakelists files for which the warnings would be enbled globally, with no placement quirks

cmake/CompilerWarnings.cmake
maxwbuckley: Plan is to roll it out in two steps rather than flipping the default here.

1. A follow-up PR turns `ISAAC_TELEOP_WARNINGS_AS_ERRORS=ON` in the Linux CI matrix, so a regression fails a check even while the default stays OFF for local builds.
2. Flip the default ON once the MSVC path (`/W4 /permissive-`) has actually been validated.

The reason for the gap: the set is clean on GCC 13.3 / x86_64 across the whole first-party tree, but I have no Windows host to test on, so an ON default today would risk breaking contributors I cannot verify against. Happy to fold step 1 into this PR instead if you would rather it land in one go.

CMakeLists.txt
maxwbuckley: Good call — done in b611cba, and the placement quirk is gone with it.

- `src/CMakeLists.txt` and `examples/CMakeLists.txt` each call `isaac_teleop_enable_compiler_warnings()` at the top, then add their own subdirectories. The flags now follow the directory layout instead of the order of the root file.
- `src/plugins/CMakeLists.txt` holds the plugin list, so `src/` no longer reaches into grandchild directories.
- The root now just does `add_subdirectory(src)` / `add_subdirectory(examples)`, and `include(cmake/CompilerWarnings.cmake)` only defines the helpers, so it can sit anywhere. `deps/` never calls the function, so third-party trees keep their own flags no matter what order things are added in.

Two things that did not change: binary directories (CMake mirrors the relative source path either way), and the FetchContent containment in the OAK/OGLO plugins — `src/plugins/` inherits the flags, so what those fetch still would too. Nothing crossed a scope: no `PARENT_SCOPE` variables are set under `src/` or `examples/`, and `BUILD_PYTHON_BINDINGS` is an `option()`, so it stays a cache variable. Plugins are now configured before examples rather than after; neither tree references the other's targets.

Verified with `ISAAC_TELEOP_WARNINGS_AS_ERRORS=ON` on GCC 13.3 / x86_64: clean build, ctest 162/162, and `compile_commands.json` carries the exact flag sequence on all 90 first-party TUs (78 `src/`, 12 `examples/`) and on zero third-party TUs. `BUILD_VIZ` was OFF on this host, so `src/viz` was not rebuilt this round — it is added from `src/CMakeLists.txt` by the same mechanism as its siblings.

examples/native_openxr/xdev_list/main.cpp
```diff
@@ -156,7 +150,10 @@ class XDevListApp : public HeadlessApp
                 throw std::runtime_error("Failed to get properties for XDev " + std::to_string(xdevId));
             }
 
-            std::string serial_str = properties.serial ? properties.serial : "";
+            // serial is a fixed char[256], never a pointer, so a null check would always be true.
+            // Bound the length so a runtime that fills the array without a terminator cannot
+            // walk past it.
+            std::string serial_str(properties.serial, ::strnlen(properties.serial, sizeof(properties.serial)));
```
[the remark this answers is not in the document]
maxwbuckley: Valid — fixed in 41ba773. `name` and `serial` are both `char[256]` in `XrXDevPropertiesMNDX` (deps/cloudxr/openxr_extensions/XR_MNDX_xdev_space.h:53-54), so bounding one and not the other was inconsistent. `name_str` is now built with the same `strnlen`-bounded construction and used in both log lines.

maxwbuckley:
{"body":"Rebased onto current `main` and out of draft. The only conflict was in the root `CMakeLists.txt`, from 0467dc5 adding `examples/mujoco_xr` to the examples block this PR relocates — it now sits in `examples/CMakeLists.txt` under the same `BUILD_VIZ` guard.\n\nRe-verified after the rebase on GCC 13.3 / x86_64 with `ISAAC_TELEOP_WARNINGS_AS_ERRORS=ON`: clean build, ctest 166/166, and `compile_commands.json` shows the warning set on all 94 first-party TUs (76 from the source tree, 18 generated) and on no third-party TU.\n\nTwo caveats worth stating rather than hiding. `BUILD_VIZ` was OFF on this host (no Vulkan/CUDA/glslang), so `src/viz` and the new `examples/mujoco_xr` were not compiled under `-Werror` here — both are added by the same mechanism as their siblings, but CI is the real check. And MSVC is still unvalidated, which is why the option stays OFF by default.\n\nWhich is the ask: the workflow runs are sitting at `action_required` behind the fork-PR approval gate, so CI has not run on this PR yet. Could you hit \"Approve and run workflows\" when you get a chance? Happy to fix anything it turns red."}

aristarkhovNV (approved):

PR #883, opened 2026-08-04
chore: configure ruff lint rules and fix what they surfaced

maxwbuckley:
## Description

The `ruff` and `ruff-format` pre-commit hooks were running with no configuration at
all, so only ruff's built-in defaults (E4/E7/E9 + F) were ever enforced. Adds a
`[tool.ruff]` section selecting the correctness-oriented groups: B, C4, PIE, PERF,
PLE, PLW, LOG, G, ASYNC and RUF. Formatting is unchanged (ruff-format defaults, which
the tree already matches).

Everything in `select` is currently clean, so the hook is enforceable as-is. Each entry
in `ignore` is a deliberate call with its reason inline; the notable ones:

- `B905` (`zip` `strict=`) is a per-call-site behaviour change, turning a silent
  truncation into a runtime exception. Worth adopting deliberately, not in a lint sweep.
- `RUF005`, `C408` and `RUF007` are style-only rewrites of code that is already correct
  and readable (`a + [b]` → `[*a, b]`, `dict(a=1)` → `{"a": 1}`,
  `zip(x[:-1], x[1:])` → `itertools.pairwise(x)`). Enforcing them means churning working
  call sites for no behaviour change, so they are left to author preference.
- `RUF022` sorts `__all__` alphabetically, which scrambles the semantic grouping
  comments the schema/package `__init__` files rely on.
- `RUF100` is evaluated against `select`, so it flags every `noqa` written for a rule not
  yet enabled (`BLE001`, `PLC0415`, `N803`, ...). Re-enable once those groups are adopted.

No live defects were found. Every rule fired on code that behaves correctly today;
what follows removes fragility, not bugs.

### Correct today, fragile to a later change

- `LOG014`: `_record_error` passes `exc_info=True`, which reads the ambient
  `sys.exc_info()`. Both of its callers invoke it from inside an `except` block, so the
  traceback is logged correctly; the rule is lexical and fires because the logging call
  sits in a helper rather than in the handler itself. Passing the exception explicitly
  makes it independent of the caller's context.
- `B023`: a closure in the frame loop captured the loop variable `phase` by reference.
  It is called immediately in the same iteration, so the value was always correct.
  `fill` is hoisted above the loop and takes `phase` as a parameter, which also stops
  re-creating the function object every frame. Binding it as a default argument would
  satisfy the rule equally.
- `B011`: `assert False` in a test, which `python -O` strips, turning a failure into a
  silent pass. The suite is not run under `-O` today. Replaced by calling the
  constructor directly, so an exception fails the test with its own traceback.
- `B017`: a blind `pytest.raises(Exception)` that would also accept an unrelated
  failure. The test passes for the right reason today; naming the accepted exception set
  keeps it that way.
- `RUF043`: pytest `match="libcloudxr.so"` treats `.` as a regex wildcard where a literal
  filename was meant. The real message contains the literal, so the assertion passes
  correctly, but it is weaker than it reads. The remaining patterns are intentional
  regexes and are now raw strings.

### Typing and explicitness

- `RUF012`: three mutable class attributes annotated `ClassVar`, one of them on the
  `EnvConfig` singleton.
- `RUF013`: implicit Optional spelled out as `str | None`.
- `B904`: `raise ... from` on re-raises, so the original cause is not lost.
- `G004`: log calls take `%s` arguments rather than eagerly formatted f-strings.
- `PLW1510`: `subprocess.run` calls that inspect `returncode` say `check=False`. The
  tree already spelled this out at 10 call sites, 8 of them in `oob_teleop_adb.py`; this
  covers the stragglers.

`TRY004` (`ValueError` → `TypeError` in `TeleopSessionConfig` validation) was left
alone: it is a public API behaviour change, not a lint fix.

Python only — no C++ or CMake is touched. A companion PR adds the C++ warning set; the
two are independent and can land in either order.

## Type of change

- [ ] Documentation update

## Testing

**Ubuntu 24.04 / x86_64, Python 3.12** —

- `ruff check` and `ruff format --check` both clean at v0.15.1, the version pinned in
  `.pre-commit-config.yaml` (268 files).
- `ctest`: 309/310. The one failure, `cloudxr_test_launcher`, is the missing CloudXR SDK
  (no NGC key on this host, so the download 404s and `get_sdk_path()` raises) — an
  environment gap, not a code failure.

Forward-compat note, not addressed here: at ruff 0.16 `format --check` wants to reflow
Python code blocks embedded in three Markdown files (`examples/teleop_ros2/README.md`,
`src/plugins/oak/README.md`, and one other) — newer ruff formats fenced code in Markdown,
which 0.15.1 does not. None of those files are touched by this PR, but it will surface
whenever the pre-commit rev is bumped past 0.16.

## Checklist

Documentation: no user-facing behaviour changes, so no doc updates. Every non-obvious
`ignore` entry is documented inline in `pyproject.toml`.

Tests: no new tests — this is a lint-configuration change plus the fixes it surfaced,
and it is exercised by the existing suite passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting with preserved exception details and clearer diagnostics.
  * CloudXR headset URL and import checks now report command failures without unexpected exceptions.
  * Refined validation and error matching for more reliable test and runtime behavior.

* **Quality Improvements**
  * Added expanded Python linting and bug-detection rules.
  * Improved type annotations, logging consistency, and handling of abstract interfaces.
  * Simplified internal example and tooling code without changing behavior.

### DIFF
--- a/docs/source/_ext/ecosystem_grid.py
+++ b/docs/source/_ext/ecosystem_grid.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import os
 import re
+from typing import ClassVar
 
 import yaml
 from docutils import nodes
@@ -668,7 +669,7 @@ class DeviceMatrix(Directive):
     """Render one section of ``devices.yaml`` as a table: name, value, Details panel."""
 
     has_content = False
-    option_spec = {
+    option_spec: ClassVar[dict] = {
         "section": lambda arg: directives.choice(arg, tuple(DEVICE_SECTIONS)),
         "data": directives.unchanged,
     }
--- a/examples/camera_viz/pipeline/runner.py
+++ b/examples/camera_viz/pipeline/runner.py
@@ -203,7 +203,7 @@ class VizRunner:
         with self._error_lock:
             if self._error is None:
                 self._error = exc
-        logger.error("VizRunner %s thread failed: %s", where, exc, exc_info=True)
+        logger.error("VizRunner %s thread failed: %s", where, exc, exc_info=exc)
         self._stop.set()
         with self._data_cond:
             self._data_cond.notify_all()
--- a/examples/camera_viz/sources/oakd.py
+++ b/examples/camera_viz/sources/oakd.py
@@ -30,7 +30,7 @@ import logging
 import threading
 import time
 from dataclasses import dataclass, field
-from typing import List, Optional
+from typing import ClassVar, List, Optional
 
 import numpy as np
 
@@ -120,7 +120,7 @@ class _OakdDevice:
     closes it.
     """
 
-    _SOCKET_MAP = {
+    _SOCKET_MAP: ClassVar[dict[str, str]] = {
         "RGB": "CAM_A",
         "CAM_A": "CAM_A",
         "LEFT": "CAM_B",
--- a/examples/camera_viz/sources/synthetic.py
+++ b/examples/camera_viz/sources/synthetic.py
@@ -234,27 +234,29 @@ class SyntheticStereoSource(FrameSource):
             diag_l = (x_grid_l + y_grid) / float(w + h)
             diag_r = (x_grid_r + y_grid) / float(w + h)
 
+            # Defined once, above the loop: taking phase as a parameter avoids
+            # rebuilding a closure over it on every generated frame.
+            def fill(buf, diag, phase):
+                r = (cp.sin((diag + phase) * 6.2831853) * 127.0 + 128.0).astype(
+                    cp.uint8
+                )
+                g = (
+                    cp.sin((diag + phase + 0.3333) * 6.2831853) * 127.0 + 128.0
+                ).astype(cp.uint8)
+                b = (
+                    cp.sin((diag + phase + 0.6667) * 6.2831853) * 127.0 + 128.0
+                ).astype(cp.uint8)
+                buf[..., 0] = r
+                buf[..., 1] = g
+                buf[..., 2] = b
+                buf[..., 3] = 255
+
             while not self._stop.is_set():
                 t = (time.monotonic_ns() - self._t0_ns) * 1e-9
                 phase = (t * self._hue_speed_hz) % 1.0
 
-                def fill(buf, diag):
-                    r = (cp.sin((diag + phase) * 6.2831853) * 127.0 + 128.0).astype(
-                        cp.uint8
-                    )
-                    g = (
-                        cp.sin((diag + phase + 0.3333) * 6.2831853) * 127.0 + 128.0
-                    ).astype(cp.uint8)
-                    b = (
-                        cp.sin((diag + phase + 0.6667) * 6.2831853) * 127.0 + 128.0
-                    ).astype(cp.uint8)
-                    buf[..., 0] = r
-                    buf[..., 1] = g
-                    buf[..., 2] = b
-                    buf[..., 3] = 255
-
-                fill(self._left[self._write_idx], diag_l)
-                fill(self._right[self._write_idx], diag_r)
+                fill(self._left[self._write_idx], diag_l, phase)
+                fill(self._right[self._write_idx], diag_r, phase)
                 cp.cuda.Stream.null.synchronize()
 
                 with self._lock:
--- a/examples/camera_viz/sources/zed.py
+++ b/examples/camera_viz/sources/zed.py
@@ -310,7 +310,7 @@ class _ZedCamera:
 
         # Pyzed Mats are pitched GPU buffers — one per eye, reused every
         # grab() / retrieve_image() pair.
-        for eye, slot in self._slots.items():
+        for slot in self._slots.values():
             slot.zed_mat = sl.Mat()
 
         self._camera = camera
--- a/examples/camera_viz/transports/rtp_h264_sender.py
+++ b/examples/camera_viz/transports/rtp_h264_sender.py
@@ -252,7 +252,7 @@ class RtpH264Sender:
                     raise RuntimeError(
                         f"RtpH264Sender: encode failed {consecutive_encode_failures} "
                         f"times in a row; surfacing to supervisor for full restart"
-                    )
+                    ) from e
                 continue
 
             for pkt in packets:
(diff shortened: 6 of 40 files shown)


nv-jakob:
Hey we quickly looked over this, general feel is good. What are you feeling about merging this?

maxwbuckley:
Thanks for looking! I'm happy to merge this — it's rebased onto current `main` and out of draft now.

One thing worth calling out from the rebase. The only conflict was in `cloudxr/launcher.py`, and it resolved in `main`'s favour: 53a1d7a ("refuse to start over a live runtime") replaced `_cleanup_stale_runtime`'s `fuser` call with the `is_runtime_live` guard, which deleted the exact `subprocess.run` this PR was adding `check=False` to. That fix is obsolete, so I dropped it — `cloudxr/launcher.py` is now byte-identical to `main` and the PR touches 28 files instead of 29. Nothing in the `start_wss_proxy` deprecation or the `health_check` docs change is affected.

Re-verified after the rebase, at the ruff version pinned in `.pre-commit-config.yaml` (v0.15.1):

- `ruff check .` and `ruff format --check .` both clean.
- `SKIP=check-copyright-year pre-commit run --all-files` passes.
- The rule set is still load-bearing: applied to unmodified `main` it reports 46 violations, all of which this PR fixes.
- The Python suites (cloudxr, schema, retargeting_engine, teleop_session_manager, camera_viz) give identical results on this branch and on `main` — no regressions. I wasn't able to run the full `ctest` suite locally this time as I don't have the CloudXR SDK/CUDA build configured on this machine, so CI is the real check.

Which is the ask: the workflow runs are all sitting at `action_required` (the fork-PR approval gate), so CI has never actually run on this PR. Could you hit "Approve and run workflows" on the Checks tab when you get a chance? Happy to fix up anything it turns red.

nv-jakob:
> Thanks for looking! I'm happy to merge this — it's rebased onto current `main` and out of draft now.
> 
> One thing worth calling out from the rebase. The only conflict was in `cloudxr/launcher.py`, and it resolved in `main`'s favour: [53a1d7a](https://github.com/NVIDIA/IsaacTeleop/commit/53a1d7a00f1caf61aa4c5e3198540b37c78cbf97) ("refuse to start over a live runtime") replaced `_cleanup_stale_runtime`'s `fuser` call with the `is_runtime_live` guard, which deleted the exact `subprocess.run` this PR was adding `check=False` to. That fix is obsolete, so I dropped it — `cloudxr/launcher.py` is now byte-identical to `main` and the PR touches 28 files instead of 29. Nothing in the `start_wss_proxy` deprecation or the `health_check` docs change is affected.

@jiwenc-nv this is okay right?

aristarkhovNV:
@maxwbuckley would you please merge with the latest upstream main that contains fixes for CI

maxwbuckley:
@aristarkhovNV Done — `main` is merged in as of 0f3db7c.

I merged rather than rebased so the "changes since last review" view stays intact. The merge was clean, no conflicts, and the PR's own diff is unchanged: still the same 33 files, +147/-74 against `main` (that count includes the follow-up commit f43c5e1, which is why it is 33 rather than the 28 I quoted earlier in this thread).

That pulls in the CI fix you were pointing at. The one red check here was **Build Teleop Web App**, and it was an upstream breakage rather than anything in this PR:

```
npm error Could not resolve dependency:
npm error peer react@">=19 <19.3" from @react-three/fiber@9.7.0
```

56a7adc ("fix(webxr): constrain React to the Fiber-compatible 19.2 minor", #1091) is the fix, and it's now in this branch. Worth noting this PR touches no webclient files at all, so that job was never failing on its own account.

Re-verified after the merge at the ruff version pinned in `.pre-commit-config.yaml` (v0.15.1):

- `ruff check .` — all checks passed.
- `ruff format --check .` — 324 files already formatted.
- The rules still hold against everything that landed in `main` since the last update, so no follow-up fixup commit was needed this time.

One ask: the push has put the workflows back to `action_required` (the fork-PR gate re-arms on every push, not just force-pushes). Could you hit "Approve and run workflows" on the Checks tab once more? Happy to chase down anything that comes back red.

aristarkhovNV (approved):

aristarkhovNV:
@maxwbuckley could you please merge & resolve a new conflict? Thanks!

maxwbuckley:
@aristarkhovNV Merged and resolved — head is now 0974388.

**The conflict.** One file, `src/python/isaacteleop/rig/launcher.py`. It was a modify/delete: 9052470 ("refactor!: drop the tmux rig launcher") removed the whole `isaacteleop.rig` module, and the only thing this branch did to that file was add `check=False` to a `subprocess.run`. That call no longer exists, so the fix is obsolete and I resolved in favour of main's deletion. The three `rigs/*.yaml` deletions from the same commit merged cleanly on their own. Nothing else in the branch was affected.

This is the second time that has happened here — the same thing hit `cloudxr/launcher.py` on the previous merge. Both times the lint fix was deleted out from under the PR rather than contested, which I think is a good sign for the rule set rather than a bad one.

**One follow-up commit.** Code that landed since tripped two of the new rules, so 0974388 clears them:

- `B007` — `examples/camera_viz/controls/bindings.py`: the `enumerate()` index in `_cycle_shapes` is never used, so the loop is now `for target in self._targets`. Note the two neighbouring loops in that class *do* use the index for `self._strategies[i]`; only this one was unused, and I confirmed via AST that `i` is unreferenced in the function body.
- `C420` — `examples/mcap_record_replay/python/replay_joint_se3_pose.py`: `{cid: 0 for cid in trackers}` becomes `dict.fromkeys(trackers, 0)`.

**Verification** at the ruff pinned in `.pre-commit-config.yaml` (v0.15.1):

- `ruff check .` — all checks passed.
- `ruff format --check .` — 339 files already formatted.
- `tests/python/examples/camera_viz/test_controls.py` — 73 passed.

Two caveats worth stating rather than glossing: that suite does not actually exercise `_cycle_shapes`, which is why I did the AST check instead of leaning on it; and `test_runner_controls.py` can't run here because `isaacteleop` isn't installed on this machine, so CI remains the real check on both.

The PR is `MERGEABLE` again — 34 files, +148/-76 against `main`. All commits including the merge are DCO signed off. The push has re-armed the fork-PR gate, so the workflows are back at `action_required`; a "Approve and run workflows" click when you have a moment and this should be green again, as it was at 0f3db7c.

Issue #871, opened 2026-08-03
WebXR client: default reference space (`auto`) and default Y offset (-155 cm) are mutually inconsistent


jiwenc-nv:
The CloudXR WebXR client ships two defaults that contradict each other, and the combination leaves world-locked content sitting at a fixed offset from the viewer's head instead of anchored to the room. This is independent of the OpenXR application being streamed: anything that places content at a known height above the floor is affected. Setting **Reference Space = `local-floor`** and **Y Offset = `0`** in the client's config panel resolves it, and both are non-default.

Two defaults that cannot both be right:

- `referenceSpaceType` defaults to `'auto'` (`src/CloudXR2DUI.tsx:539`), which tries `local-floor → local → viewer → unbounded` (`helpers/react/CloudXRComponent.tsx:219-231`).
- `xrOffsetY` defaults to `-155` cm (`src/index.html:1287`), applied unconditionally with no branch on which space `auto` actually won (`helpers/react/CloudXRComponent.tsx:284-290`).

A −1.55 m vertical offset only makes sense for a head-height origin (`local` / `viewer`) — it is a hand-rolled floor estimate. But `auto` *prefers* `local-floor`, which is already floor-referenced, so on any headset that supports it the offset is a second correction stacked on a correct one.

Two smaller things in the same chain:

- `viewer` is head-locked by definition, so anything landing on it gets content rigidly attached to the face. `unbounded` is world-locked and is ranked *below* it, which looks backwards.
- Every fallback failure is swallowed by `catch (_) {}`, so nothing records which space won. The OpenXR server side cannot observe it either, which makes misplaced content hard to attribute to the client at all.

Suggested fixes: default `referenceSpace` to `local-floor` and `xrOffsetY` to `0`; apply a synthetic floor offset only when the chosen space is not floor-referenced; log the winning space; move `viewer` last.

jiwenc-nv:
Had some internal discussion and we concluded that `local` tracking mode is the right default.

In Isaac Lab, viewport height is also not determined by floor height, instead it should be determined by the scene.

PR #899, opened 2026-08-04
viz: let apps choose the XR reference space

jiwenc-nv:
## Description

`VizSession` never set a reference space, so `OpenXrSession` fell through to its `XR_REFERENCE_SPACE_TYPE_LOCAL` default — an origin at head height. Any app that draws world-locked geometry at a known height above the floor gets that height wrong by roughly a whole person, and the symptom is only visible on a headset. `VizSessionConfig::xr_reference_space` now names the choice (`kLocal` / `kLocalFloor` / `kStage`), exposed to Python as `viz.XrReferenceSpace`. The default stays `kLocal`, so no existing consumer changes behaviour.

An unavailable space throws naming the space rather than silently substituting a different origin, and `create_reference_space` logs what the runtime does offer beside what was asked for — the evidence anyone would want before adding a fallback chain.

The CloudXR WebXR client has its own half of this problem: `auto` prefers `local-floor` while a −155 cm vertical offset is applied whichever space it lands on, so the two corrections stack. The root README records the pairing that works; #871 asks why it is not the default.

Extracted from `jiwenc/mujoco-xr-app`. #900 (`examples/mujoco_xr`) is the first consumer and stacks on this.

## Type of change

- [ ] Documentation update

## Testing

Ubuntu 22.04 / aarch64 (Tegra). Full `cmake --preset py3.12 -DBUILD_VIZ=ON` build green; `ctest -R viz` green (4 `viz_python_*` + 3 `camera_viz_*`). Verified from Python that `viz.XrReferenceSpace` imports and `VizSessionConfig().xr_reference_space` defaults to `kLocal`.

`kLocalFloor` itself needs a headset and a CloudXR runtime, so the space actually being created is not exercised here.

## Checklist

- [ ] I have added tests that prove my fix/feature works (or explained why not) — viz has no test that opens an XR session; the default-unchanged half is what the existing suite covers.

## Summary by CodeRabbit

* **New Features**
  * Added configurable OpenXR reference spaces: local, local-floor, and stage.
  * Exposed reference-space selection through Python session configuration.
  * Sessions now use the selected world-origin convention, with clear handling when it is unavailable.
* **Documentation**
  * Added Quick Start guidance for keeping CloudXR WebXR content floor-anchored.
  * Documented `local-floor` configuration, zero vertical offset, defaults, and fallback behavior.
* **Diagnostics**
  * Improved visibility into reference spaces supported by the XR runtime.


farbod-nv (approved):

jiwenc-nv:
We don't really need this one after all.

Issue #866, opened 2026-08-01
cxr: XR_SPACE_LOCATION_POSITION_TRACKED_BIT tracking bits unconditionally


jiwenc-nv:
## Summary

CloudXR runtime reports `XR_SPACE_LOCATION_POSITION_TRACKED_BIT` and `ORIENTATION_TRACKED_BIT` set on every frame regardless of whether anything is being tracked. 

On those interfaces the bits carry no information, so applications cannot implement occlusion or degradation handling.

PR #909, opened 2026-08-05
Expose is_tracked for OpenXR tracking loss

sgrizan-nv:
Require VALID and TRACKED bits for head is_valid so CloudXR client disconnect no longer leaves a placeholder pose marked valid.

## Summary by CodeRabbit

* **New Features**
  * Added separate head-pose validity and active-tracking status indicators.
  * Valid pose data remains available during temporary tracking loss, while tracking status accurately reflects inactivity.
  * Exposed tracking status through head-pose inputs, outputs, transformations, and Python interfaces.

* **Bug Fixes**
  * Improved handling of invalid head poses by providing a default pose when pose data is unavailable.

* **Documentation**
  * Clarified the difference between valid pose data and active tracking.

### DIFF
--- a/src/core/live_trackers/cpp/live_head_tracker_impl.cpp
+++ b/src/core/live_trackers/cpp/live_head_tracker_impl.cpp
@@ -57,8 +57,12 @@ void LiveHeadTrackerImpl::update(int64_t monotonic_time_ns)
         throw std::runtime_error("[HeadTracker] xrLocateSpace failed: " + std::to_string(result));
     }
 
-    bool position_valid = (location.locationFlags & XR_SPACE_LOCATION_POSITION_VALID_BIT) != 0;
-    bool orientation_valid = (location.locationFlags & XR_SPACE_LOCATION_ORIENTATION_VALID_BIT) != 0;
+    // OpenXR: VALID => pose may be read; TRACKED => actively tracked. Keep these
+    // separate so is_valid stays OpenXR-faithful (CloudXR may leave VALID set after disconnect).
+    const bool position_valid = (location.locationFlags & XR_SPACE_LOCATION_POSITION_VALID_BIT) != 0;
+    const bool orientation_valid = (location.locationFlags & XR_SPACE_LOCATION_ORIENTATION_VALID_BIT) != 0;
+    const bool position_tracked = (location.locationFlags & XR_SPACE_LOCATION_POSITION_TRACKED_BIT) != 0;
+    const bool orientation_tracked = (location.locationFlags & XR_SPACE_LOCATION_ORIENTATION_TRACKED_BIT) != 0;
 
     if (!tracked_.data)
     {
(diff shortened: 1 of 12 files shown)


src/core/live_trackers/cpp/live_head_tracker_impl.cpp
```diff
@@ -57,26 +57,32 @@ void LiveHeadTrackerImpl::update(int64_t monotonic_time_ns)
         throw std::runtime_error("[HeadTracker] xrLocateSpace failed: " + std::to_string(result));
     }
 
-    bool position_valid = (location.locationFlags & XR_SPACE_LOCATION_POSITION_VALID_BIT) != 0;
-    bool orientation_valid = (location.locationFlags & XR_SPACE_LOCATION_ORIENTATION_VALID_BIT) != 0;
+    // OpenXR: VALID => pose may be read; TRACKED => actively tracked. Keep these
+    // separate so is_valid stays OpenXR-faithful (CloudXR may leave VALID set after disconnect).
+    const bool position_valid = (location.locationFlags & XR_SPACE_LOCATION_POSITION_VALID_BIT) != 0;
+    const bool orientation_valid = (location.locationFlags & XR_SPACE_LOCATION_ORIENTATION_VALID_BIT) != 0;
+    const bool position_tracked = (location.locationFlags & XR_SPACE_LOCATION_POSITION_TRACKED_BIT) != 0;
```
jiwenc-nv, line 64: see https://github.com/NVIDIA/IsaacTeleop/issues/866:  I don't think we are setting XR_SPACE_LOCATION_POSITION_TRACKED_BIT correctly...

sgrizan-nv: Doesn't the bug mention that the head poses are unaffected and forward the client's flags faithfully?

jiwenc-nv: hmm, the ticket seems to be a bit confusing.  let me fix the wording.

src/core/schema/fbs/head.fbs
```diff
@@ -12,8 +12,11 @@ table HeadPose {
     // The concrete pose data
     pose: Pose (id: 0);
 
-    // Whether the head pose data is valid
+    // Whether the pose value is valid to read (OpenXR VALID bits)
     is_valid: bool (id: 1);
+
+    // Whether the pose is actively tracked (OpenXR TRACKED bits)
+    is_tracked: bool (id: 2);
```
jiwenc-nv, line 19: since we are on it.  let's expose both position_tracked and rotation_tracked.

how about use a enum instead of a bool?

nv-jakob (approved):

src/core/live_trackers/cpp/live_head_tracker_impl.cpp
sgrizan-nv: The new wording changes a lot. @nv-jacob, does this mean that currently we're going to always publish is_tracked as true, regardless of the headset connectivity status? The idea behind this MR was to provide the app a way of determining that the headset got disconnected

src/core/schema/fbs/head.fbs
sgrizan-nv: I'd prefer to keep the new field consistent with the existing validity flag and update both to an enum if we decide to separately

PR #913, opened 2026-08-06
build: drop CMake presets, move the wheel tree out of build/

jiwenc-nv:
The presets existed to give each Python version its own build directory, since the interpreter and ABI are baked into a build tree. That isolation came at a price: `cmake -B build` — the form every other page and every plugin still used — became the unsupported path, and the managed trees sat *underneath* `build/`, so `rm -rf build` took out all of them at once.

This goes back to `cmake -B build` as the single documented flow and enforces the isolation directly instead. `ISAAC_TELEOP_PYTHON_VERSION` is stamped into the cache, and changing it on an existing tree is now a hard error naming a fresh directory to use.

That guard is the one addition beyond the mechanical sweep, and it closes a real silent bug: passing `-DISAAC_TELEOP_PYTHON_VERSION=3.12` to a tree configured for 3.11 previously printed "Configuring for Python 3.12" and cached 3.12 while still compiling against the 3.11 venv. Happy to drop it if you'd rather just document "use a fresh directory".

The scikit-build-core tree moves to `build-wheel/<cache-tag>`, outside `build/`, so pip and a classic configure can't collide. Python bindings stay ON by default.

CI keeps its 3-Python matrix via `-DISAAC_TELEOP_PYTHON_VERSION=`; only GitHub-hosted runners build, so a plain `build/` is safe.

## Testing

On Ubuntu 24.04 aarch64:

- `cmake -B build` → build → `cmake --install`: green, 0 errors, produces `wheels/isaacteleop-1.5+local-cp311-cp311-linux_aarch64.whl`.
- Upgrade path: trees stamped with the old boolean `TRUE` are adopted silently rather than erroring — verified a `TRUE`-stamped cache reconfigures clean and re-stamps to `3.11`.
- `sphinx -W` builds clean.


jiwenc-nv:
Superseded by #916, which carries the same change from a branch on NVIDIA/IsaacTeleop rather than a personal fork.

PR #923, opened 2026-08-07
refactor(cloudxr): deprecate the start_wss_proxy knob

jiwenc-nv:
## Description

`CloudXRLauncher(start_wss_proxy=False)` and `--no-launch-wss-proxy` have no callers. Outside the tests that exercise the flag itself, nothing in the tree passes either one — no example, no rig file, no doc page. Meanwhile `add_launcher_arguments` is called from 32 example scripts, so all 32 advertise a flag nobody uses and that, if used, hands you a runtime with no web-client path and no error explaining the silence. The flag's only other contribution is a `ValueError` about its own incompatibility with `setup_oob` / `usb_local` / `host_client`.

The first commit removed it outright. Per review, this now deprecates instead: the parameter and the CLI flag stay as accepted-and-ignored no-ops that announce their own removal in 1.6. The proxy always starts with the runtime, so an out-of-tree integration that already passes `False` gets a notice rather than a `TypeError`, and there's no second code path to keep working. A `TODO(1.6)` marks the removal.

Both default to `None` so an absent knob is distinguishable from an explicit one, and only the latter warns. The notice is both a `DeprecationWarning` and a `WARNING` log line — Python's default filters drop the former unless `stacklevel` lands in `__main__`, which no `--launch-wss-proxy` run does.

## Type of change

- [ ] Documentation update

_(none of the above — deprecation of an unused option)_

## Testing

`pytest src/core/cloudxr_tests/python/test_launcher.py` (30 passed) and `src/core/rig_tests/python/` (68 passed); `SKIP=check-copyright-year pre-commit run --all-files`. The `test_oob_teleop_*` async tests need `pytest-asyncio`, which isn't in the venv I ran locally; untouched by this change.

## Checklist

- [x] I have made corresponding changes to the documentation (no doc referenced this flag)
- [x] I have added tests that prove my fix/feature works (one: the knob is still accepted, warns, and doesn't stop the proxy)

### DIFF
--- a/src/python/isaacteleop/cloudxr/launcher.py
+++ b/src/python/isaacteleop/cloudxr/launcher.py
@@ -21,6 +21,7 @@ import subprocess
 import sys
 import threading
 import time
+import warnings
 from datetime import datetime, timezone
 from pathlib import Path
 
@@ -83,6 +84,7 @@ class CloudXRLauncher:
         setup_oob: bool = False,
         usb_local: bool = False,
         host_client: bool = False,
+        start_wss_proxy: bool | None = None,
     ) -> None:
         """Launch the CloudXR runtime and the WSS proxy.
 
@@ -112,6 +114,9 @@ class CloudXRLauncher:
             host_client: Serve the web client at ``/client/`` on the WSS
                 proxy port.  Assets are fetched once from GitHub Pages into
                 ``TELEOP_WEB_CLIENT_STATIC_DIR`` or ``~/.cloudxr/static-client``.
+            start_wss_proxy: Deprecated no-op kept so existing callers do
+                not hit a :class:`TypeError`; the proxy always starts with
+                the runtime.  Any value emits a deprecation notice.
 
         Raises:
             RuntimeError: If the EULA is not accepted or the runtime
@@ -124,6 +129,8 @@ class CloudXRLauncher:
         self._setup_oob = setup_oob
         self._usb_local = usb_local
         self._host_client = host_client
+        if start_wss_proxy is not None:
+            self._warn_start_wss_proxy_deprecated()
 
         if self._usb_local or self._host_client:
             from .oob_teleop_env import require_web_client_static_dir  # noqa: PLC0415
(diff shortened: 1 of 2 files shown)


src/python/isaacteleop/cloudxr/launcher.py
```diff
@@ -485,15 +430,15 @@ def stop(self) -> None:
...
+        Returns immediately when the runtime is running and the WSS proxy's
+        background thread is alive.  Raises :class:`RuntimeError` with
+        diagnostic details when any monitored component has stopped
+        unexpectedly, allowing embedding applications to perform a
+        controlled teardown.
 
         Raises:
             RuntimeError: If the launcher has not been started, or if
-                the runtime process or (when enabled) the WSS proxy has stopped.
+                the runtime process or the WSS proxy has stopped.
```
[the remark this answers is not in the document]
jiwenc-nv: changed. we should always expect wss is there.

src/python/isaacteleop/cloudxr/launcher.py
```diff
@@ -83,13 +83,11 @@ def __init__(
         setup_oob: bool = False,
         usb_local: bool = False,
         host_client: bool = False,
-        start_wss_proxy: bool = True,
```
yanziz-nvidia, line 86: i feel you should still keep this just in case some downstream app integrated and put this as False? Make a deprecation warning first and then completely remove it in the next version usually is a better developer flow imo.

yanziz-nvidia (approved):

src/python/isaacteleop/cloudxr/launcher.py
jiwenc-nv: makes sense . adding it back for now.

https://github.com/NVIDIA/IsaacTeleop/pull/923/changes/f323d7fe3b262a7651b1be1dced7b7fea9b82de9#diff-d225c17e5efee8c95550c91851e1a55720f006dce6de2189098abf2c919dcaa4

PR #926, opened 2026-08-08
feat(deviceio): identify the headset from its interaction profile

jsepulveda-nvidia:
## Description

Before conducting a code review, its best to learn how the big picture of headset detection and remapping works form this doc: https://docs.google.com/document/d/1dOX4aN6cnj9VV8iaMVsjhS7loJGrN_lSWixJVMpJelY/edit?usp=sharing

A streaming runtime presents every headset through the same OpenXR surface, so the obvious fields do not distinguish them. CloudXR reports `NV_DEVICE_PROFILE=Quest3` **with a PICO connected**, and that same string appears in all 131 archived server logs regardless of hardware. Consumers that must know the real device — a body-tracking skeleton correction, say, where the wrong answer means permuted joint axes — have had no way to ask.

The interaction profile the runtime binds for the controllers *does* track the hardware. This suggests the PICO profile alongside the Oculus one already suggested, and exposes whichever the runtime binds:

- `ControllerTracker.get_interaction_profile(session)` → the OpenXR path, or empty until actions have synced
- `isaacteleop.deviceio.identify_headset(profile)` → `"pico"`, `"quest"`, or `None`

## Type of change

Additive. The existing Oculus suggestion, and every current code path, behave exactly as before.

## Testing

Measured over CloudXR on real hardware, both hands, both sessions:

| Headset | Bound interaction profile |
|---|---|
| PICO 4 Ultra | `/interaction_profiles/bytedance/pico4_controller` |
| Quest 3 | `/interaction_profiles/oculus/touch_controller` |

18 unit tests cover the mapping: the two measured profiles, that they do not collapse to one answer, that a `bytedance` path can never map to `quest`, and that unknown input returns `None` — including `"quest3"` (the misleading CloudXR device-profile string) and near-misses like a path missing its leading `/interaction_profiles/`.

`clang-format-14 --dry-run -Werror`, `pre-commit`, and the mapping tests all pass.

This was arrived at empirically. A throwaway probe first queried the profile at tracker construction and returned `XR_NULL_PATH` on **both** headsets: nothing is bound until actions have synced. That is why the read happens after sync, and it is worth knowing before anyone tries to resolve this earlier.

## Design notes for reviewers

**The PICO suggestion is deliberately tolerant.** `xrSuggestInteractionProfileBindings` rejects an entire profile if any binding path is unsupported, and the binding list here is Oculus-shaped. A rejection must not take controller initialisation down with it, so it is a separate call whose failure is logged rather than thrown. It was accepted on both headsets, so this is belt-and-braces rather than a live concern.

**The profile is cached, not polled.** It is read after a successful sync and re-read only when the runtime reports `interactionProfileChanged`, which keeps it off the per-frame IPC path while still catching a headset swap mid-session.

**`get_interaction_profile` is virtual with a default, not pure**, so the replay implementation and any future one inherit `"unknown"` without modification.

**`identify_headset` never guesses.** Unrecognised input returns `None` rather than a default. Note that Quest resolves through `oculus/touch_controller`, which is the generic Meta-compatible fallback — so PICO is a positive match while Quest is identification by elimination. Enabling `XR_META_touch_controller_plus` on the instance would make Quest positive too; it was rejected here because the extension is not enabled, and it is not needed for the two devices to be told apart.

## Checklist

No `docs/` page yet; the module and tests carry their own documentation. The C++ accessor itself has no unit test — it needs a live runtime and a headset — but the code path it uses was exercised on both devices via the probe described above.

## Summary by CodeRabbit

- **New Features**
  - Added access to the controllers’ runtime interaction profile through the tracker API and Python bindings.
  - Added headset identification for supported PICO and Quest controller profiles.
  - Exposed headset identification utilities and known profile mappings through the device package.
  - Interaction profiles are refreshed during synchronization and may be unavailable before initial synchronization.
  - Unknown, missing, or unsupported profiles are handled safely without guessing.


jiwenc-nv:
Does that mean we can require `NV_DEVICE_PROFILE`? It has been a pain for users to know which one to set.

src/python/isaacteleop/cloudxr/headset.py
```diff
@@ -0,0 +1,53 @@
...
+    "/interaction_profiles/bytedance/pico_neo3_controller": "pico",
+    "/interaction_profiles/bytedance/pico_g3_controller": "pico",
+    "/interaction_profiles/meta/touch_controller_plus": "quest",
+    "/interaction_profiles/meta/touch_controller_quest_2": "quest",
+    "/interaction_profiles/facebook/touch_controller_pro": "quest",
+    "/interaction_profiles/oculus/touch_controller": "quest",
+}
+
+
+def identify_headset(interaction_profile: Optional[str]) -> Optional[str]:
```
jiwenc-nv, line 37: I better place for this will be python/isaacteleop/cloudxr

context: the deviceio module is meant to be C++ only and the python bits are its bindings.

jiwenc-nv (changes requested):

jsepulveda-nvidia:
Hey @jiwenc-nv , So I finally finished all the real robot testing just now and I'll address all these concerns.  First, in the case of NV_DEVICE, I found this to be highly unreliable and possibly either hard-coded or just not set. In all the sessions I ran the logs showed that NV_DEVICE_PROFILE=Quest3 appeared with a PICO connected, in something like 131 archived logs. It seems to be a configuration input rather than a detection output. After the partial success of this project, detection could plausibly set it — which could be for our users, but it's also a CloudXR change and I've been able to do all this work without touching CloudXR so far.

jsepulveda-nvidia:
Closing this change now in favor of https://github.com/NVIDIA/IsaacTeleop/pull/921.

Issue #880, opened 2026-08-04
Examples have no systematic test coverage; GPU example tests must be hand-registered


jiwenc-nv:
## Problem

There is no systematic way to test an example, and no convention for where an example test belongs. Examples are separate uv projects, so nothing in a normal build imports or runs them — an example can drift away from the library beside it and only fail when a user runs it. That is not hypothetical: `mcap_record_replay` and `deviceio_live_view` shipped pinned to `isaacteleop==1.3.9rc1` and every script died on `ImportError: cannot import name 'BodyJointIndex'` before it started.

## Survey of `examples/` (15 examples)

| State | Examples |
|---|---|
| Tests that run | `teleop_ros2` (C++ integration tests + topic verifier in the ROS2 CI job — the only real end-to-end coverage), `camera_viz` (3 pytest files via ctest, but behind `if(BUILD_VIZ)`), `oxr` (3 python + 2 C++ via `run_tests_with_cloudxr.sh`), `native_openxr` (`xdev_list`, same script) |
| No tests at all | `haptic_feedback`, `retargeting`, `teleop`, `teleop_session_manager`, `schemaio`, `cloudxr_mujoco_teleop`, `lerobot`, `noitom`, `oglo_tactile` |
| Not built by CMake at all | `teleop`, `cloudxr_mujoco_teleop`, `lerobot`, `noitom`, `oglo_tactile` — nothing compiles or imports them in CI |

**Orphaned tests.** `examples/oxr/python/` holds nine `test_*.py`; only three are listed in `CXR_PYTHON_GPU_TESTS`. These six are copied into the test container by `Dockerfile.test` and then never invoked:

- `test_controller_tracker.py`
- `test_full_body_tracker.py`
- `test_hand_inactive_on_plugin_stop.py`
- `test_oak_camera.py`
- `test_session_sharing.py`
- `test_synthetic_hands.py`

## Why it keeps happening

1. **GPU/runtime tests are hand-registered** in a bash array in `scripts/run_tests_with_cloudxr.sh`. Nothing discovers them, so adding a file next to nine others silently does nothing — which is how six accumulated.
2. **The container only sees one example.** `deps/cloudxr/Dockerfile.test` hardcodes `COPY examples/oxr/python/ /app/tests/`, so no other example's tests can reach the CloudXR runtime without editing the Dockerfile.
[...]

Issue #738, opened 2026-07-06
[RFC] In-headset operator feedback as a first-class concern (HUD, haptic grammar, runtime scale control) — seeking design direction


johnnynunez:
## Context

During teleoperation the operator is inside the HMD and **physically cannot read the terminal** — which is where all of the stack's warnings, harness interventions, and state changes are reported today. Over a week of hardware sessions this produced real operational failures, not just inconvenience:

- Safety-harness interventions (clamps, rejections) happened with zero operator awareness — the arm "stopped following" with no explanation, and the operator's instinct (bigger, faster motions) is exactly wrong.
- A session ran for minutes with the **wrong script variant** (no harness active); nothing in the headset indicated the difference. Diagnosed afterwards via `ps aux`.
- Workspace-edge, clutch state, and tracking-loss events were invisible at the moment they mattered.

## What we built at the example layer (works, validated in sessions)

1. **Haptic feedback grammar** — distinct vibration patterns for: workspace edge (leash engaged), clutch engage/disengage, pose-validity loss (PoseGate hold), scale change confirmation.
2. **In-headset HUD (Televiz)** — harness status, clutch state, joint-limit proximity.
3. **Pointing compass + world-anchored aim laser dot** (controller aim ray ∩ table plane) — spatial orientation without leaving the headset.
4. **Runtime motion-scale control** on the thumbstick (×0.25–×2.0, haptic confirmation) — no terminal round-trip to retune.
5. **Preflight calibration game** ("push forward" check) — doubles as automatic yaw alignment for `base_T_anchor`.

## Ask: design feedback + upstream direction

This is deliberately an RFC-style issue — we'd like the Isaac Teleop team's view on:

- Which of these belong **in the library** (e.g., a standard operator-feedback channel that retargeter nodes can emit events into, with haptic/HUD sinks) vs. staying example-level?
- Is there an existing/planned event bus in the retargeting graph that feedback primitives should attach to, so every node (rate limiter, bounds, PoseGate) can signal the operator uniformly instead of each integration reinventing it?
- Any constraints from the CloudXR/compositor side on overlay layers we should design against?

Our position from field experience: **operator feedback is a safety feature, not cosmetics** — most of our "the arm ignores me → push harder" incidents trace back to interventions the operator couldn't perceive. We're happy to upstream any/all of the above once the target architecture is agreed.

---
[...]

PR #928, opened 2026-08-08
examples/mujoco_xr: the ghost renders the governed pose, and says so

jiwenc-nv:
## Description

Stacked on #900 — **review that first; this diff is only the two commits on top.**

The ghost gripper's pose now comes from an `EePoseRateLimiter` rather than straight off the controller, so what the operator sees in the headset is *the command a follower would execute*, not where their hand is. That alone isn't perceptible: under the limits the limiter is a pass-through, and over them a clamp and a refusal both read as "the tool is behind my hand". So `InterventionMonitor` recovers the band by comparing what the limiter was handed against what it emitted, and recolours the ghost — amber clamping, red rejecting. Colour goes on the shared material (one write recolours the tool; `geom_rgba` would silently win over it) and alpha stays 1.0 in every band, because `leader_gripper.xml`'s opacity is what keeps draw order free and the ghost out of CloudXR's reprojection buffer.

Prerequisite for #738, whose thesis is that operators lose sessions to interventions they cannot perceive.

**Three calls to confirm rather than inherit.** (1) A second translucent ghost showing the raw pose would be the obvious contrast, and is wrong here: the renderer draws with `CULL_MODE_NONE`, `LEQUAL` and unconditional depth writes, so a coincident copy wins wherever the two agree — the pass-through case, i.e. exactly when nothing is wrong. (2) `SO101ClutchRetargeter` is the shipped producer of this `ee_pose` contract and is unused, because it re-bases onto a follower's base frame at every engage and a leader in the operator's hand has no home to clutch to; `GripPoseSource` adapts the controller instead. (3) **The limits are chosen for this demo, not measured against an SO-101** — 0.5 m/s and 2.5 rad/s clamp, 2.0 m/s and 10 rad/s reject, set so ordinary reaching passes through and a deliberate flick trips both bands. `RateLimiterConfig` itself defaults to 0.25 m/s, which would clamp during normal reaching and make the demo read as lag.

The first commit is **#727 cherry-picked** (5 commits, squashed), byte-identical to that branch — git's own rename detection did the rehoming from `src/retargeters/` to `src/python/isaacteleop/retargeters/`, and the `__init__.py` registration is the automerge against the current file, so `WujiHandRetargeter` survives. Squashed because three of #727's commits are web edits with no `Signed-off-by`. **If #727 lands first this will conflict**, and that ordering is yours to decide.

## Type of change

- [ ] Documentation update

## Testing

Ubuntu 22.04 / aarch64 (Tegra). `ctest -L mujoco_xr` 5/5 (new `test_harness.py`, 17 tests) and `ctest -R retargeting_` 26/26 including mypy and the 60 rehomed limiter tests.

`test_harness.py` drives the real `GripPoseSource → EePoseRateLimiter` chain at the app's own `_HARNESS` limits, so retuning them past what a hand can reach turns three tests red rather than silently producing a demo that never intervenes. It also asserts the band reaches the **`mjvGeom` rgba the Vulkan draw loop actually memcpys** rather than `mat_rgba`, since `geom_rgba` would override the material and an assertion on the model would prove nothing. It executes `_build_pipeline()` too — the graph fans the controller out to two consumers and exposes one port under two combiner names, neither obviously legal, and a connect-time rejection would otherwise surface only on a headset.

One regression this caught and fixed: under tracking loss the jaw kept following the trigger while the body froze, articulating the ghost on a stale pose. `_loop` now freezes the whole gripper when the limiter's *input* is absent, not its output.

**Not verified, and not verifiable without hardware:** whether an amber gripper reads as "the harness is holding you back" to someone wearing the headset. No `mujoco_xr` test can run in CI yet — nothing in `.github/workflows/` installs `mujoco` (#880).

## Checklist


jiwenc-nv:
this is done via #995

PR #929, opened 2026-08-08
examples/mujoco_xr: pick the gripper ghost with --robot

jiwenc-nv:
## Description

Stacked on #900 — review that first; the base here is `jiwenc/mujoco-xr-scene`, and #928 is a sibling on the same base (it changes `_update_ghost`'s signature, so one of the two will need a trivial merge).

Adds the reBot DevArm gripper beside the SO-101 leader gripper, selected with `--robot={so101,rebot}`. Everything that differs between them — scene, meshes, mocap bodies, how each moving part is driven, where it sits on the hand — moves into a new `robots.py`, so `app.py` keeps only the machinery and a moving part is a **hinge** or a **slide**. Those two cases are the whole of `_update_ghost`.

**The two ghosts are not the same kind of thing, and the placement follows.** The SO-101 leader is the device the fist is closed around, so its posture stays a headset measurement with no test pinning it. The reBot leader is a back-driven arm on a table, so what the controller commands is the *follower's* parallel jaw — nobody holds it, there is nothing to tune, and its placement is the OpenXR grip frame's own definition (ghost `+x` → grip `−Z`, the pen-tip axis; ghost `±y` → grip `±X`, so the jaws open the way the index finger squeezes) with the fist on the drive motor. A test may pin that, and does. This is the call most worth challenging.

**Two things came out of the geometry rather than a preference.** Upstream disagrees with itself on the jaw travel — `joint_left` says `0.05` and `joint_right` `0.0715` for one rack-driven pair — and the `cnc7` rail plate breaks the tie: at 0.05 the carriage's outer edge stops 3.4 mm inside the plate's end, at 0.0715 it hangs 18.1 mm past it. And the jaw origins are written as quaternions rather than copied from `rpy`, because URDF `rpy` is fixed-axis `Rz·Ry·Rx` while MuJoCo `euler=` is intrinsic `Rx·Ry·Rz`, and these have two non-zero angles. Both are checked against the fetched URDF.

**Licence wrinkle worth a decision.** Meshes come from Seeed's own `reBotArm_control_py` at a pinned, checksummed commit — the only source with both the gripper geometry and the URDF the tests read. That repository declares MIT in its `README.md` and ships **no `LICENSE` file**, so the fetch script pulls the README in its place. The hardware itself is `Seeed-Projects/reBot-DevArm`, CERN-OHL-W-2.0. Say the word if you would rather it came from somewhere else — it is one pinned line in a 70-line script.

## Type of change

- [ ] Documentation update

## Testing

Ubuntu 22.04 / aarch64 (Tegra). Full `cmake --preset py3.12 -DBUILD_VIZ=ON` build green; `ctest -L mujoco_xr` green — 4 files, 57 tests, all executed rather than skipped. The generic ghost assertions are parametrized over both robots; the reBot ones cover the URDF-derived jaw frames, the travel, the closing sweep and the placement convention. `uv pip install ./examples/mujoco_xr` was checked to carry both asset directories and the extension into the wheel.

The rail-plate assertion was checked to actually discriminate: it reads −3.39 mm at 0.05 and +18.11 mm at 0.0715, and fails at the latter. `isaiahbjork/rebot-teleop`'s independent MJCF of the same arm agrees with the derived jaw frames; nothing here depends on it.

**Not verified, and not verifiable without hardware:** everything a headset touches, unchanged from #900 — plus the reBot ghost has never been seen through one. Its placement is derived rather than measured, so what is untested is whether a 190 mm gripper reaching 250 mm out of the fist is *usable*, not whether it is where the convention says.

## Checklist


jiwenc-nv:
duplicated by https://github.com/NVIDIA/IsaacTeleop/pull/1088

PR #934, opened 2026-08-10
docs(cloudxr): say Quest3 is the WebXR profile, not a Quest restriction

jiwenc-nv:
## Description

Triage of NVBug 6582650, filed by a PICO user who read `NV_DEVICE_PROFILE=Quest3` as "my headset is being served the wrong profile". `Quest3` is in fact the profile every WebXR client uses — Quest, PICO, and the desktop-browser IWER emulator. That was already written down on main, but only inside a `.. TODO` comment, which does not render, so the reader who needs it never sees it.

This moves the fact into the prose and leaves the TODO holding just the action it proposes (rename the profile, or add a `webxr` alias in the runtime).

No behavior change; main's code and default are already correct as of 53a1d7a0. The 1.4 docs are wrong in substance and are fixed separately in #935.

Not fixed here: `scripts/setup_cloudxr_env.sh:115` writes the same `~/.cloudxr/run/cloudxr.env` with `${NV_DEVICE_PROFILE:-auto-webrtc}` — a second writer of that file with a different default. Aligning it is a behavior change on the build-from-source path, so it wants its own PR.

## Type of change

- [x] Documentation update

## Testing

Prose and one comment block; no build or runtime surface touched.

## Checklist

### DIFF
--- a/docs/source/getting_started/quick_start.rst
+++ b/docs/source/getting_started/quick_start.rst
@@ -108,16 +108,16 @@ The first launch downloads the CloudXR Web Client SDK and asks you to review and
 accept the EULA on the terminal; answer the prompt once and the acceptance is
 remembered for subsequent runs.
 
-.. TODO: ``Quest3`` is the profile every WebXR client uses — Quest, PICO, and the
-   desktop-browser IWER emulator — so the name reads as a hardware restriction it
-   does not impose. Either rename the profile or add a generic alias (e.g.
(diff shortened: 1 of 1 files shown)


jiwenc-nv:
Redundant. main already documents `Quest3` as the default as of 53a1d7a0 — the net delta here was one clarifying sentence, and the underlying naming problem is better fixed by the profile rename / `webxr` alias the TODO already tracks. The actual doc defect is on 1.4, fixed in #935.

PR #937, opened 2026-08-10
docs: read the built ref name from the command line, not the environment

jiwenc-nv:
## Description

The published 1.4 quick start tells you to install `~=1.5.0`. Found alongside NVBug 6582650.

Screenshot 2026-08-10 at 12 20 30 PM

**Why `|version|` is right on the same page and the pin is not.** Sphinx resolves `|version|` in the `DefaultSubstitutions` transform, which reads `config.version` at doctree time; sphinx-multiversion restores `version`/`release`/`today` per ref in its `config-inited` handler, so that lands correctly. Anything baked into `rst_epilog` when `conf.py` *executes* misses that entirely — and it executes from the wrong tree, because sphinx-multiversion builds every ref with `-c` pointing at the checkout it was invoked from. Only the sources come from the ref. (So fixing `release/1.4.x`'s own `conf.py` would change nothing.)

Two stale values, one line each:

- `_smv_name` read `SPHINX_MULTIVERSION_NAME`, which sphinx-multiversion 0.2.4 **never sets** — verified in its source. Always `""`, so `_VERSION_ICON_MAP` has never applied, every version linked to `client/main`, and every "Edit this page" pointed at main. The ref name does reach the build as `-D smv_current_version=`; Sphinx applies it just after `conf.py` runs, so read it off `sys.argv`.
- The install pin came from the deploy branch's `VERSION`. Take the series from the ref name instead, falling back to `VERSION` for `main` and for a plain `sphinx-build`.

Trade-off worth a look: the pin now comes from the ref *name* rather than the ref's `VERSION` file. Those agree by the convention this file already documents (`release/1.4.x` -> `1.4.x`), and `smv_branch_whitelist` only admits `main`, `release/*` and `v*` tags.

Out of scope: `release/1.0.x` emits pre-existing duplicate-target errors — its own `.rst` redefines a link target the epilog provides. Predates this change.

## Type of change

## Testing

`sphinx-build -W --keep-going` (what `make current-docs` runs) succeeds with no warnings; a plain build is unchanged.

Multiversion builds, reading the generated HTML:

| | `release/1.4.x` | `main` |
|---|---|---|
| pip pin | `~=1.4.0` | `~=1.5.0` |
| `\|version\|` | `1.4.x` | `1.5.x` |
| web client | `client/release-1.4.x` | `client/main` |
| GitHub ref | `tree/release/1.4.x` | `tree/main` |

A separate `release/1.0.x` build renders its `_VERSION_ICON_MAP` entry (CloudXR 6.1 badge, IsaacLab `develop`) for the first time.

Note for anyone re-running this locally: sphinx-multiversion prefers local refs, so a stale local `release/1.4.x` silently shadows the remote. Pass `-D smv_prefer_remote_refs=1`. CI checks out only `main` locally, so it is unaffected.

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Improved multi-version documentation builds by accepting the selected version through a command-line option.
  * Added clearer package-version compatibility handling, including validation and fallback behavior for invalid or missing version information.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -5,6 +5,7 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 
 import os
+import sys
 from datetime import datetime, timezone
 
 # -- Project information -----------------------------------------------------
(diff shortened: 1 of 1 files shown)


yanziz-nvidia (approved):

docs/source/conf.py
```diff
@@ -15,20 +16,45 @@
 copyright += f", last updated on {build_time.strftime('%B %d, %Y')}"
 author = "NVIDIA"
 
+
+def _smv_ref_name():
+    """Git ref name sphinx-multiversion is building, or "" outside it.
```
nv-jakob, line 21: Those `""` are weird?

jiwenc-nv: hmm, indeed...

PR #938, opened 2026-08-10
examples/mujoco_xr: draw with MuJoCo's renderer, read back through a PBO

farbod-nv:
The hand-written Vulkan renderer is gone. mjr_render draws into MuJoCo's offscreen framebuffer; cpp/gl_readback.cpp blits that into a sampleable pair, runs one fullscreen pass, and glReadPixels into a GL_PIXEL_PACK_BUFFER that cudaGraphicsGLRegisterBuffer imports -- which is the CUDA-linear pointer ProjectionLayer.submit() already took. Still no host copy, and now every geom type, the scene XML's materials, lights and shadows, and MuJoCo's own mesh handling. No Vulkan, no glslangValidator, no viz change.

README.md claimed this was impossible because cudaGraphicsGLRegisterImage registers no depth format and no multisampled renderbuffer. That is true, and it is the wrong entry point: RegisterBuffer has neither restriction.

Three things checked against MuJoCo 3.11.0's source rather than assumed. frustum_width is a HALF-width and IS used (render_gl3.c setView), so an asymmetric per-eye fov is expressible exactly -- mjvisualize.h calls the field "not used for rendering" and is wrong, and leaving it 0 silently switches on a viewport-aspect fallback that renders something plausible from a fov carrying nothing. mjr_render is reverse Z, so the readback shader's 1.0 - d is MuJoCo's own flipDepthIfRequired moved onto the GPU. And glBlitFramebuffer rejects a depth blit between differing formats, so the blit target is matched to whatever MuJoCo picked (DEPTH32F_STENCIL8 or DEPTH24_STENCIL8, per ARB_depth_buffer_float) rather than assumed.

tests/test_readback.py drives the real GPU path headlessly, which nothing here did before: that row 0 is the top of the operator's view, that the image is not mirrored, that the submitted depth is standard Z with the background at exactly 1.0, and that the eyes carry parallax of the right sign. It skips with a reason when there is no GPU. It caught mjr_render drawing into whatever framebuffer is bound -- leave ours bound and the next frame goes to it, with no GL error anywhere.

Two things this loses. The crease-aware vertex normals go with the renderer, and mjModel's own are smeared: measured on the pinned meshes, 11.4% / 4.5% / 1.0% / 16.6% of face corners point away from their own triangle, and render_gl3.c lights one-sided, so the ghost may render as shattered facets. test_ghost.py pins those fractions, and the fix -- if a headset says one is needed -- belongs in the asset, not in a second renderer. Separately, a multi-GPU host now needs MUJOCO_EGL_DEVICE_ID: the OpenGL context is created independently of viz's Vulkan device, so the renderer checks and names both device numbers rather than render into the wrong card's memory.

Everything downstream of the readback is still executed nowhere -- submit(), the frame loop, OpenXR session sharing, and whether the runtime accepts the depth layer -- and the example remains unconfigured in CI.

## Description

Fixes #(issue)

## Type of change

- [ ] Documentation update

## Checklist

- [ ] I have read and understood the [contribution guidelines](../CONTRIBUTING.md)


jiwenc-nv:
merged in the PR #869

PR #778, opened 2026-07-15
Add SE3 tracker device type with example controller plugin

jiwenc-nv:
Add a generic SE3 (6-DoF pose) tracker device type end-to-end:

- se3_tracker.fbs: Se3TrackerPose/Tracked/Record wrapper triple built on core.Pose, with normative conventions (units, Hamilton x,y,z,w quaternion, producer-defined reference frame with the XR session-base-space rule, and two-level validity semantics) documented in the schema itself.
- Tracker stack mirroring the joint_state pusher-fed pattern: ISe3TrackerImpl base interface, Se3Tracker facade (collection_id per instance; wire rendezvous constants TENSOR_IDENTIFIER / DEFAULT_MAX_FLATBUFFER_SIZE hosted on the facade as the single source of truth), SchemaTracker-backed LiveSe3TrackerImpl, ReplaySe3TrackerImpl with once-per-gap logging, Se3TrackerRecordingTraits, and dispatch entries in both the live and replay factories.
- Python bindings for the schema triple and the Se3Tracker facade, exported from isaacteleop.schema and isaacteleop.deviceio_trackers.
- Tests: C++ schema test with a VT/TYPE static_assert field-id freeze block, Python binding test, and a replay round-trip TEST_CASE that pins the frozen wire strings and the replay null-at-EOF contract.
- controller_se3_tracker example plugin: producer-only logical tracker that republishes an XR controller grip pose at 90 Hz, pushing is_valid=false with an identity filler pose on tracking loss, with pre-update sample-time stamping.

Design doc: /code/docs (se3-tracker-design.md, meta-repo). Build green, ctest 150/150, clang-format and pre-commit clean.

## Description

Fixes #(issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read and understood the [contribution guidelines](../CONTRIBUTING.md)
- [ ] I have run the linter and formatter with `SKIP=check-copyright-year pre-commit run --all-files`
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix/feature works (or explained why not)
- [ ] I have signed off all my commits (`git commit -s`) per the [DCO](../CONTRIBUTING.md#signing-your-work)

## Summary by CodeRabbit

* **New Features**
[...]

PR #986, opened 2026-08-20
feat(vive_se3_tracker): stream VIVE Ultimate Trackers as per-tracker SE3 poses

jenny-yu-htc:
Add a standalone plugin plus record/replay examples that stream each VIVE Ultimate Tracker as its own generic SE3 (6-DoF) pose collection (se3_tracker.fbs from PR #778) — one tensor collection per device.

Plugin (src/plugins/vive_se3_tracker):
- Reads live poses from the VIVEHub tracker_server via the VUT SDK client (on a receiver thread) and pushes Se3TrackerPose through a per-device SchemaPusher keyed on Se3Tracker::TENSOR_IDENTIFIER / DEFAULT_MAX_FLATBUFFER_SIZE.
- Timestamps pass through untranslated: VUT pose timestamps are host CLOCK_MONOTONIC, the same clock SchemaPusher documents as the local common clock, so the VUT sample time is forwarded verbatim (no push-loop bias).
- Collections are named by the tracker's physical serial number, which rides on the pose wire (VUT SDK, VIVEHub >= 1.0.1; Pose::serial), fixed on the first pose (collections can't rename); an empty serial falls back to vive_tracker_<device_id>. The serial is the stable identity; mapping a tracker to a role belongs downstream.
- Push policy per device: valid at the VUT sample time on a new sample, skip when unchanged (no duplicate timestamps in MCAP), is_valid=false + identity filler when stale.
- Advertises live collection ids to VIVE_SE3_COLLECTIONS_FILE (default /tmp/vive_se3_collections.txt) so readers auto-discover them; the tensor-list enumeration used by SchemaTracker is C++-only, so this is the Python-side discovery hook.
- Requires the VUT SDK: pass -DVUT_SDK_DIR=/path/to/VIVEHub-Linux/sdk at configure time; if unset the plugin is skipped (like the other vendor-SDK plugins), so a plain source or wheel build still configures.
- VIVE_SE3_SYNTHETIC=1 fakes trackers for a no-hardware smoke test.

Examples (examples/mcap_record_replay/python):
- record_se3_vive.py records N seconds (default 10) of every advertised collection, no arguments needed.
- replay_se3_vive.py replays the newest recording, auto-discovers collections from the MCAP channel names, derives the playback rate from the recording's own timestamps, and renders each tracker as a viser 3D coordinate frame.

Verified on real hardware (VIVEHub 1.0.1): live trackers -> per-serial collections -> se3_printer / no-arg MCAP record + replay round-trip.

## Description

Fixes #(issue)

## Type of change

- [ ] Documentation update

## Checklist

- [ ] I have read and understood the [contribution guidelines](../CONTRIBUTING.md)

## Summary by CodeRabbit

* **New Features**
  * Added support for VIVE Ultimate Tracker SE3 devices, publishing individual tracker pose streams with automatic reconnection and device discovery.
  * Added optional synthetic tracking mode for testing without hardware.
  * Added MCAP recording and replay examples with configurable playback, looping, status reporting, and optional pose visualization.
  * Added plugin installation metadata and build integration when the required SDK is available.

* **Documentation**
  * Added setup, configuration, runtime behavior, and recording/replay guidance for the VIVE SE3 Tracker plugin.

### DIFF
--- a/examples/mcap_record_replay/python/record_se3_vive.py
+++ b/examples/mcap_record_replay/python/record_se3_vive.py
@@ -1,0 +1,151 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 HTC Corporation. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Record live VIVE Ultimate Tracker SE3 pose streams to an MCAP file — headset-free.
+
+Reads the per-tracker "vive_tracker_<serial>" tensor collections pushed by the
+vive_se3_tracker plugin (one core.Se3Tracker per collection) and records each
+one to its own MCAP channel pair (<cid> / <cid>_tracked) via the standard
+Se3TrackerRecordingTraits. Replay with replay_se3_vive.py.
+
+Prerequisites (separate terminals):
+  1. CloudXR runtime:  python -m isaacteleop.cloudxr
+  2. the pusher:       ./vive_se3_tracker_plugin
+     (VIVEHub tracker_server running, or VIVE_SE3_SYNTHETIC=1 for a smoke test)
+
+With no arguments it records 10 s of every collection the pusher currently
+advertises (see --collections), to a timestamped file under ../recordings/.
+
+Usage:
+    source ~/.cloudxr/run/cloudxr.env
+    uv run record_se3_vive.py [duration_s] [output.mcap] [--collections a,b,c]
+"""
+
+import argparse
+import os
+import sys
+import time
+from datetime import datetime
+from pathlib import Path
+
+from isaacteleop.deviceio_session import DeviceIOSession, McapRecordingConfig
+from isaacteleop.deviceio_trackers import Se3Tracker
+from isaacteleop.oxr import OpenXRSession
+
+
+def _collections_file() -> str:
+    """Where the pusher advertises live collection ids (one per line).
+
+    Must mirror resolve_collections_file() in the plugin: VIVE_SE3_COLLECTIONS_FILE
+    overrides wholesale; otherwise the per-user $XDG_RUNTIME_DIR (systemd's
+    mode-0700 dir); otherwise a private per-user dir under /tmp
+    (vive_se3_tracker-<uid>) — never a fixed world-writable /tmp path.
+    """
+    env = os.environ.get("VIVE_SE3_COLLECTIONS_FILE")
+    if env:
+        return env
+    xdg = os.environ.get("XDG_RUNTIME_DIR")
+    if xdg:
+        return os.path.join(xdg, "vive_se3_collections.txt")
+    base = os.path.join("/tmp", f"vive_se3_tracker-{os.getuid()}")  # noqa: S108 - per-user 0700 dir, matches plugin
+    return os.path.join(base, "vive_se3_collections.txt")
+
+
+DEFAULT_COLLECTIONS_FILE = _collections_file()
+
+
+def discover_collections() -> list[str]:
+    """Read the collection ids the running pusher advertises, in file order."""
+    path = Path(DEFAULT_COLLECTIONS_FILE)
+    if not path.is_file():
+        return []
+    seen = {}
+    for line in path.read_text().splitlines():
+        cid = line.strip()
+        if cid:
+            seen.setdefault(cid, None)
+    return list(seen)
+
+
+def main(argv: list[str]) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "duration", nargs="?", type=float, default=10.0, help="Recording duration (s)"
+    )
+    parser.add_argument("output", nargs="?", help="Output .mcap path")
+    parser.add_argument(
+        "--collections",
+        default=None,
+        help="Comma-separated collection ids (default: auto-discover from the running pusher)",
+    )
+    args = parser.parse_args(argv[1:])
+
+    if args.output:
+        mcap_path = Path(args.output)
+        mcap_path.parent.mkdir(parents=True, exist_ok=True)
+    else:
+        out_dir = Path(__file__).resolve().parent.parent / "recordings"
+        out_dir.mkdir(exist_ok=True)
+        mcap_path = out_dir / f"se3_vive_{datetime.now():%Y%m%d_%H%M%S}.mcap"
+
+    if args.collections:
+        collections = [c.strip() for c in args.collections.split(",") if c.strip()]
+    else:
+        collections = discover_collections()
+        if not collections:
+            print(
+                "[record-se3] no live collections found "
+                f"({DEFAULT_COLLECTIONS_FILE} missing or empty). "
+                "Is the vive_se3_tracker pusher running and a tracker streaming? "
+                "Or pass --collections explicitly.",
+                file=sys.stderr,
+            )
+            return 1
+        print(f"[record-se3] discovered collections: {', '.join(collections)}")
+
+    trackers = {cid: Se3Tracker(cid) for cid in collections}
+
+    # MCAP channel base name = collection id (channels "<cid>" + "<cid>_tracked").
+    recording = McapRecordingConfig(
+        str(mcap_path), tracker_names=[(t, cid) for cid, t in trackers.items()]
+    )
+
+    print(f"[record-se3] writing {mcap_path} for {args.duration:.1f}s")
+    for cid in collections:
+        print(f"[record-se3]   collection '{cid}'")
+
+    tracker_list = list(trackers.values())
+    extensions = DeviceIOSession.get_required_extensions(tracker_list)
+    with OpenXRSession("McapSe3ViveRecord", extensions) as oxr_session:
+        with DeviceIOSession.run(
+            tracker_list, oxr_session.get_handles(), recording
+        ) as session:
+            start = time.time()
+            frame = 0
+            while time.time() - start < args.duration:
+                session.update()
+                if frame % 60 == 0:
+                    parts = []
+                    for cid, tracker in trackers.items():
+                        data = tracker.get_data(session)
+                        if data.data is None:
+                            parts.append(f"{cid}: -")
+                        elif not data.data.is_valid:
+                            parts.append(f"{cid}: lost")
+                        else:
+                            p = data.data.pose.position
+                            parts.append(f"{cid}: [{p.x:+.2f} {p.y:+.2f} {p.z:+.2f}]")
+                    print(
+                        f"[record-se3] t={time.time() - start:5.2f}s  "
+                        + "  ".join(parts)
+                    )
+                frame += 1
+                time.sleep(1 / 90)
+
+    print(f"[record-se3] done — {mcap_path}")
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv))
--- a/src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
+++ b/src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
@@ -1,0 +1,407 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 HTC Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "vive_se3_tracker_plugin.hpp"
+
+#include <deviceio_trackers/se3_tracker.hpp>
+#include <flatbuffers/flatbuffers.h>
+#include <oxr/oxr_session.hpp>
+#include <oxr_utils/os_time.hpp>
+#include <schema/se3_tracker_generated.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+#include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <unistd.h>
+#include <utility>
+#include <vector>
+
+namespace plugins
+{
+namespace vive_se3_tracker
+{
+
+namespace
+{
+
+int64_t env_int(const char* name, int64_t fallback)
+{
+    const char* v = std::getenv(name);
+    if (!v || !*v)
+        return fallback;
+    return std::strtoll(v, nullptr, 10);
+}
+
+// A directory is safe to place the advertisement file in only if it is a real
+// directory (not a symlink) that we own with no group/other write access — so no
+// other local user can pre-plant a symlink and redirect our truncating open.
+bool is_secure_dir(const std::string& dir)
+{
+    struct stat st
+    {
+    };
+    if (::lstat(dir.c_str(), &st) != 0)
+        return false;
+    if (!S_ISDIR(st.st_mode))
+        return false; // e.g. a symlink or a regular file planted at that path
+    if (st.st_uid != ::geteuid())
+        return false; // not owned by us
+    if (st.st_mode & (S_IWGRP | S_IWOTH))
+        return false; // group/other-writable
+    return true;
+}
+
+// Resolve where to write the collection-id advertisement file, securely:
+//   1. VIVE_SE3_COLLECTIONS_FILE, if set, wins verbatim (caller's responsibility).
+//   2. else $XDG_RUNTIME_DIR (systemd gives a per-user, mode-0700 dir).
+//   3. else a private /tmp/vive_se3_tracker-<uid> we create 0700 and verify we own.
+// Falls through to the plain /tmp file only if the private dir can't be secured,
+// and warns — never silently write to a world-writable fixed path.
+// Keep in sync with record_se3_vive.py.
+std::string resolve_collections_file()
+{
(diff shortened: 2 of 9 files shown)


jsepulveda-nvidia:
I've tested the new plugin and the record replay demos. It seems to be working properly. Observe these 2 recordings I made with 5 trackers placed on my wrists, ankles, and back:

https://github.com/user-attachments/assets/0314cfec-54f4-442d-a348-b9ff8af74e08


https://github.com/user-attachments/assets/294b10fd-1709-44c1-82cc-e49167b6a3f6

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
```diff
@@ -0,0 +1,333 @@
...
+        std::cout << "[vive_se3_tracker] devices reported by daemon:" << std::endl;
+        for (const auto& d : devs)
+            std::cout << "    id=" << d.device_id << " state=" << d.state << " name=\"" << d.name << "\"" << std::endl;
+    }
+}
+
+ViveSe3TrackerPlugin::~ViveSe3TrackerPlugin()
+{
+    if (vut_client_)
+        vut_client_->disconnect();
```
jsepulveda-nvidia, line 176: [polish][correctness] Destructor calls vut_client_->disconnect() while the pose callback (on_vut_pose) runs on the SDK's own receiver thread, bound via `this`, with no code-visible guarantee that disconnect() joins that thread before returning.

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
```diff
@@ -0,0 +1,333 @@
...
+        }
+        else
+        {
+            return it->second;
+        }
+    }
+
+    DeviceStream stream;
+    stream.serial = serial;
+    stream.collection_id = make_collection_id(device_id, serial);
```
jsepulveda-nvidia, line 270: [nit][input-validation] stream.collection_id (built from the tracker's raw serial straight off the VUT IPC socket) is written unsanitized, newline-delimited, into the collections advertisement file that record_se3_vive.py trusts for auto-discovery.

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
```diff
@@ -0,0 +1,333 @@
...
+        lp.pos[0] = p.position[0];
+        lp.pos[1] = p.position[1];
+        lp.pos[2] = p.position[2];
+        lp.quat[0] = p.orientation[0];
+        lp.quat[1] = p.orientation[1];
+        lp.quat[2] = p.orientation[2];
+        lp.quat[3] = p.orientation[3];
+        // Serial arrives on the pose wire (VUT SDK >= 1.0.1); NUL-terminated, "" if
+        // unknown. Fixed on the first pose — collections can't be renamed later.
+        lp.serial = p.serial;
```
jsepulveda-nvidia: [nit][defensive-coding] p.serial (a fixed char[kDeviceNameLen] from the VUT SDK) is converted to std::string via implicit strlen with no explicit bound check.

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
```diff
@@ -0,0 +1,333 @@
...
+        std::ofstream out(tmp, std::ios::trunc);
+        if (!out)
+        {
+            std::cerr << "[vive_se3_tracker] could not write collections file " << tmp << std::endl;
+            return;
+        }
+        for (const auto& kv : streams_)
+            out << kv.second.collection_id << "\n";
+    }
+    std::rename(tmp.c_str(), collections_file_.c_str());
```
jsepulveda-nvidia: [nit][error-handling]  std::rename(tmp, collections_file_) return value is unchecked in write_collections_file(), inconsistent with the error handling one line above for the ofstream open failure.

jsepulveda-nvidia (changes requested):
I put in one "polish" and 3 "nit" comments. Please take a look at the "polish comment and see if you can fix it. For the "nit" comments you can address them if you would like but they are optional, we can resolve them without changes if you prefer. Thank you!

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
jenny-yu-htc: I've checked this, and it's already safe — so **no behavior change**: `disconnect()` signals the I/O thread to stop, unblocks it via `shutdown(fd)`, and `join()`s it before returning, so the single thread that fires the callbacks has exited by the time `disconnect()` returns. We only ever call it from the main-thread destructor, never from inside a callback, so there's no self-join.
**I added a destructor comment documenting that guarantee**. On the SDK side, VIVEHub 1.0.3 will ship a `vut_client.h` that spells out the threading contract (single I/O thread; callbacks serialized; none fire after `disconnect()`/`~Client()` returns; don't block or disconnect from inside a callback), so the guarantee becomes visible at the header level rather than relying on the implementation.

However, If you'd prefer belt-and-suspenders for double protection, I'm happy to add an `atomic<bool>` shutdown guard on top — just let me know.

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
jenny-yu-htc: Added `sanitize_serial()` — keeps only `[A-Za-z0-9._-]` and drops everything else, so a malformed serial can't inject or split lines in the file `record_se3_vive.py` reads.

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
jenny-yu-htc: Same helper `sanitize_serial()` now reads via `strnlen(p.serial, sizeof(p.serial))` instead of relying on NUL-termination; verified no over-read under ASan even with a fully-populated, non-terminated buffer.

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
jenny-yu-htc: Now checks the return value, logs on failure, and cleans up the temp file — consistent with the ofstream error handling just above it.

jenny-yu-htc:
Thanks for the review! I addressed all four and replied to each comment.
Please feel free to let me know if you have any further questions.

yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
This adds per-device VIVE Ultimate Tracker SE3 streaming plus MCAP record/replay examples, and incorporates most prior feedback. Two issues remain; the failing arm64 CI job is an unrelated CloudXR termination test, with downstream jobs cancelled by the failure.

Legend: 🚫 Blocker · 💡 Suggestion · 🔍 Nit

| | Finding |
|---|---|
| 🚫 | [`src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp:245`](https://github.com/NVIDIA/IsaacTeleop/blob/5f501c214f336fca6329a65544b3605ce1c97c86/src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp#L245) — When `XDG_RUNTIME_DIR` is unset, the fixed `/tmp` temporary filename can be pre-created as a symlink and make the plugin truncate any file writable by its user; use a private mode-0700 per-user directory or securely created temporary file and keep Python discovery synchronized · previously raised |
| 💡 | [`examples/mcap_record_replay/python/replay_se3_vive.py:80`](https://github.com/NVIDIA/IsaacTeleop/blob/5f501c214f336fca6329a65544b3605ce1c97c86/examples/mcap_record_replay/python/replay_se3_vive.py#L80) — Playback consumes the coalesced `/se3_tracker_tracked` channel but derives its rate from the raw sample channel, so producer bursts containing multiple samples per recording update make replay run too fast; calculate the rate from the selected tracked channel instead |
| 🔍 | none |

**Actionables (for bots — copy-paste-ready for AI)**

> Fix if it makes sense in context — these are agent-generated suggestions, not human-vetted obligations. Skip anything that's wrong, already addressed, or not worth the churn.

- `src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp:245` — Replace the shared fixed `/tmp` fallback with secure atomic publication inside a private per-user directory, and update the Python discovery path to match.
- `examples/mcap_record_replay/python/replay_se3_vive.py:80` — Derive playback pacing from the `/se3_tracker_tracked` channel actually consumed during replay.

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
```diff
@@ -0,0 +1,286 @@
...
+namespace
+{
+
+int64_t env_int(const char* name, int64_t fallback)
+{
+    const char* v = std::getenv(name);
+    if (!v || !*v)
+        return fallback;
+    return std::strtoll(v, nullptr, 10);
+}
```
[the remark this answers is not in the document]
jenny-yu-htc: Clamped `VIVE_SE3_STALE_MS`: values that are non-numeric (parse to 0), negative, or above 1 h are rejected with a warning and fall back to the 250 ms default, so a bad env can no longer stall every sample or overflow the ns conversion.

src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp
```diff
@@ -0,0 +1,286 @@
...
+    DeviceStream stream;
+    stream.collection_id = make_collection_id(device_id, serial);
+    stream.pusher =
+        std::make_unique<core::SchemaPusher>(session_->get_handles(), make_pusher_config(stream.collection_id));
+    std::cout << "[vive_se3_tracker] created tensor collection '" << stream.collection_id
+              << "' for device_id=" << device_id << std::endl;
+    DeviceStream& ref = streams_.emplace(device_id, std::move(stream)).first->second;
+    write_collections_file(); // advertise the updated live set to readers
+    return ref;
+}
```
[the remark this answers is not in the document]
jenny-yu-htc: `DeviceStream` now remembers its serial; if a device_id reappears with a different serial, the old stream is torn down and a new collection is created for the new tracker, so samples never land in the previous tracker's collection.

examples/mcap_record_replay/python/record_se3_vive.py
```diff
@@ -0,0 +1,136 @@
...
+    """Read the collection ids the running pusher advertises, in file order."""
+    path = Path(DEFAULT_COLLECTIONS_FILE)
+    if not path.is_file():
+        return []
+    seen = {}
+    for line in path.read_text().splitlines():
+        cid = line.strip()
+        if cid:
+            seen.setdefault(cid, None)
+    return list(seen)
```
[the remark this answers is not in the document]
jenny-yu-htc: The advertisement file now lives under the per-user `$XDG_RUNTIME_DIR` (falling back to the temp dir), and both the plugin and `record_se3_vive.py` resolve the path the same way, removing the shared-directory trust issue.

jenny-yu-htc:
For the other 2 nitpick comments posted by coderabbitai, we solved them as well.

1. **Console output runs while pose_mutex_ is held.** (Trivial/ Performance & Scalability)
Moved the first-pose logging outside the `pose_mutex_` scope (the lock now only guards the pose copy), so the push thread never blocks on that I/O.

2. **Ctrl+C leaves the advertisement file on disk.** (Trivial/ Stability & Availability)
Added an atomic stop flag with `SIGINT`/`SIGTERM` handlers so the loop exits cleanly and the destructor removes the advertisement file on Ctrl+C / kill, instead of leaving a stale file behind.

jenny-yu-htc:
> Reviewed by yanziz-review-bot
> 
> **Summary** This adds per-device VIVE Ultimate Tracker SE3 streaming plus MCAP record/replay examples, and incorporates most prior feedback. Two issues remain; the failing arm64 CI job is an unrelated CloudXR termination test, with downstream jobs cancelled by the failure.
> 
> Legend: 🚫 Blocker · 💡 Suggestion · 🔍 Nit
> 
> Finding
> 🚫	[`src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp:245`](https://github.com/NVIDIA/IsaacTeleop/blob/5f501c214f336fca6329a65544b3605ce1c97c86/src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp#L245) — When `XDG_RUNTIME_DIR` is unset, the fixed `/tmp` temporary filename can be pre-created as a symlink and make the plugin truncate any file writable by its user; use a private mode-0700 per-user directory or securely created temporary file and keep Python discovery synchronized · previously raised
> 💡	[`examples/mcap_record_replay/python/replay_se3_vive.py:80`](https://github.com/NVIDIA/IsaacTeleop/blob/5f501c214f336fca6329a65544b3605ce1c97c86/examples/mcap_record_replay/python/replay_se3_vive.py#L80) — Playback consumes the coalesced `/se3_tracker_tracked` channel but derives its rate from the raw sample channel, so producer bursts containing multiple samples per recording update make replay run too fast; calculate the rate from the selected tracked channel instead
> 🔍	none
> **Actionables (for bots — copy-paste-ready for AI)**
> 
> > Fix if it makes sense in context — these are agent-generated suggestions, not human-vetted obligations. Skip anything that's wrong, already addressed, or not worth the churn.
> 
> * `src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp:245` — Replace the shared fixed `/tmp` fallback with secure atomic publication inside a private per-user directory, and update the Python discovery path to match.
> * `examples/mcap_record_replay/python/replay_se3_vive.py:80` — Derive playback pacing from the `/se3_tracker_tracked` channel actually consumed during replay.

Hi Yanzi thanks for your review, we've fixed the 2 findings you reported.

1. **When XDG_RUNTIME_DIR is unset, the fixed /tmp temporary filename can be pre-created as a symlink and make the plugin truncate any file writable by its user** (Blocker)
The fallback no longer uses a fixed name in world-writable `/tmp`. When `$XDG_RUNTIME_DIR` is unset it creates a private, mode-0700 per-user directory (`/tmp/vive_se3_tracker-<uid>`), verifies via `lstat` that it's a real directory we own (not a pre-planted symlink), and writes the advertisement file there; `record_se3_vive.py` resolves the same path. This removes the symlink-truncation vector.
2. **Playback consumes the coalesced /se3_tracker_tracked channel but derives its rate from the raw sample channel, so producer bursts containing multiple samples per recording update make replay run too fast** (Suggestion)
`capture_rate_hz` now derives the rate from the `/se3_tracker_tracked` channel that replay actually consumes (one message per recording tick), instead of the raw per-sample channel — so bursts of multiple samples per tick no longer make playback run fast.

yanziz-nvidia:
Reviewed by yanziz-review-bot

**Summary**  
The latest revision fixes playback pacing and most earlier feedback, but the insecure temporary-file fallback remains exploitable and the Python discovery path can diverge from the plugin. DCO and pre-commit pass; build jobs remain in progress.

Legend: 🚫 Blocker · 💡 Suggestion · 🔍 Nit

| | Finding |
|---|---|
| 🚫 | [`src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp:85`](https://github.com/NVIDIA/IsaacTeleop/blob/58861a8c204091d4359d8c73c401b232fadef503/src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp#L85) — After detecting that the fallback directory is unsafe, the code still returns a path through it, allowing a pre-planted directory and `.tmp` symlink to redirect the later truncating `ofstream`; fail initialization or disable advertisement instead of using the rejected path · previously raised |
| 💡 | [`examples/mcap_record_replay/python/record_se3_vive.py:48`](https://github.com/NVIDIA/IsaacTeleop/blob/58861a8c204091d4359d8c73c401b232fadef503/examples/mcap_record_replay/python/record_se3_vive.py#L48) — The reader accepts any nonempty `XDG_RUNTIME_DIR` while the plugin rejects insecure or missing directories and falls back, so no-argument discovery looks at the wrong file in that scenario; apply the same validation and fallback in both implementations |
| 🔍 | none |

**Actionables (for bots — copy-paste-ready for AI)**

> Fix if it makes sense in context — these are agent-generated suggestions, not human-vetted obligations. Skip anything that's wrong, already addressed, or not worth the churn.

- `src/plugins/vive_se3_tracker/vive_se3_tracker_plugin.cpp:85` — Never return a path through a directory that failed the security check; fail initialization or disable collection advertisement.
- `examples/mcap_record_replay/python/record_se3_vive.py:48` — Mirror the plugin’s `XDG_RUNTIME_DIR` validation and fallback exactly.

yanziz-nvidia:
@jenny-yu-htc I don't have further comments other than the bot asks for a more secured way to implement path. I am ok for this to merge.

yanziz-nvidia (approved):

jsepulveda-nvidia (approved):
Thank you for addressing all issues!

august-yao-htc (approved):

PR #995, opened 2026-08-21
Televiz: promote the robot twin and clutch preview into the library

jiwenc-nv:
## Description

One commit. The MuJoCo-backed digital twin was buried in `examples/mujoco_xr` — reachable only by installing that example, and gated on the developer having a matching `mujoco` wheel. It is a Televiz capability, not an example. The backend moves to `src/viz/robot_twin` and its Python surface to `isaacteleop.viz.robot`, both shipping in the wheel; the app left behind is renamed `examples/robot_viz` and is pure Python.

`deps/third_party` builds MuJoCo under a private name, reached through `dlopen`/`dlsym` so the extension carries no undefined `mj*` for a foreign `libmujoco` to answer. `Mujoco.cmake` beside it states that contract. Users may `pip install mujoco` at any version, or none.

The clutch preview comes with it: LeRobot's `isaac_teleop_to_so101` needs the same in-headset preview, and copying it would have duplicated the phase machine, the ghost calibration and the SO-ARM100 asset fetch. `app.py` drops from 1121 to 329 lines, and the fetch script becomes `assets.ensure_so101_scene()`, which assembles the scene into a cache directory rather than into the source tree — so a consumer outside this repo can call it. The three MJCF wrappers it reads ship as `isaacteleop.viz.robot` package data.

`OperatorFrame` measures the XR-to-robot yaw that the rebase used to assume. A wrong one reads as an intuitive-but-wrong hand-to-EE mapping rather than an error: stand 90° off the arm and pushing the controller away from yourself moves the jaw 90° off what you meant.

Two calls for a reviewer: the wheel now redistributes MuJoCo binaries, and every `BUILD_VIZ` build compiles it with no opt-out (~40 s, needs `libegl-dev`).

**No tests here** — they are #999, one commit on top of this branch. This branch registers no ctest for the twin, so it reports green by skipping; land the two together.

A review pass since the first push fixed five defects it turned up, all here: `RobotTwin.publish` declared `(joints)` while every caller passes `bodies=`/`groups=`/`materials=` and `runtime_checkable` cannot catch a signature; a runtime recentre was swallowed by `openxr_session.cpp` although `anchor.py` tells callers to re-anchor on it, which mid-engagement walked a real arm to the wrong pose once the harness stopped rejecting; the example fed a nominal `dt` to the gate's dwell while pacing on the wall clock; `TwinRunner.start()` waited unbounded, hanging `TeleopSession` forever with no headset; and the wheel shipped MuJoCo with no licence. The public surface is down from 45 exported names to the 17 with a caller, and ~930 lines of comment, docstring and README restatement are gone — no measurement or trap removed.

## Where the diff comes from

GitHub reports +5,909/−4,177. Git with rename detection (`git diff -M15% -C15%`) reports **+4,929/−3,462 — a net +1,467 lines of repository.** The gap is files that were moved *and* rewritten in the same commit; GitHub's 50%-similarity default will not pair them (`app.py` → `clutch_preview.py` matches at 20%, `follower.py` → `preview_arm.py` at 21%).

By folder, with each moved file counted once against its destination:

| Folder | Files | Moved | Diff | What it is |
|---|---:|---:|---|---|
| `src/python/isaacteleop/viz/robot/` | 17 | 3 | +2,558 / −1,720 | The promoted Python API. The 3 moved are `app.py` → `clutch_preview.py`, `follower.py` → `preview_arm.py`, `harness.py`; the other 14 are split out of those two, so much of the +2,558 is the same code redistributed |
| `src/viz/robot_twin/cpp/` | 20 | 11 | +1,054 / −272 | The backend, moved from `examples/mujoco_xr/cpp/`. Changed: include paths, and MuJoCo reached by `dlsym` rather than a link. New: `gl_context.cpp` (+230), the headless EGL context |
| `examples/mujoco_xr/` | 5 | 0 | −1,264 | Deleted outright. Nothing to review |
| `examples/robot_viz/` | 5 | 2 | +614 / −8 | The example, renamed. `app.py` 1,121 → 329 lines, README 817 → 246 |
| `src/python/isaacteleop/teleop_session_manager/` | 3 | 0 | +467 | `TwinRunner`, `TwinRenderConfig`, and `TeleopSession` owning the render thread |
| `deps/third_party/` | 2 | 0 | +99 | MuJoCo fetch, private SONAME, licence staging |
| `src/viz/session/`, `src/viz/xr/` | 7 | 0 | +38 / −3 | `XR_TYPE_EVENT_DATA_REFERENCE_SPACE_CHANGE_PENDING` surfaced to Python |
| `src/python/isaacteleop/retargeters/` | 3 | 1 | +29 / −141 | `ControllerPoseSource` |
| repo root, `cmake/`, `src/core/python/` | 7 | 0 | +36 / −18 | Wheel package-data, REUSE, build deps |
| `src/python/isaacteleop/viz/robot/assets/` | 3 | 3 | +20 / −19 | The three MJCF wrappers, now package data |
| `rigs/`, `docs/` | 4 | 1 | +14 / −17 | |

The first two rows are ~80% of the diff and hold everything worth reviewing; rows three onward are a deletion, a rename, or build glue.

## Type of change

- [ ] Documentation update

`isaacteleop_examples.mujoco_xr` is gone; the docs that referenced it are updated.

## Testing

Verified through #999, which is this branch plus its tests: 22 test files, every one green on Linux aarch64 (Jetson AGX Orin), Release, CUDA, Python 3.12 — including `ensure_so101_scene()` fetching the SO-101 assets and `test_engage_sequence.py` driving the engage handshake against a real `SceneTwin`.

After the review fixes: 201 passed, 23 skipped, on Linux aarch64 (Jetson AGX Orin), Python 3.12. `test_symbol_isolation.py` passes with a foreign `mujoco` in the same process, which is the packaging claim.

`readelf` on the built artifacts confirms the isolation: one export, no `libmujoco` or EGL `NEEDED`, private SONAME, `SYMBOLIC` set.

Writing those tests turned up a packaging bug that is fixed here: `src/python/CMakeLists.txt` globbed only `isaacteleop/*.py`, and `[tool.setuptools.package-data]` for `isaacteleop.viz.robot` listed only `*.so`/`*.pyd`/`*.pyi`. The three MJCF wrappers under `viz/robot/assets/` therefore reached neither the staged `python_package/` nor the wheel, so `ensure_so101_scene()` raised `FileNotFoundError` anywhere but a source checkout. Both are widened; the assets are staged and packaged.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added `robot_viz`, a pure-Python SO-101 robot visualization and CloudXR teleoperation example.
  * Added robot-twin rendering with headless EGL/OpenGL support, stereo XR views, depth output, and GPU selection.
  * Added cached, checksum-verified scene asset downloads.
  * Added controller pose selection, clutch engagement gating, pose monitoring, and retargeting utilities.

* **Changes**
  * MuJoCo XR example and standalone package were removed and replaced by `robot_viz`.
  * Updated build requirements and documentation for EGL and the new example.


src/viz/CMakeLists.txt
```diff
@@ -31,6 +31,24 @@ cmake_minimum_required(VERSION 3.20)
...
 
+# Robot twin: MuJoCo's renderer read back into CUDA, over a MuJoCo of its own that the
+# rest of the process cannot see. It links no viz target and meets the compositor through
+# __cuda_array_interface__ alone, so it is a sub-module here only by where it ships.
+#
+# There is no option for it: a Televiz build has it. Linux only, and that is a fact
+# rather than a knob -- gl_context.cpp makes its OpenGL context through EGL's device
+# platform, which is how it renders with no display, and Windows has no equivalent yet.
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    # MuJoCo itself. Included here rather than at the root so the decision and the
```
jiwenc-nv: why don't we add it even deeper?  such as inside `robot_twin`.

Also, why would robot_twin_tests needs mujoco direclty?  I think it should just test the interfaces of robot_twin, not its mujoco internals.

.gitignore
```diff
@@ -73,11 +73,7 @@ deps/v2d/wheels/
...
 # authored wrapper XML is tracked.
-#
-# Keep this rule HERE, not in examples/mujoco_xr/.gitignore: scikit-build-core
-# resolves .gitignore against the project root, so a rule there would strip the
-# meshes out of the wheel too.
-/examples/mujoco_xr/python/isaacteleop_examples/mujoco_xr/assets/leader/*
-!/examples/mujoco_xr/python/isaacteleop_examples/mujoco_xr/assets/leader/leader_gripper.xml
-/examples/mujoco_xr/python/isaacteleop_examples/mujoco_xr/assets/follower/*
-!/examples/mujoco_xr/python/isaacteleop_examples/mujoco_xr/assets/follower/follower_arm.xml
+/examples/robot_viz/python/isaacteleop_examples/robot_viz/assets/leader/*
```
jiwenc-nv: where does those xml come from?  mujoco?  if so, can we just download them?

CMakeLists.txt
```diff
@@ -172,7 +172,11 @@ if(BUILD_EXAMPLES)
     add_subdirectory(examples/haptic_feedback)
     if(BUILD_VIZ)
         add_subdirectory(examples/camera_viz/tests)
-        add_subdirectory(examples/mujoco_xr)
+    endif()
+    # Pure Python, but its tests import the robot twin, which ships with Televiz on
+    # Linux (src/viz/CMakeLists.txt says why not on Windows).
+    if(BUILD_VIZ AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
jiwenc-nv: this check against `CMAKE_SYSTEM_NAME STREQUAL "Linux"` is YAGNI.  Fix or prove me wrong.

deps/third_party/CMakeLists.txt
```diff
@@ -202,3 +202,49 @@ if(BUILD_VIZ)
     FetchContent_MakeAvailable(glfw)
     message(STATUS "GLFW 3.4 fetched")
 endif()
+
+# ==============================================================================
+# MuJoCo (robot twin physics + renderer)
```
jiwenc-nv: do we really use any physics from MuJoCo at all?  The use case of MuJoCo here in this PR should be visualization only.  Please double check whether we are doing physics at all.  That's YAGNI, if so, please remove.

deps/third_party/CMakeLists.txt
```diff
@@ -202,3 +202,49 @@ if(BUILD_VIZ)
...
+    # has, which is the whole point; bumping it is an isaacteleop release decision.
+    set(ISAACTELEOP_MUJOCO_VERSION 3.11.0
+        CACHE INTERNAL "MuJoCo version the robot twin is built against")
+
+    message(STATUS "Fetching MuJoCo ${ISAACTELEOP_MUJOCO_VERSION} from GitHub...")
+    FetchContent_Declare(
+        mujoco
+        GIT_REPOSITORY https://github.com/google-deepmind/mujoco.git
+        GIT_TAG        ${ISAACTELEOP_MUJOCO_VERSION}
+        # The ref is a tag, so it is fetchable without history. Measured here: 140 MB of
```
jiwenc-nv: too verbose.  no need to comment `GIT_SHALLOW` remove.

deps/third_party/CMakeLists.txt
```diff
@@ -202,3 +202,49 @@ if(BUILD_VIZ)
...
+        # SHA, which is why the entries above set it that way.
+        #
+        # Not the order-of-magnitude win it looks like, because CMake spells GIT_SHALLOW
+        # as `--depth 1 --no-single-branch` -- depth 1 of EVERY branch. A release tarball
+        # (URL + URL_HASH) would carry no history at all; it is the bigger change and has
+        # not been measured on a fast link.
+        GIT_SHALLOW    TRUE
+    )
+
+    # Plain variables, not the `CACHE ... FORCE` the entries above use: MuJoCo sets
```
jiwenc-nv: check: should those go into mujoco.cmake?  or it is required to configure before `FetchContent_MakeAvailable(mujoco)`

docs/source/getting_started/build_from_source/index.rst
```diff
@@ -47,6 +47,14 @@ Prerequisites
    ``pip install isaacteleop`` already ships the compiled ``isaacteleop.viz`` module. See
    `Other Build options`_ for the full option table.
 
+   On **Linux**, ``BUILD_VIZ=ON`` also builds the robot twin's scene backend
```
jiwenc-nv: YAGNI.  Please remove.

docs/source/getting_started/build_from_source/index.rst
```diff
@@ -236,7 +244,7 @@ The CMake options (defined in root :code-file:`CMakeLists.txt` and :code-file:`c
      - ``ON`` on Linux
    * - **Televiz visualization**
      - ``BUILD_VIZ``
-     - Auto: ``ON`` when Vulkan and the CUDA Toolkit are detected, else ``OFF``. Force with ``-DBUILD_VIZ=ON`` / ``-DBUILD_VIZ=OFF``. (Most users don't need this — ``pip install isaacteleop`` already ships the compiled ``isaacteleop.viz`` module.)
+     - Auto: ``ON`` when Vulkan and the CUDA Toolkit are detected, else ``OFF``. Force with ``-DBUILD_VIZ=ON`` / ``-DBUILD_VIZ=OFF``. On Linux it also builds the robot twin's scene backend, which needs ``libegl-dev`` and fetches MuJoCo on the first configure. (Most users don't need this — ``pip install isaacteleop`` already ships the compiled ``isaacteleop.viz`` module.)
```
jiwenc-nv: YAGNI.  Please remove.

src/viz/CMakeLists.txt
jiwenc-nv: (claude: partly addressed.)

**The `Linux` guard and the fetch are both gone from this file** — `bce10b785` moved MuJoCo to `deps/third_party/`, which is where every other `FetchContent` in this repo lives (glm, GLFW, pybind11, Catch2). Your later comments at `deps/third_party/CMakeLists.txt:207/226/237` reviewed it in that location, so I've read those as accepting the placement. Say the word if you'd still rather it sit under `src/viz/robot_twin/` and I'll move it.

**On the tests: agreed, and the coupling is gone.** `ISAACTELEOP_MUJOCO_VERSION` existed as a `CACHE INTERNAL` variable *only* so `robot_twin_tests` could tell our copy from the wheel's, and `Mujoco.cmake`'s header named `test_symbol_isolation.py`. Both removed. #999's tests exercise `_robot_twin`'s Python surface; the one thing they legitimately assert about MuJoCo is the isolation itself — one export, no `libmujoco` NEEDED, private SONAME — and they read that off the built artifact, not off a CMake variable.

Also folded away `src/viz/robot_twin/CMakeLists.txt`, whose entire body was `add_subdirectory(cpp)`.

.gitignore
jiwenc-nv: (claude: no change — I think the premise doesn't hold; happy to be told otherwise.)

They're not MuJoCo's, and everything that *can* be downloaded already is.

The two tracked files — `leader/leader_gripper.xml` and `follower/follower_arm.xml` — are **ours**: hand-authored NVIDIA-copyright MJCF wrappers carrying the mesh transforms, the `leader_ghost` material, the geom-group policy and the mm→m scale. There's nothing upstream to fetch them from.

Everything upstream *is* fetched: `scripts/fetch-so-arm.sh` pulls 20 files from `TheRobotStudio/SO-ARM100` at a pinned commit with a per-file sha256 — 18 STLs plus upstream's own `so101_new_calib.xml` and `.urdf` — all gitignored. That's exactly what these `.gitignore` rules exist for: keep the fetched assets out of git, keep our two wrappers in.

I can reword the comment from "the authored wrapper XML" to "the two wrappers authored here" if that reads less ambiguously.

CMakeLists.txt
jiwenc-nv: (claude: this is addressed.)

That hunk no longer exists. The guard wrapped `add_subdirectory(examples/robot_viz/tests)`; the tests moved to #999 and the example is pure Python, so it isn't added from CMake at all now. The `if(BUILD_VIZ)` block above it is unchanged from `main`.

deps/third_party/CMakeLists.txt
jiwenc-nv: (claude: this is addressed — you were right, and it was worse than the header comment suggested.)

**The call: yes we were, and it's gone.** `Scene::forward()` called `mj_forward` — the *full* forward dynamics: collision detection, constraint solve, CRB inertia, actuation, sensors — once per rendered frame, and nothing reads any of it. It's now `mj_kinematics` + `mj_camlight`, which is precisely the position stage `mjv_updateScene` needs (body/geom/site transforms, camera and light poses). `mj_forward` is off `mj_functions.inc` entirely, so the module no longer even resolves the symbol.

Verified on the shipped scene: 6 hinges, 5 of 10 bodies move on a joint publish, all transforms finite. `ctest`: 316/316.

One caveat worth recording: with only the position stage, a scene using **tendons** or **rangefinder sensors** would render those wrong — both default to visible in `mjv_defaultOption`, and both need stages that no longer run. Ours has neither, and `_read_joint_map` already refuses any joint that isn't a hinge. #999's `test_readback.py` is what would catch a pixel-level regression.

**The build: no, and it can't be.** Upstream ships one `libmujoco` — engine, renderer and MJCF parser in a single target, no renderer-only option — so the ~40 s is the whole library either way. Header comment now says "scene + renderer" and states that outright instead of claiming physics.

deps/third_party/CMakeLists.txt
jiwenc-nv: (claude: this is addressed.) All 8 lines gone — `GIT_SHALLOW TRUE` is now bare, matching the glm and GLFW entries immediately above it.

deps/third_party/CMakeLists.txt
jiwenc-nv: (claude: this is addressed.) The latter — they *must* precede it, so they can't move into `Mujoco.cmake`.

`FetchContent_MakeAvailable(mujoco)` is what runs MuJoCo's own `CMakeLists.txt`, and that's where its `option()` calls happen; `Mujoco.cmake` is included *after*, and only touches the `mujoco` target, which doesn't exist until `MakeAvailable` returns. The comment now leads with that constraint instead of burying it under the CMP0077 explanation.

docs/source/getting_started/build_from_source/index.rst
jiwenc-nv: (claude: this is addressed.) All 8 lines removed.

docs/source/getting_started/build_from_source/index.rst
jiwenc-nv: (claude: this is addressed.) Sentence removed; the row is back to its wording on `main`.

The one doc change I've kept in this file is `libegl-dev` in the `apt-get install` line — without it the configure fails, and `CheckBuildDeps.cmake` names the package in its error message.

farbod-nv (approved):
televiz changes LGTM

PR #996, opened 2026-08-21
docs: prefer one build directory over one per Python version; retire the orphaned mujoco_xr tests

jiwenc-nv:
## Description

Rebased onto `main`, and the shape changed. 7513288d0 deleted `examples/mujoco_xr` outright, taking the `cmake --preset py3.12` and `build/cmake-cpython-312` text this branch was named for — `git grep -E 'cmake --preset|cmake-cpython'` on `main` is now empty. Two commits remain.

**1. Build-directory advice.** `cmake/SetupPython.cmake`'s error and two build pages still tell readers to keep a `build-py3.12/` tree beside `build/`, though `build/` is the convention everywhere else and nothing in the tree supports the parallel layout. They now say to delete and re-configure.

**2. Retire the orphaned `mujoco_xr` tests.** The promotion left `tests/python/examples/mujoco_xr/` behind, and nothing has run it since: `mujoco_xr_py` no longer exists, so its `CMakeLists.txt` returns early and registers no ctest entry, and `conftest.py` resolves a `sys.path` entry under the deleted `examples/mujoco_xr/python`.

Worth flagging on its own: #999 would have ported `test_frames`, `test_projection` and `test_readback` to the twin, and it was closed unmerged — so the robot twin has no coverage on `main`. This PR does not change that, it only stops the tree from implying otherwise. Reviving #999 is the real fix.

## Type of change

- [x] Documentation update

## Testing

`SKIP=check-copyright-year pre-commit run` is clean over the changed files. The deletion is self-verifying: the removed suite registered no ctest entry before the change, so `ctest -L mujoco_xr` is empty on both sides.

## Checklist


ivany-nv:
**Verdict: the documentation fixes are right, but three of the eight files now edit content `main` deleted, and the stale references the PR targets are already gone.** A rebase would show how much is still needed.

The presets really are gone: `CMakePresets.json` was removed in `f8b5eaa59`, and `py3.11` / `py3.12` / `py3.13` exist nowhere on `main` (the surviving `src/plugins/manus/CMakePresets.json` is unrelated). Every replacement command the docs now recommend resolves against real options, paths, and labels on `main`.

Two things have overtaken it:

1. **Three of eight files edit deleted content.** `7513288d0` deleted `examples/mujoco_xr/` wholesale, giving modify/delete conflicts on that example's `CMakeLists.txt`, `README.md`, and `__init__.py`. The other five auto-merge, with the tests `pyproject.toml` following its rename.
2. **`git grep -e '--preset' origin/main` is now empty** — the stale references this PR set out to remove are already gone (6 of them existed at the merge base).

Two small content points on the parts still worth landing:

- `src/python/isaacteleop/rig/launcher.py`: the rewritten comment names only `cmake -B build` and the managed wheel trees (`build-wheel/cpython-311`), but the tuple on the next line is still `("build", "build-cmake", "build-wheel")`, so the comment and the code disagree slightly. `build-cmake` was never a real layout — presets used `build/cmake-cpython-3XX` — so the entry is dead either way, and it is harmless since the comment says it only sharpens an error message.
- `examples/mujoco_xr/tests/pyproject.toml:14-15`: the comment says the ABI match comes from "CMakeLists.txt passing the build interpreter to `uv run --python`". It passes a version string — `tests/CMakeLists.txt:23-24` uses `uv run --python ${ISAAC_TELEOP_PYTHON_VERSION}`, i.e. `3.11` — so uv selects or downloads some 3.11 interpreter rather than the build venv. Same CPython minor, so the outcome is ABI-adequate; only the stated mechanism differs.

Given the above, it might be simplest to rebase and see what remains — by my count the four files that still have live targets.

aristarkhovNV (approved):

PR #997, opened 2026-08-21
ROS: add Wuji hand retargeting

sgrizan-nv:
## Description

Support Wuji gloves in controller and hand teleoperation while preserving existing retargeter behavior and documenting external plugin setup.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] Documentation update

## Summary by CodeRabbit

- **New Features**
  - Added Wuji hand retargeting support for ROS2 teleoperation.
  - Added Wuji hand models, joint mappings, wrist-source selection, calibration guidance, and topic documentation.
  - Added retargeter-specific controller and wrist end-effector profiles.
  - Added validation for Wuji joint names and published values.

- **Bug Fixes**
  - Improved scenario-specific replay verification and startup error reporting.
  - Refined controller-to-hand transform handling across teleoperation modes.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -698,10 +698,20 @@ jobs:
           --entrypoint bash teleop_ros2_ref:${{ env.ROS_DISTRO }} -c \
           "source /usr/local/bin/teleop-env-setup && exec /opt/isaacteleop/install/examples/teleop_ros2/cpp/integration_tests/teleop_ros2_mcap_generator '$REPLAY_MCAP'"
 
-        for mode in controller_teleop hand_teleop controller_raw full_body; do
-          echo "Testing mode: $mode"
-          LOG_FILE="${LOG_FILE_BASE}_${mode}.log"
-          RUN_NAME="teleop-ros2-${PROJECT_NAME}-${mode//_/-}"
+        scenarios=(
+          "controller_teleop mode_default"
+          "hand_teleop mode_default"
+          "controller_raw mode_default"
+          "full_body mode_default"
+          "controller_teleop wuji"
+          "hand_teleop wuji"
+        )
+        for scenario in "${scenarios[@]}"; do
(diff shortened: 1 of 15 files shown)


lotusl-code (approved):

lotusl-code:
1. [High] Wuji joint validity is discarded before ROS publication
The new graph connects only `hand_joints` and drops `hand_valid in session_config.py

2. [High] an untracked wrist can be published as a valid world-frame EE pose in wuji_glove_plugin.cpp

3. [Medium] the Wuji integration verifier uses an invalid numeric heuristic
teleop_ros2_topic_verifier.py#L281-L288 contradicts the retargeter contract: a valid flat-open hand can produce all zeros. It can therefore reject correct output. Conversely, because left and right positions are checked together, one nonzero side can hide an invalid all-zero side.

4. [Medium] the mode table contradicts the implemented Wuji profile
The prose correctly states that Wuji supplies hand poses, EE poses, and wrist TFs in controller mode in examples/teleop_ros2/README.md#L56-61 . But the mode table instead says EE poses and TFs come from controller aim poses and that hand is published only for Sharpa in examples/teleop_ros2/README.md#187-191

5. [Low] changed Python unit tests are not registered with project CI
test_teleop_profiles is not registered in examples/teleop_ros2/CMakeLists.txt

sgrizan-nv:
#1, #3: this is WAI. The Joint Command message in ROS2 doesn't contain validity flags; instead, downstream ROS2 nodes treat zero joint commands as invalid and ignore them

#2: This cannot be fixed reliably inside the ROS 2 node because the tracking state has already been discarded before HandsSource produces its output. Gating on controller validity would also be incorrect because the plugin may use optical tracking, and hand_teleop does not require controllers.
Both reliable remedies are outside this PR’s ROS-node scope: extending the core hand schema/source to preserve TRACKED, or correcting the Wuji plugin so the wrist is not marked valid when no absolute wrist source exists.

#4: It doesn't state that

#5: Fixed

PR #1005, opened 2026-08-24
build: check that docs do not reference code that is not there

ivany-nv:
Two classes of doc rot ship silently today. This adds one pre-commit hook that catches both.

## What goes wrong now

**Dead links.** The `:code-file:` / `:code-dir:` roles (`docs/source/conf.py`) build a `github.com` URL from whatever string they are handed and validate nothing. Move a file and the docs ship a live 404 while the Sphinx build stays green. 150 such references existed with nothing checking any of them — #1003 was five of them, found this way.

**Examples that raise on paste.** The `teleopcore` → `isaacteleop` rename left documented imports pointing at a package that does not exist (#1004). Nothing looks inside documented code, so nothing caught it.

## What the hook does

1. `:code-file:` / `:code-dir:` targets must exist, **with the shape the role implies** — a `:code-dir:` aimed at a file is reported as a kind mismatch, not as "missing". That check is what caught the two CloudXR links fixed in the first commit here.
2. `isaacteleop.*` imports inside documented Python (reST `code-block:: python` and Markdown ```python) must resolve under `src/python/`. Module paths only, not symbols — `isaacteleop.retargeters` uses a lazy export table a symbol check would have to interpret.

Prose and shell paths are **deliberately out of scope**. A sweep of those was mostly false positives (build artifacts, `cd`-relative paths, other repos), and a committer-facing gate has to be right every time to stay worth having.

## Allowlist

`KNOWN_BROKEN` ships empty. `KNOWN_BROKEN_IMPORTS` carries one entry — `teleop_session.rst:39`, whose example is broken four ways and needs an author rewrite, not an import swap. Entries carry their reason, and `--all` reports any that start resolving **or** that no doc references any more, so the list cannot outlive its problem. That second check already paid for itself: it is what flagged the five entries #1003 resolved, in the last commit here.

## Verification

Run under Python 3.12.13, matching `setup-python` in the pre-commit workflow:

```
full sweep, clean tree                        exit 0 (100 files, 150 roles, 50 blocks)
absent target                                 exit 1
:code-dir: at a real file                     exit 1, "use :code-file:"
:code-file: at a real directory               exit 1, "use :code-dir:"
unresolvable documented import                exit 1, correct file:line
all hooks x whole tree (CI's own default)      all pass, no file rewritten
```

+297/-2 across 4 files. No runtime code touched.

## Summary by CodeRabbit

* **Documentation**
  * Corrected source directory references in build-from-source and WebXR documentation.
  * Improved accuracy when directing readers to SDK and WebXR source locations.

* **Quality Improvements**
  * Added automated validation for documentation references and code examples.
  * Checks now identify missing or mismatched references before documentation changes are committed.

### DIFF
--- a/docs/source/getting_started/build_from_source/index.rst
+++ b/docs/source/getting_started/build_from_source/index.rst
@@ -122,7 +122,7 @@ tarballs such as:
   those platforms.
 - ``nvidia-cloudxr-<version-for-web-sdk>.tgz`` (CloudXR Web SDK)
 
-You can place them in the :code-file:`deps/cloudxr/` directory and update the ``deps/cloudxr/.env``
+You can place them in the :code-dir:`deps/cloudxr/` directory and update the ``deps/cloudxr/.env``
 file to locally override the default version defined in :code-file:`deps/cloudxr/.env.default`,
 like this:
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,3 +59,11 @@ repos:
         # COMMIT_EDITMSG path (outside the work tree), which makes the script's
         # `git check-attr` fail with "outside repository" (exit 128).
         stages: [pre-commit]
+      - id: check-doc-refs
+        name: "Check :code-file:/:code-dir: targets exist in the tree"
+        entry: python3 scripts/check_doc_refs.py
+        language: system
+        # The roles in docs/source/conf.py build a github.com URL from the target
+        # without validating it, so a moved file ships a live 404 and the Sphinx
+        # build still passes. Only .rst carries these roles.
+        files: ^docs/source/.*\.rst$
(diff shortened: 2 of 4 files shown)


docs/source/getting_started/build_from_source/index.rst
```diff
@@ -122,7 +122,7 @@ tarballs such as:
   those platforms.
 - ``nvidia-cloudxr-<version-for-web-sdk>.tgz`` (CloudXR Web SDK)
 
-You can place them in the :code-file:`deps/cloudxr/` directory and update the ``deps/cloudxr/.env``
```
jiwenc-nv: interesting that the `:code-file:` also works.

jiwenc-nv (approved):

PR #1009, opened 2026-08-24
Fix: stale API usage in the teleop session doc

xsun2445:
## Description

The Python blocks in `docs/source/getting_started/teleop_session.rst` predated the single-sided `GripperRetargeter` and the `TensorGroup` input contract, so several of them raised on the first call rather than being merely imprecise. A reader copying the Quick Start got an `ImportError`; a reader copying the "Before" block got three different exceptions in sequence.

**Quick Start**
- `GripperRetargeter` moved to `isaacteleop.retargeters` and now requires a `GripperRetargeterConfig`
- one retargeter handles one hand and emits `gripper_command`, not `gripper_left` / `gripper_right`

**Before vs After**
- `oxr.OpenXRSession.create(...)` does not exist — construct the class directly
- the source owns its `ControllerTracker`, so take it from `controllers.get_tracker()`; a standalone `deviceio.ControllerTracker()` is never registered with the session and raises `RuntimeError: Tracker implementation not found`
- pipeline inputs must be `TensorGroup`s — `poll_tracker()` replaces the hand-built dict of lists, which raised `AttributeError: 'list' object has no attribute 'group_type'`

**Also**
- `OpenXRSessionHandles` import path: `teleopcore.oxr` -> `isaacteleop.oxr` (`teleopcore` is not importable)
- corrected the gripper example path and added the two-sided example reference
- added the missing `teleop_control_pipeline` and `sinks` fields to the `TeleopSessionConfig` listing

## Type of change

- [x] Documentation update

## Testing

Every code block was extracted into a scratch harness and executed against a live
CloudXR session (CloudXR Runtime 6.3.0, `Quest3` device profile, web-client emulator),
rather than only checked against the source. Each block ran with a bounded loop and
exited 0.

Not covered: the `pm.PluginManager([...])` / `plugin_manager.start(...)` lines in the
"Before" block remain `...` placeholders in a deliberately schematic section, and were
not executed.

Platform: Ubuntu, Python 3.11.15, isaacteleop 1.5.115rc1.

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Updated the teleoperation quick-start guide with the current two-sided gripper setup.
  - Clarified optional control pipeline and sink configuration for teleoperation sessions.
  - Updated OpenXR import instructions and complete-example references.
  - Refreshed setup examples to include combined gripper commands and tracker polling.

### DIFF
--- a/docs/source/getting_started/teleop_session.rst
+++ b/docs/source/getting_started/teleop_session.rst
@@ -36,14 +36,18 @@ Here's a minimal example:
        TeleopSessionConfig,
    )
    from isaacteleop.retargeting_engine.deviceio_source_nodes import ControllersSource
-   from isaacteleop.retargeting_engine.examples import GripperRetargeter
+   from isaacteleop.retargeters import (
+       GripperRetargeter,
+       GripperRetargeterConfig,
+   )
 
    # Create source and build pipeline
    controllers = ControllersSource(name="controllers")
-   gripper = GripperRetargeter(name="gripper")
-   pipeline = gripper.connect({
-       "controller_left": controllers.output("controller_left"),
-       "controller_right": controllers.output("controller_right")
+
+   # A gripper retargeter handles one hand; build the left one here
+   gripper_left = GripperRetargeter(GripperRetargeterConfig(hand_side="left"), name="gripper_left")
+   pipeline = gripper_left.connect({
+       ControllersSource.LEFT: controllers.output(ControllersSource.LEFT),
    })
 
    # Configure session
@@ -58,8 +62,7 @@ Here's a minimal example:
        while True:
            result = session.step()
            # Access outputs
-           left = result["gripper_left"][0]
-           right = result["gripper_right"][0]
+           left = result["gripper_command"][0]
 
 Configuration Classes
 ---------------------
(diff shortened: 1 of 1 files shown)


ivany-nv:
**Verdict: the premise is sound and every symbol checks out, but four of the five edits already landed on `main`, and the one thing I would fix first is a line that auto-merges into an inconsistent state without appearing in any conflict marker.**

The premise is right: every old form this PR replaces is genuinely absent from `main` — `isaacteleop.retargeting_engine.examples`, `teleopcore.oxr`, `oxr.OpenXRSession.create(...)`, and single-argument `GripperRetargeter(name=...)` (the real signature at `gripper_retargeter.py:51` takes `config` as a required positional). Every symbol the PR introduces does exist.

Four of the five edits landed on `main` in `8278e3fd9`, `42a1c63aa`, and `77de60476`, and `main`'s versions went further — a two-sided gripper pipeline where this PR has left only.

**The merge edge case is the part I would look at first.** The merge produces three conflict hunks, but the Quick Start's output line lands *outside* them:

```
line  90  [outside conflict markers]  left = result["gripper_command"][0]
line  67  [main's side of hunk 2]     "gripper_left":  connected_left.output("gripper_command"),
line  68  [main's side of hunk 2]     "gripper_right": connected_right.output("gripper_command"),
line 590  [outside conflict markers]  left_gripper = result["gripper_left"][0]
```

`main`'s side routes through `OutputCombiner`, which renames the outputs to `gripper_left` / `gripper_right`. Resolving in `main`'s favour — the natural choice — leaves line 90 reading a key the pipeline does not produce, while line 590 of the same file reads `main`'s key. Nothing in the conflict markers points a resolver at line 90.

For clarity, `main` is self-consistent today: it reads `result["gripper_left"]` and `result["gripper_right"]`, matching its own `OutputCombiner`. This PR replaces those two lines with the single `gripper_command` read because it went one-sided, and since `main` never touched those lines they merge in silently while the pipeline above them takes `main`'s side.

Two edits still look worth landing, both independent of the above:

- `teleop_session.rst:94-103` really does omit `teleop_control_pipeline` and `sinks`, which exist on the dataclass in exactly the order the PR uses.
- Replacing the hand-written raw-dict injection with `controllers.poll_tracker(...)` — `poll_tracker` returns precisely the keys the manual dict built, and this hunk merges cleanly.

A rebase would make it clear how much of the rest is still needed.

PR #1013, opened 2026-08-25
fix: enforce the DCO identity match AGENTS.md already promises

ivany-nv:
AGENTS.md says the sign-off name and e-mail "must match the committer's git identity" and that "a `commit-msg` pre-commit hook enforces this".

It does not. `check_dco_signoff.py` matched `^Signed-off-by: \S.*<\S+>` and stopped there — it never read git config, and the identity requirement appeared only inside its own error text. `dco.yml` is the same shape: present, well-formed, no identity comparison.

So this passed everywhere:

```
Signed-off-by: Anybody <a@b.c>
```

That matters because a sign-off certifies that **you** have the right to submit the work. A trailer naming someone else certifies nothing.

## What changed

At least one `Signed-off-by` must match the committer. **At least one, not all** — relaying a patch keeps the original author's sign-off and adds yours, and that stays legal.

Compared against `GIT_COMMITTER_IDENT`, not `git config user.name`, because that is what `--signoff` itself uses: it honours the `GIT_COMMITTER_*` environment overrides the config cannot see, so the two can disagree and only one is the identity that lands in the commit. E-mail compares case-insensitively; the name as written.

Failure prints your identity next to what it found, plus `git commit --amend -s`.

## Tested

| | |
|---|---|
| own sign-off | pass |
| relayed sign-off + own | pass |
| e-mail in different case | pass |
| no trailer | fail |
| someone else's trailer alone | fail |
| wrong name | fail |
| wrong e-mail | fail |
| `GIT_COMMITTER_NAME` override | picked up, where `git config` would have missed it |

## Scope

This closes the local half only. `dco.yml` still accepts any well-formed trailer, so a `--no-verify` commit reaches CI unchallenged. Worth a follow-up if we want the guarantee end to end — happy to do it in a separate PR, since it changes what CI rejects.

1 file, no behaviour change for anyone already using `git commit -s`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sign-off validation to require an exact contributor name match and a case-insensitive email match.
  * Added clearer diagnostics when sign-off information is missing or does not match.
  * Improved handling of unavailable or malformed identity information while validating commits.

### DIFF
--- a/scripts/check_dco_signoff.py
+++ b/scripts/check_dco_signoff.py
@@ -1,41 +1,107 @@
 #!/usr/bin/env python3
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Commit-msg hook: reject commits that lack a DCO Signed-off-by line."""
+"""Commit-msg hook: require a DCO Signed-off-by that matches the committer.
+
+A sign-off certifies that *you* have the right to submit the work, so a trailer
+naming someone else certifies nothing. Presence alone was checked before this;
+the identity half was documented in AGENTS.md but never enforced.
+
+Extra trailers are fine — relaying a patch keeps the original author's sign-off
+and adds yours — so the rule is that at least one matches, not that all do.
+"""
+
+from __future__ import annotations
 
 import re
+import subprocess
 import sys
 
-# Matches a real git trailer line: "Signed-off-by: Name <email>" at the start
-# of a line, with at least one non-whitespace character before the angle bracket.
-_TRAILER_RE = re.compile(r"^Signed-off-by: \S.*<\S+>", re.MULTILINE)
+# A real git trailer: "Signed-off-by: Name <email>" at the start of a line, with
+# at least one non-whitespace character before the angle bracket.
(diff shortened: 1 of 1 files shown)


ivany-nv:
Closing this — the rule is wrong for how this repo is actually used, and I only found that after checking the change against real history rather than synthetic cases.

Across the last 6 months of non-merge commits on main, only 26% carry a sign-off matching the author. Of the mismatches, the common shape is not abuse:

```
author=farbod-nv <fmotlagh@nvidia.com>
signed=Farbod Motlagh <fmotlagh@nvidia.com>      same address, git name is the GitHub handle

author=rwiltz <...+rwiltz@users.noreply.github.com>
signed=Rafael Wiltz <rwiltz@nvidia.com>          GitHub noreply vs corporate address
```

Both are arguably *more* DCO-correct than the git identity is — the DCO asks for a real legal name, and git author names are frequently the GitHub handle. A strict name comparison would have blocked at least three contributors mid-workflow, which is a worse outcome than the gap it closes.

The underlying finding stands: `AGENTS.md` states the name and e-mail "must match the committer's git identity" and that a commit-msg hook "enforces this", and neither the hook nor `dco.yml` compares anything. That needs either a check that fits real usage (e-mail only? noreply mapping? warn rather than block?) or a doc that stops promising enforcement. Taking it to an issue first, since it is a policy call about what CI rejects rather than a bug fix.

PR #1015, opened 2026-08-25
Add keyboard SE3/SE2 device: schema, tracker, plugin, retargeters

rwiltz:
## Description

Ports Isaac Lab's Se2Keyboard/Se3Keyboard off raw carb/kit keyboard events onto the IsaacTeleop session API.

- keyboard.fbs + (codegen'd) KeyboardTracker: raw evdev key-press state (pressed_keys: [ushort]), no semantic mapping.
- keyboard plugin: standalone Linux evdev reader, self-discovers its device under /dev/input/by-path/, auto-launched by PluginManager.
- KeyboardSource (IDeviceIOSource) exposing "keyboard" (fixed 13-key SE3 subset) and "keyboard_all_keys" (256-entry bitmap covering every standard key) -- keys are reported as-is; retargeters own all key-to-action mapping.
- KeyboardToSe3RelRetargeter + KeyboardGripperRetargeter (WASDQE/ZXTGCV -> EE delta, K-key toggle -> gripper) and KeyboardToSe2Retargeter (arrow/numpad -> base velocity), matching the legacy Se3Keyboard/Se2Keyboard key bindings.

Fixes #(issue)

## Type of change

- [ ] Documentation update

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added Linux keyboard input support with automatic device discovery and standalone operation.
  - Exposes pressed-key states through device I/O and schema APIs.
  - Added movement-control outputs and full key bitmaps.
  - Added keyboard retargeters for SE(3) motion, gripper toggling, and SE(2) base velocity commands.
  - Supports configurable motion sensitivities and keyboard-driven gripper control.
- **Documentation**
  - Added setup, permissions, supported keys, usage, and autorepeat guidance.
- **Tests**
  - Added coverage for keyboard tracking, source outputs, schemas, motion, gripper, and base controls.


examples/teleop/python/keyboard_printer_example.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+PLUGIN_ROOT_DIR = Path(__file__).resolve().parent.parent.parent.parent / "plugins"
+PLUGIN_NAME = "keyboard"
+PLUGIN_ROOT_ID = "keyboard"
+
+# Evdev key codes (linux/input-event-codes.h) -> display name, for the keys on a
+# standard PC keyboard. Codes not listed here still show up (as "code<N>") --
+# this table is for readability, not a completeness gate.
+KEY_NAMES = {
```
nv-jakob, line 33: Should the plugin or the example be using https://github.com/xkbcommon/libxkbcommon to translate the keycodes instead?

rwiltz: Short answer: no, and I'd keep it that way, but it's a fair question that deserves the actual reasoning, not just "it's intentional."
  
  What libxkbcommon actually does: takes a raw scancode + the active keymap (layout, modifier state, dead keys) and resolves it to a keysym — "what character/symbol does this physically produce right
  now." That's a fundamentally different question from what this plugin answers, which is "is this physical key currently held."

  Why raw evdev codes are the right layer here, not a limitation:
  - The SE3 bindings (WASDQE for translation, arrows/numpad for base velocity, K for gripper) are physical-position bindings, the same convention every game uses for movement controls. A French AZERTY
    user's "W" key is in the same physical spot as a US QWERTY "W" — it just prints a different letter. If we resolved through a keymap, the same physical key would silently rebind to a different KEY_*
    code depending on what layout happened to be active when the teleop session started, which is exactly backwards for a control scheme meant to be muscle-memory-stable regardless of layout.
  - The wire schema (KeyboardOutput.pressed_keys), the tracker, the source node, and every retargeter are built around evdev key codes end to end. Switching to keysyms wouldn't be a local fix in the
    plugin — it'd mean threading layout-resolution through the whole chain, and redesigning the retargeter bindings as "the key that types W" instead of "physical position of W," which is a real semantic
    change, not a translation-layer swap.
    
  Where it would matter, and why it's not worth it even there: the only place a keymap actually helps is display — showing an operator "you're holding the key labeled Ä" instead of "code30" if they're on
  a non-US layout. That's what the printer's KEY_NAMES table is for, and it's cosmetic. Pulling in libxkbcommon for that means a new build dependency and solving keymap discovery for a context where
  there's no compositor to query it from — this plugin runs as a standalone binary reading /dev/input/eventN directly, headless, with no X/Wayland session to ask "what's the active layout." You'd have to
  hardcode an assumed layout or read raw XKB config files yourself, which is more complexity than the cosmetic benefit justifies.

examples/teleop/python/keyboard_printer_example.py
```diff
@@ -0,0 +1,185 @@
...
+from isaacteleop.cloudxr import CloudXRLauncher
+from isaacteleop.retargeting_engine.deviceio_source_nodes import KeyboardSource
+from isaacteleop.teleop_session_manager import (
+    TeleopSession,
+    TeleopSessionConfig,
+    PluginConfig,
+)
+
+
+PLUGIN_ROOT_DIR = Path(__file__).resolve().parent.parent.parent.parent / "plugins"
```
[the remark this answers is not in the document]
rwiltz: Confirmed the underlying issue, but the suggested fix doesn't exist on this branch main has no wheel-bundling story at all yet for any plugin (verified against pre-existing plugins like generic_3axis_pedal too the only mechanism is a bare install() rule requiring an explicit cmake --install step, landing outside the wheel's package tree). This PLUGIN_ROOT_DIR computation is copied verbatim from every other example script in this directory, a pre-existing repo-wide convention, not something this PR introduced. Building real wheel-bundling for main is a bigger, separate effort affecting every plugin. Left as-is to match convention; happy to file a follow-up issue for the broader gap.

src/python/isaacteleop/retargeting_engine/tensor_types/standard_types.py
```diff
@@ -299,6 +299,49 @@ def RobotHandJoints(name: str, joint_names: list[str]) -> TensorGroupType:
...
+            BoolType("keyboard_key_d"),
+            BoolType("keyboard_key_q"),
+            BoolType("keyboard_key_e"),
+            BoolType("keyboard_key_z"),
+            BoolType("keyboard_key_x"),
+            BoolType("keyboard_key_t"),
+            BoolType("keyboard_key_g"),
+            BoolType("keyboard_key_c"),
+            BoolType("keyboard_key_v"),
+            BoolType("keyboard_key_k"),
```
aristarkhovNV: why just these specific keys versus a universal all keys keyboard?

src/core/schema/fbs/keyboard.fbs
```diff
@@ -0,0 +1,32 @@
...
+    // Whether the tracker has emitted at least one sample.
+    is_valid: bool (id: 1);
+}
+
+// MCAP recording wrapper for KeyboardOutput.
+//
+// Record types are the root types written to MCAP channels by the McapRecorder.
+// Trackers serialize into Record types via their serialize() method, but the
+// public query API returns the inner data type directly.
+table KeyboardOutputRecord {
```
aristarkhovNV, line 23: implementation notes do not belong here (i.e. how exactly trackers compose the buffer)

src/python/isaacteleop/retargeting_engine/tensor_types/standard_types.py
rwiltz: Good call, I thought I asked claude to use an enum instead. updating.

src/core/schema/fbs/keyboard.fbs
rwiltz: Fixed across all 6 PRs

examples/teleop/python/keyboard_printer_example.py
jiwenc-nv: +1 to @rwiltz's answer.   But we are OSS, so people who want it can either file an issue or propose a PR.

that being said, @ivany-nv we may want to beef up languages in https://github.com/NVIDIA/IsaacTeleop?tab=contributing-ov-file to encourage contributions.

examples/teleop/python/keyboard_printer_example.py
```diff
@@ -0,0 +1,185 @@
...
+    103: "UP",
+    104: "PAGEUP",
+    105: "LEFT",
+    106: "RIGHT",
+    107: "END",
+    108: "DOWN",
+    109: "PAGEDOWN",
+    110: "INSERT",
+    111: "DELETE",
+    **{2 + i: str((i + 1) % 10) for i in range(10)},  # KEY_1..KEY_0 -> "1".."9","0"
```
jiwenc-nv, line 58: nit: wondering whether really worth saving the lines...

src/core/schema/fbs/keyboard.fbs
```diff
@@ -0,0 +1,28 @@
...
+// (see linux/input-event-codes.h, e.g. KEY_W = 17). Carries no semantic mapping --
+// which keys mean what (an axis, a toggle, a mode switch) is entirely up to the
+// consuming retargeter.
+//
+// All fields are always present whenever this table itself is present.
+table KeyboardOutput {
+    // Evdev key codes currently held down, in no particular order.
+    pressed_keys: [ushort] (id: 0);
+
+    // Whether the tracker has emitted at least one sample.
```
jiwenc-nv, line 18: is this a duplicate of `pressed_keys` being an empty list.

src/core/schema/python/keyboard_bindings.h
```diff
@@ -0,0 +1,58 @@
...
+#include <cstdint>
+#include <string>
+#include <vector>
+
+namespace py = pybind11;
+
+namespace core
+{
+
+inline void bind_keyboard(py::module& m)
```
jiwenc-nv, line 23: @aristarkhovNV do we still need this binding?

jiwenc-nv (approved):
mostly nits. but i want to confirm that whether we still really need to the python binding.

src/core/schema/python/keyboard_bindings.h
aristarkhovNV: yes

examples/teleop/python/keyboard_printer_example.py
aristarkhovNV: could this use a reverse mapping of `EvdevKeyCode` with something like
```py
CODE_TO_NAME = {member.value: member.name for member in EvdevKeyCode}
```

aristarkhovNV (approved):

PR #1029, opened 2026-08-26
ROS: gate finger joint publishing on hand_valid

somudro:
## Problem

`WujiHandRetargeter` reports tracking validity separately from the joint values,
because an all-zero pose is also a legitimate flat-open hand. Its docstring says
so explicitly:

> Consumers deciding whether to forward commands to hardware must check this
> rather than inspect the joint values (an all-zero pose is also a legitimate
> result for a flat-open hand).

`examples/retargeting/python/wuji_hand_retargeter_demo.py` does exactly that.
The ROS example did not: `hand_valid` was never wired into the session result,
so on tracking loss `xr_teleop/finger_joints` carried the zero pose and
downstream controllers drove the hand open rather than holding.

Consumers cannot compensate for this downstream, because the flag never reaches
ROS at all and the zeros are indistinguishable from a real flat hand.

## Change

Surface `hand_valid` where a retargeter provides it, and omit that side from the
message when it reports untracked. Controllers that hold the last commanded
position for absent joints then hold through a dropout instead of snapping the
hand open.

Gating is per side, so one glove dropping out does not freeze both hands, and
no message is published when neither side is tracked.

Retargeters with no `hand_valid` output pass `None` and are treated as always
valid, so DexPilot, Pink IK and TriHand behaviour is unchanged.

## Tests

Four cases added to `tests/python/examples/teleop_ros2/test_messages.py`:
untracked side omitted, both sides kept when tracked, no message when neither is
tracked, and the no-validity path preserved for retargeters that do not report
it.

## Context

Found while wiring Wuji glove teleoperation of a simulated Wuji Hand 2 in Isaac
ROS (ISAACOS-1062), where a dropout mid-grasp visibly flung the simulated hand
open.

## Summary by CodeRabbit

* **New Features**
  * Added per-hand tracking validity support for finger-joint publishing.
  * Hand data continues publishing independently when only one hand is tracked.
  * Missing tracking-validity information preserves existing publishing behavior.

* **Bug Fixes**
  * Prevents untracked hand data from being published.
  * Suppresses output when neither hand has valid tracking.

* **Tests**
  * Added coverage for tracked, untracked, missing-validity, and no-hand scenarios.

### DIFF
--- a/examples/teleop_ros2/python/session_config.py
+++ b/examples/teleop_ros2/python/session_config.py
@@ -50,7 +50,10 @@ from isaacteleop.teleop_session_manager import (
     TeleopSessionConfig,
 )
 from node_parameters import NodeParameters
-from teleop_ros2_retargeters import JointNameAliasRetargeter
+from teleop_ros2_retargeters import (
+    HandTrackingGateRetargeter,
+    JointNameAliasRetargeter,
+)
 from tensor_group_helpers import joint_names_from_group_type
 
 
@@ -72,6 +75,37 @@ def _maybe_alias_hand_joints(
         {"hand_joints": connected_hand_retargeter.output("hand_joints")}
     )
     return alias_connected.output("hand_joints")
+
+
+def _maybe_gate_hand_joints_on_tracking(
+    hand_retargeter,
+    connected_hand_retargeter,
+    finger_joints,
+    output_joint_names,
+    name: str,
+):
+    """Gate a side's hand joints on tracking validity, where it is reported.
+
+    Retargeters that report validity separately get a HandTrackingGateRetargeter
+    inserted, and the gated ``hand_joints`` output is returned. Retargeters
+    without a ``hand_valid`` output get ``finger_joints`` back unchanged, so they
+    stay connected directly and this remains invisible to the publisher and to
+    profile validation.
+    """
+    if "hand_valid" not in hand_retargeter.output_spec():
+        return finger_joints
+
+    joint_names = output_joint_names or joint_names_from_group_type(
+        hand_retargeter.output_spec()["hand_joints"]
+    )
+    gate = HandTrackingGateRetargeter(joint_names=joint_names, name=name)
+    gate_connected = gate.connect(
+        {
+            "hand_joints": finger_joints,
+            "hand_valid": connected_hand_retargeter.output("hand_valid"),
+        }
+    )
+    return gate_connected.output("hand_joints")
 
 
 def _resolve_hand_tracking_plugin_configs(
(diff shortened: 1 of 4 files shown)


examples/teleop_ros2/python/session_config.py
```diff
@@ -386,7 +398,29 @@ def build_tracked_hand_finger_joint_outputs(
...
-    return left_finger_joints, right_finger_joints
+    # Only some retargeters report tracking validity separately. Where they do,
+    # surface it: an all-zero pose is also a legitimate flat-open hand, so the
+    # joint values alone cannot tell a consumer that tracking was lost.
+    left_valid = maybe_hand_valid(left_hand_retargeter, left_hand_connected)
+    right_valid = maybe_hand_valid(right_hand_retargeter, right_hand_connected)
+    return left_finger_joints, right_finger_joints, left_valid, right_valid
+
+
+def hand_valid_outputs(left_valid, right_valid) -> dict:
```
sgrizan-nv: Could we make both of these helpers private and name them after what they return? `maybe_hand_valid` sounds like a boolean predicate, but it returns an optional connected pipeline output, while `hand_valid_outputs` builds a mapping of pipeline outputs. Names such as `_get_optional_hand_valid_output` and `_build_hand_valid_pipeline_outputs` would make their roles clearer.

examples/teleop_ros2/python/session_config.py
```diff
@@ -386,7 +398,29 @@ def build_tracked_hand_finger_joint_outputs(
...
+    """Pipeline entries for whichever sides report tracking validity."""
+    outputs = {}
+    if left_valid is not None:
+        outputs["finger_joints_left_valid"] = left_valid
+    if right_valid is not None:
+        outputs["finger_joints_right_valid"] = right_valid
+    return outputs
```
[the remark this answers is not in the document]
sgrizan-nv: Could we keep `hand_valid` internal to session construction instead of exposing it as additional `SessionResult` keys?

One approach would be to add a small package-local retargeter that accepts `hand_joints` and `hand_valid`, passes the joints through when tracking is valid, and emits an absent `OptionalTensorGroup` otherwise. This gate would only be inserted for retargeters that expose `hand_valid`; other retargeters would remain connected directly.

The existing message builder already omits absent sides, so this would preserve the current `finger_joints_left`/`finger_joints_right` contract, keep the ROS node unaware of retargeter-specific capabilities, and avoid changing profile validation. Would this be a cleaner separation?

examples/teleop_ros2/python/session_config.py
```diff
@@ -74,6 +77,37 @@ def _maybe_alias_hand_joints(
     return alias_connected.output("hand_joints")
 
 
+def _build_tracking_gated_hand_joints_output(
```
sgrizan-nv: nit: Would `_maybe_gate_hand_joints_on_tracking` fit better here? Unlike the `build_*` helpers, this function does not always construct a new graph component - it returns `finger_joints` unchanged when the retargeter has no `hand_valid` output. A `maybe_*` name would also align with the adjacent `_maybe_alias_hand_joints()` helper.

sgrizan-nv (approved):

Issue #985, opened 2026-08-20
examples: use proper pyproject.toml setup for all python ones


jiwenc-nv:
Partner feedback on #736 (item 5): the Physical AI Developer team copied example code into their own project, hit import failures, and patched their copies with dual-mode import guards. Worth settling a convention rather than leaving each example to pick one.

Two conventions coexist under `examples/`:

- `mujoco_xr` — its own wheel, `isaacteleop_examples.mujoco_xr` namespace, relative imports, documented as `python -m`. Copies cleanly.
- `camera_viz` and `teleop_ros2/python` — 21 files import siblings as *top-level* modules: `from pipeline import ...`, `from sources import ...`, `from messages import ...`, `from geometry import ...`. These resolve only because the script's own directory lands on `sys.path`.

So the second group breaks under `python -m`, breaks the moment a file is copied into a package (what the partner hit), and squats on very generic top-level names — `pipeline`, `sources`, `messages`, `constants`, `assets`. `examples/teleop_ros2/python/pyproject.toml` already carries `pythonpath = ["."]` to keep pytest working around it.

Suggested direction: make both directories real packages with relative imports plus a `-m` entry point, matching `mujoco_xr`, and write the convention down for examples. I'd avoid adopting `try: from .x / except ImportError: from x` guards repo-wide — that doubles every import site to paper over an invocation mistake.

Cross-repo note: `huggingface/lerobot`'s `examples/isaac_teleop_to_so101/` uses `files(__package__)` (`common.py:76`) and relative imports. Its README documents `python -m` only, so it works as specified; recording it here for traceability.

jiwenc-nv:
Remaining review feedback on #1038, after ivany-nv's [verification pass](https://github.com/NVIDIA/IsaacTeleop/pull/1038#issuecomment-5597514904):

- [x] Root `.gitignore` didn't cover `./recordings/` when run from the repo root — fixed in 887db467f
- [x] `--host 0.0.0.0` flip was bundled into the packaging commit — split out in 681bc222b
- [x] Auto-discovery picked the newest `.mcap` regardless of type — type-prefixed in 0981f53ff
- [x] `rigs/full_body.yaml`'s recording directory disagrees with the new replay default — decoupled in 588223708 (rig is slated for deprecation, not worth reconciling)
- [x] Usage lines, `--help` text, and error messages still said `python replay_full_body.py ...` instead of the `-m` form — fixed in 73ed7fe00
[...]

PR #1034, opened 2026-08-28
examples(deviceio_live_view): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985; #1035 has landed. Split into three commits per review — the two preparation pieces stand on their own and can be read (or landed) without the conversion.

**1. `cmake: let install_python_example() take either example layout`** — `examples/README.md` already tells authors to put `pyproject.toml` at the example root and call `install_python_example(DESTINATION examples/<name>)`, but the macro only read `python/pyproject.toml`, so that call installed nothing usable. It now picks the layout by where `pyproject.toml` sits, which decides the `find-links` depth and whether `python/` is installed as a directory or flattened. No example uses the converted layout at this commit, so it is a no-op; the legacy branch goes away with the last conversion.

**2. `examples: bind viser viewers to every interface by default`** — @ivany-nv was right that a one-example flip was wrong, so this goes the other way and moves all eight. These viewers run where the headset is and get opened from a laptop elsewhere on the network; `127.0.0.1` means the one machine that cannot practically open the browser is the only one allowed to. The startup line now reports the address it actually bound, which was already wrong whenever `--host` was passed. The `# noqa: S104` is gone — ruff selects no `S` rules here. `--host 127.0.0.1` restores the old behaviour.

**3. `examples(deviceio_live_view): package under isaacteleop_examples`** — the conversion itself. `live_deviceio.py` imported its sibling as a top-level module, which resolved only because the invoked script's directory landed on `sys.path`; that breaks under `python -m`, breaks when the files are copied into another project — what the Physical AI Developer team hit — and squats the bare name `deviceio_viser`. Plus the grid fix: it lay in viser's default XY plane, which stands up as a wall under `set_up_direction("+y")`.

```bash
uv pip install -e ./examples/deviceio_live_view
python -m isaacteleop_examples.deviceio_live_view
```

Still open from review: `install_python_example()` does not install the example-root `README.md`, which #1044 / #1036 / #1046 need. Happy to add the `install(FILES ... OPTIONAL)` to commit 1 — say the word.

## Type of change

- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. `uv build --wheel ./examples/deviceio_live_view` ships `isaacteleop_examples/deviceio_live_view/{__init__,__main__,deviceio_viser,live_deviceio}.py` and nothing else — no namespace `__init__.py`, no bare top-level name. `SKIP=check-copyright-year pre-commit run` clean on every touched file; all nine example scripts byte-compile.

The macro body is byte-identical to the previously approved `a916fca56` (that diff is comments only), so @ivany-nv's install-tree verification still applies. Not re-run: the full CMake install tree, and a live headset run.

## Checklist

<sub>No new tests: this is a move plus a default flip, and neither example has automated coverage today (`TODO(#880)`). Verification is the wheel-layout check above plus a human run on a headset.</sub>

### DIFF
--- a/cmake/InstallPythonExample.cmake
+++ b/cmake/InstallPythonExample.cmake
@@ -11,8 +11,14 @@
 # way it pins isaacteleop to the version this build produced and drops any
 # [tool.uv.sources], so the installed example resolves the wheel next to it.
 #
+# Two source layouts are accepted while the examples are converted to the
+# convention in examples/README.md (NVIDIA/IsaacTeleop#985), picked by where
+# pyproject.toml sits. Converted: at the example root, sources under python/.
+# Legacy: under python/. Drop the legacy branch once every example has moved.
+#
 # Usage:
-#   install_python_example(DESTINATION examples/oxr/python)
+#   install_python_example(DESTINATION examples/<name>)          # converted
+#   install_python_example(DESTINATION examples/<name>/python)   # legacy
 #   install_python_example(DESTINATION examples/teleop_ros2/python
 #       EXTRA_UV_EXTRA_BUILD_DEPS "nlopt = [\"numpy\"]")
 # ==============================================================================
@@ -32,8 +38,17 @@ macro(install_python_example)
     endif()
 
     # Read the bare pyproject.toml and append uv configuration for the
-    # installed environment.
-    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/python/pyproject.toml" _PYPROJECT_BASE)
+    # installed environment. `find-links` is relative to the installed
+    # pyproject, so its depth follows the layout: <prefix>/examples/<name>/ for
+    # a converted example, one level deeper for a legacy one.
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/pyproject.toml")
+        set(_IPE_PYPROJECT "${CMAKE_CURRENT_SOURCE_DIR}/pyproject.toml")
+        set(_IPE_FIND_LINKS "../../wheels")
+    else()
+        set(_IPE_PYPROJECT "${CMAKE_CURRENT_SOURCE_DIR}/python/pyproject.toml")
+        set(_IPE_FIND_LINKS "../../../wheels")
+    endif()
+    file(READ "${_IPE_PYPROJECT}" _PYPROJECT_BASE)
 
     # A source-tree [tool.uv.sources] path is relative to the pyproject, so from
     # the installed copy it points at the install prefix, which is not a Python
@@ -47,7 +62,7 @@ macro(install_python_example)
     string(REGEX REPLACE "\"(isaacteleop(\\[[^]]*\\])?)\""
         "\"\\1==${ISAAC_TELEOP_PYPROJECT_VERSION}\"" _PYPROJECT_BASE "${_PYPROJECT_BASE}")
     set(_TOOL_UV_BLOCK "[tool.uv]
-find-links = [\"../../../wheels\"]
+find-links = [\"${_IPE_FIND_LINKS}\"]
 python-preference = \"only-managed\"
 environments = [\"python_version == '${ISAAC_TELEOP_PYTHON_VERSION}'\"]
 required-environments = [\"sys_platform == 'linux' and platform_machine == '${_IPE_PLATFORM_MACHINE}'\"]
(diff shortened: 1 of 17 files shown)


examples/deviceio_live_view/python/isaacteleop_examples/deviceio_live_view/live_deviceio.py
```diff
@@ -9,44 +9,44 @@
...
 )
 
 
 def main(argv: list[str]) -> int:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument(
         "--host",
-        default="127.0.0.1",
-        help="Viser HTTP bind address (default: 127.0.0.1; pass 0.0.0.0 to expose externally)",
+        default="0.0.0.0",  # noqa: S104 - a dev viewer you open from another machine
```
ivany-nv: This default flip doesn't belong in a packaging PR, and it makes this the only viser example that binds all interfaces — the eight in `mcap_record_replay` all default to `127.0.0.1` with the identical help string, and this `# noqa: S104` is the only one in the repo.

If the intent is reaching the viewer from another machine, that's what `--host 0.0.0.0` is for; flipping the default silences the warning for everyone to save a flag for one workflow. Suggest reverting this and keeping `setup_scene` — that part is a clear improvement.

ivany-nv (approved):
Built and installed the full CMake install tree on x86_64 Linux, default prefix — the case the description lists as not covered. Re-verified on `e9b6e9b9`. Layout, generated pyproject, and `uv run python -m isaacteleop_examples.deviceio_live_view --help` from the install tree all check out: no doubled `python/python/`, no stray `isaacteleop_examples/__init__.py`, `find-links` resolves to `<prefix>/wheels` with the built wheel, `isaacteleop` pinned and resolved from there rather than an index, legacy examples install unchanged, `clang_format_check` and pre-commit clean.

One fix needed, in the macro: the example-root `README.md` is not installed, since `install_python_example()` only installs `python/`. No-op for `deviceio_live_view`, but three downstream examples ship `python/README.md` into the install tree today and lose it on conversion:

- #1044 — `rename from examples/haptic_feedback/python/README.md to examples/haptic_feedback/README.md`
- #1036 — `examples/teleop_session_manager/python/README.md` deleted
- #1046 — `examples/oxr/python/README.md` deleted

```cmake
install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/README.md"
    DESTINATION ${_IPE_DESTINATION} OPTIONAL)
```

Here or in #1036, your call.

ivany-nv (approved):
Re-approving for `f8239a89`. Re-ran the install-tree verification on the new head: layout, generated pyproject, `uv run python -m isaacteleop_examples.deviceio_live_view --help` from the install tree, `clang_format_check` and pre-commit all still clean. The macro is byte-identical, and the delta since `e9b6e9b9` is runtime-only (`GroundGrid`), so nothing in the packaging conclusion changes.

The two open items from the previous review still stand: the example-root `README.md` is still not installed, and the `--host` default flip is still there.

cmake/InstallPythonExample.cmake
```diff
@@ -32,8 +39,17 @@ macro(install_python_example)
...
+    # pyproject, so its depth follows the layout: <prefix>/examples/<name>/ for
+    # a converted example, one level deeper for a legacy one.
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/pyproject.toml")
+        set(_IPE_PYPROJECT "${CMAKE_CURRENT_SOURCE_DIR}/pyproject.toml")
+        set(_IPE_FIND_LINKS "../../wheels")
+    else()
+        set(_IPE_PYPROJECT "${CMAKE_CURRENT_SOURCE_DIR}/python/pyproject.toml")
+        set(_IPE_FIND_LINKS "../../../wheels")
+    endif()
+    file(READ "${_IPE_PYPROJECT}" _PYPROJECT_BASE)
```
nv-jakob, line 51: Should be a separate preparation commit / MR?

cmake/InstallPythonExample.cmake
```diff
@@ -66,8 +82,17 @@ ${_IPE_EXTRA_UV_EXTRA_BUILD_DEPS}
...
+    # directory, so the namespace root in the install tree matches the source
+    # tree; a legacy one flattens python/ into the destination.
+    # `python` without a trailing slash installs the directory itself, so the
+    # namespace root survives; `python/` installs its contents, flattening it.
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/pyproject.toml")
+        set(_IPE_SOURCE_DIR "python")
+    else()
+        set(_IPE_SOURCE_DIR "python/")
+    endif()
+    install(DIRECTORY ${_IPE_SOURCE_DIR}
```
nv-jakob, line 94: Ditto on the preparation

cmake/InstallPythonExample.cmake
jiwenc-nv: right.  put that into a separate commit now.

cmake/InstallPythonExample.cmake
jiwenc-nv: ditto :)

examples/deviceio_live_view/python/isaacteleop_examples/deviceio_live_view/live_deviceio.py
jiwenc-nv: moved into a separate commit.

PR #1035, opened 2026-08-28
examples: write down the Python example convention

jiwenc-nv:
## Description

Part of #985. First in a stack of 14 — the rest build on this.

Examples are split between two import styles and two `pyproject.toml` locations, so "what does a new example look like" has no answer, and copying one into your own project is a coin flip on whether its imports survive. This writes down the convention `examples/mujoco_xr` already follows and the rest are being moved to: one distribution per example, sources under `python/isaacteleop_examples/<name>/`, relative sibling imports, `isaacteleop_examples/` kept as a PEP 420 namespace, and one command that works for every example.

It also fixes two references that outlived what they named, both found while auditing for this:

- `gripper_retargeting_simple.py` moved directory *and* was renamed; the docs still pointed at the old path.
- `examples/cxrjs` was deleted in `a61389d27` and `b6e490af8`; its `REUSE.toml` annotation stayed behind. `reuse-lint-file` lints files, so an annotation naming a path that no longer exists never fails — which is why it survived six months.

No code changes; no example moves in this PR.

## Type of change

- [x] Documentation update

## Testing

`SKIP=check-copyright-year pre-commit run --all-files` clean, which covers the `REUSE.toml` edit. Both corrected paths verified to exist; the removed `cxrjs` path verified absent from the tree.

## Checklist

<sub>No tests: documentation and a licence-metadata deletion, with nothing to assert against.</sub>

## Summary by CodeRabbit

- **Documentation**
  - Updated the teleoperation guide with the correct simplified gripper example path.
  - Added comprehensive guidance for navigating, installing, running, packaging, and extending the available examples.
  - Documented conventions for multi-entry-point examples, namespace packages, relative imports, and registering new examples with the build system.
- **Chores**
  - Simplified project licensing metadata for example content.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,96 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Isaac Teleop examples
+
+Each subdirectory is a self-contained example. Read one, run it, or copy it into
+your own project — the last of those is the reason for the layout below.
+
+## Running an example
+
+```bash
+uv pip install -e ./examples/<name>
+python -m isaacteleop_examples.<name>
+```
+
+Every example runs this way, and its `README.md` gives the exact command.
+
+An example with one obvious entry point puts it in `__main__.py`, so
+`python -m isaacteleop_examples.<name>` runs it. An example that is several
+co-equal demos has no `__main__.py` at all — each is a submodule, run as
+`python -m isaacteleop_examples.<name>.<mod>`, and the README lists them.
+Promoting one of several peers to the default only makes it look privileged.
+
+## Layout
+
+```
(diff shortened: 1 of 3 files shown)


examples/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+```
+examples/<name>/
+├── README.md            — what it does, and the command to run it
+├── pyproject.toml       — the distribution, named isaacteleop-examples-<name>
+└── python/
+    └── isaacteleop_examples/
+        └── <name>/
+            ├── __init__.py
+            ├── __main__.py   — only if there is one obvious entry point
```
aristarkhovNV, line 36: could we possibly standartize that this is always the case?

aristarkhovNV (approved):

examples/README.md
jiwenc-nv: Yes. There are 13 PRs stacked on top of this to convert all examples to this setup.  Hopefully, agents will learn from the pattern.

PR #1037, opened 2026-08-28
examples(retargeting): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1036.

`multi_retargeter_tuning_demo_imgui.py` imported `example_retargeters` as a top-level module, which resolved only because the invoked script's own directory landed on `sys.path`. It broke under `python -m`, broke when copied into another project, and claimed the bare name `example_retargeters`. This moves the tree to the layout in `examples/README.md`.

```bash
uv pip install -e ./examples/retargeting
python -m isaacteleop_examples.retargeting.sources_example
```

Six co-equal demos, so no `__main__.py`: each is a submodule, listed in the new README. The example had no README at all before — the docs were the only place its scripts were named.

The `sharpa` and `wuji` reference pages now open their command blocks with the install step and use `-m`, so each block is runnable on its own rather than assuming you are standing in a source checkout.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships `isaacteleop_examples/retargeting/` with no namespace `__init__.py`. `uv pip install -e ./examples/retargeting` into a clean venv, then imported the demo modules with the CWD outside the repo; `import example_retargeters` fails, so the flat name is gone rather than relocated. `pre-commit` clean.

Not covered: live runs. `sharpa_hand_retargeter_demo --synthetic` and `wuji_hand_retargeter_demo --mode synthetic` need no headset and are the cheapest smoke tests here.

## Checklist

- [x] I have read and understood the [contribution guidelines](../CONTRIBUTING.md)
- [x] I have run the linter and formatter with `SKIP=check-copyright-year pre-commit run --all-files`
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix/feature works (or explained why not)
- [x] I have signed off all my commits (`git commit -s`) per the [DCO](../CONTRIBUTING.md#signing-your-work)

<sub>No tests: this is a move, and this example has no automated coverage.</sub>

## Summary by CodeRabbit

* **New Features**
  * Added Sharpa bimanual and Wuji hand retargeting demos.
  * Added examples for device tracking sources and retargeting workflows.
  * Added support for synthetic, replay, and hardware-driven Wuji demos.
[...]

PR #1038, opened 2026-08-28
examples(mcap_record_replay): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1037.

Nine modules imported `common` as a top-level module, which resolved only because the invoked script's own directory landed on `sys.path`. They broke under `python -m`, broke when copied into another project, and claimed the very generic bare name `common`. This moves the tree to the layout in `examples/README.md`.

```bash
uv pip install -e ./examples/mcap_record_replay
python -m isaacteleop_examples.mcap_record_replay.record_hand
python -m isaacteleop_examples.mcap_record_replay.replay_hand
```

Twelve co-equal scripts, so no `__main__.py`; the new README maps each channel to its live/record/replay module.

**One behaviour change, forced by the move.** Recordings defaulted to `Path(__file__).resolve().parent.parent / "recordings"`, which resolved to the example directory. Three levels deeper that expression points *inside the package*, and for a pip-installed copy it would write into `site-packages`. Recordings now default to `./recordings/` relative to the working directory, and the replay scripts look in the same place. The old path only ever made sense from a source checkout.

References updated beyond `docs/`: `rigs/full_body.yaml`, `src/plugins/noitom_mocap/README.md`, `src/plugins/vive_se3_tracker/README.md`.

**A second, deliberate behaviour change:** all seven viser viewers bind `0.0.0.0` rather than `127.0.0.1`, for the same reason — these run on a robot or workstation and get opened from a laptop. `--host 127.0.0.1` restores the old behaviour. Each startup line now reports the bind address instead of always printing `localhost`.

**The viser grid was a wall, not a floor.** `add_grid` defaults to `plane="xy"`, but the scene sets `set_up_direction("+y")` — so the grid stood vertical in every viewer. It is now an `xz` ground plane, 6 m with 0.25 m cells, and the camera starts centred on the origin at eye height so a tracked person fills the view on connect. That setup is a single `setup_scene()` helper in `common.py` rather than seven copies.

## Type of change

- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships `isaacteleop_examples/mcap_record_replay/` with no namespace `__init__.py`. `uv pip install -e ./examples/mcap_record_replay` into a clean venv, then **all 12 modules import** with the CWD outside the repo; `import common` fails, so the flat name is gone rather than relocated. `pre-commit` clean.

Not covered: recording needs a live OpenXR runtime. Replay against an existing `.mcap` is the cheapest check that the new `./recordings/` default behaves.

## Checklist

<sub>No tests: this is a move, and this example has no automated coverage.</sub>

## Summary by CodeRabbit

- **Documentation**
  - Updated MCAP recording and replay instructions to use the relocated Python examples and module-based commands.
  - Added setup, usage, networking, supported-channel, runtime, and file-location guidance.
  - Updated tracker, plugin, and headless replay examples with the current commands.

- **New Features**
  - Added automatic ground-grid and camera positioning for visualization scenes.
  - Recording and replay tools now default to `./recordings/` in the current working directory.
  - Improved package installation support for editable local development.


ivany-nv:
Two things riding along with the move.

**The output directory is no longer gitignored.** `Path(__file__).parent.parent / "recordings"` -> `Path.cwd() / "recordings"` is the right call for an installed copy, but the ignore rule is scoped to the example directory:

```
$ git check-ignore -v examples/mcap_record_replay/recordings/x.mcap
examples/mcap_record_replay/.gitignore:1:recordings   examples/mcap_record_replay/recordings/x.mcap
$ git check-ignore -v recordings/x.mcap
(not ignored)
```

The README tells you to run from the repo root (`uv pip install -e ./examples/mcap_record_replay`, then `python -m ...`), so following it drops untracked `.mcap` files at the root. Needs a `/recordings/` entry in the root `.gitignore`. Same issue in #1043 with `local_datasets/`.

**Seven more `--host` default flips.** Same change I flagged on #1034; that makes eight files carrying `default="0.0.0.0",  # noqa: S104`, and they are the only `S104` suppressions in the repo. A behaviour change this broad reads better as its own PR than as a rider on eight packaging commits -- `--host 0.0.0.0` already exists for the workflow it's meant to serve.

ivany-nv:
Verified this at `887db467`: built from source in a scratch worktree, installed
into a separate prefix, and ran the replay path end-to-end. The packaging work
itself is solid. What I found is that the move to module invocation stopped at
the code — the user-visible strings and one documented workflow still describe
the old script layout — plus one edge case in the new auto-discovery filter.

**What passed**

- `cmake` configure / `--build` / `--install` into a separate prefix: clean, no
  errors, no clang-format violations, wheel builds.
- `ctest`: 354/356. The two failures are `camera_viz_test_synthetic_offscreen`
  and `camera_viz_test_video_file_source`, both offscreen-GPU/codec limits of my
  box, and this PR touches no `camera_viz` file.
- PEP 420 layout in the installed tree is correct: no `__init__.py` at
  `isaacteleop_examples/`, `only-include` path present,
  `isaacteleop[cloudxr]==1.5+local` pinned.
- The new type-prefixed auto-discovery works. With a newer `controllers_*.mcap`
  and an older `full_body_*.mcap` both in `./recordings/`, `replay_full_body`
  with no argument correctly picks the `full_body_*` file — the bug the commit
  set out to fix. All four recorder prefixes (`controllers_`, `full_body_`,
  `hands_`, `se3_vive_`) match their replay globs.

---

## 1. The `full_body` rig's documented replay step cannot find the rig's own recordings

`rigs/full_body.yaml` is modified by this PR, and the line it changes now points
at a directory the replay never looks in.

The rig runs from the repo root and tells the C++ recorder where to write:

```yaml
cwd: ..                        # -> Teleop repo root
    command: "{install}/examples/mcap_record_replay/cpp/record_full_body 5 examples/mcap_record_replay/recordings/"
```
<sub>`rigs/full_body.yaml:24`, `:29`</sub>

So takes land in `<repo root>/examples/mcap_record_replay/recordings/`, which the
comment at `:16` states explicitly. But the replay instruction this PR rewrote
at `:19` claims the module "picks up the newest recording **there** by default",
and `resolve_mcap` resolves against the process CWD:

```python
recordings = Path.cwd() / "recordings"
```
<sub>`replay_full_body.py:65`</sub>

Run from the repo root — the rig's own `cwd`, and the CWD the `/recordings/`
`.gitignore` entry in `887db467` assumes — that is `<repo root>/recordings/`, a
different directory. With a rig recording present in the right place:

```console
$ ls examples/mcap_record_replay/recordings/
full_body_20260828_164337.mcap
$ python -m isaacteleop_examples.mcap_record_replay.replay_full_body
[replay] error: no .mcap files in <repo root>/recordings. Run record_full_body.py first or pass a path.
```

So the documented rig workflow ends in an error unless the user knows to `cd`
into the example directory or pass an explicit path. Worth deciding which
directory is canonical: the `.gitignore` commit and the README assume repo-root
CWD, while the rig and `rigs/full_body.yaml:16` assume the example-local one.

## 2. The `.py` invocation form survives in the strings users actually see

The repackaging changed how these are run, but the hand-written usage lines and
error messages still tell users to run the files as scripts. These are not
comments — the module docstring is the `argparse` description, so it is printed
by `--help`:

```console
$ python -m isaacteleop_examples.mcap_record_replay.replay_full_body --help
... replay_full_body.py [path/to/file.mcap] [--port 8080] [--loop] ...
```

- Usage lines: `record_hand.py:13`, `record_controller.py:13`,
  `record_full_body.py:13`, `replay_hand.py:12`, `replay_full_body.py:12`,
  `replay_controller.py:13` — all of the form `python replay_full_body.py …`.
- Error messages: *"Run `record_full_body.py` first or pass a path"* —
  `replay_hand.py:71`, `replay_full_body.py:72`, `replay_controller.py:78`,
  `replay_se3_vive.py:51`.
- `docs/source/references/mcap_record_replay.rst:226-229` — the **Replaying**
  block was left in script form even though the **Recording** block right above
  it (line 197) was converted to `python -m …`:

  ```rst
     uv run python replay_hand.py                       # newest file in ../recordings/
     uv run python replay_hand.py path/to/file.mcap     # explicit file
     uv run python replay_hand.py --loop                # repeat until Ctrl+C
     uv run python replay_hand.py --port 8090           # change viser port
  ```

  The surrounding prose names `install/examples/mcap_record_replay` as the
  working directory. Copied verbatim there it fails, because after the
  repackaging there is no `.py` at that level:

  ```console
  $ cd install/examples/mcap_record_replay && uv run --python 3.11 python replay_hand.py
  can't open file '.../mcap_record_replay/replay_hand.py': [Errno 2] No such file or directory
  ```

  These four lines are also the only `uv run` examples in the document.

## 3. The "newest `.mcap`" wording was not updated with the behavior change

`0981f53ff` changed selection from "newest of any kind" to "newest
type-prefixed, else newest of any kind", but touched only the three function
bodies. The old rule is still stated in:

- `replay_full_body.py:14`, `replay_hand.py:14`, `replay_controller.py:15` —
  *"If no path is given, the newest file under `./recordings/` is used."*
  (again, this is `--help` text)
- `examples/mcap_record_replay/README.md:18` — *"a replay given no path picks
  the newest file there"*, and the `# newest take` comment on line 14.
- `docs/source/references/mcap_record_replay.rst:226` — the inline
  `# newest file in ../recordings/` comment.
- `rigs/full_body.yaml:19-20` — *"picks up the newest recording there"* (see §1
  for the separate directory problem in the same sentence).

## 4. The `or` fallback re-admits the failure it fixes, and exits 0

```python
candidates = list(recordings.glob("full_body_*.mcap")) or list(
    recordings.glob("*.mcap")
)
```
<sub>`replay_full_body.py:66-68`, same shape at `replay_hand.py:65` and `replay_controller.py:72`</sub>

When no same-type recording exists, the fallback selects a wrong-type file
again. I checked what that does in practice, and it is worse than the commit
message's "fail with a missing-channel error" — it does not fail. With only
`controllers_*.mcap` present:

```console
$ python -m isaacteleop_examples.mcap_record_replay.replay_full_body
ReplayFullBodyTrackerImpl: body data not found      # once per frame, 3490 frames
[replay] reached end of recording after 3490 frames
[replay] done — viser server still up; Ctrl+C to exit
```

An empty viewer and a success message, exit code 0.

The fallback has a real purpose — a recording saved under a custom name still
needs to be discoverable — so I would keep it and log that the chosen file did
not match the expected prefix, rather than degrade quietly. One warning line at
the fallback would cover it.

## 5. Minor: `GroundGrid` is now duplicated against `main`

`common.py`'s `GroundGrid` is byte-identical, all 46 lines, to the copy in
`examples/deviceio_live_view/python/isaacteleop_examples/deviceio_live_view/deviceio_viser.py`,
which landed on `main` with #1034. Two exact copies will drift. Not a blocker,
and possibly deliberate to keep each example self-contained, but worth settling
now that both are in the tree rather than in flight.

## 6. Minor: `uv run` needs `--python 3.11`, and no doc says so

From the installed example, `uv run` without an explicit version picks the
system 3.12 and hard-fails:

```console
$ uv run python -m isaacteleop_examples.mcap_record_replay.replay_full_body --help
Using CPython 3.12.14
error: The current Python platform is not compatible with the lockfile's supported environments: `python_full_version < '3.12'`
```

`--python 3.11` fixes it. `environments = ["python_version == '3.11'"]` in
`pyproject.toml` makes the failure explicit rather than mysterious, which is
good, but neither the README nor the rst mentions `--python` anywhere — and the
rst's only `uv run` lines are the stale ones in §2. Worth one sentence wherever
§2 gets fixed.

---

## Not a defect — recorded here because it corrects an earlier suspicion of mine

On an earlier head I flagged that five of the seven `setup_scene()` call sites
discard the returned `GroundGrid` (`live_hand.py:45`, `live_controller.py:52`,
`replay_hand.py:137`, `replay_controller.py:142`, `replay_se3_vive.py:119`), so
those paths never call `follow()`. That is still true here, but it is **not** a
visual bug, and the reasoning that made it look like one was wrong.

I measured the actual coordinates in my own recordings:

| | |
|---|---|
| controller `aim_pos` Y | median **+0.833 m** (range +0.647 … +1.917) |
| body lowest joint Y | median **−0.141 m** |
| body highest joint Y | median **+1.275 m** |

The data reaching the examples is floor-relative — controllers sit ~0.83 m
*above* the origin, not below it, which is what a head-centered origin would
have produced. To be precise about what this does and does not show: it is
measured from a recording, so it cannot separate "the runtime granted a real
`XR_REFERENCE_SPACE_TYPE_STAGE`" from "it granted `local` and the client's
`xrOffsetY` compensation already absorbed the difference". Either way the
coordinates the examples consume are floor-relative, which is what matters here.

So a static grid at `y=0` is already correct for the hand/controller viewers,
and those paths have no body joints for `follow()` to track anyway.
`GroundGrid.follow()` is a refinement that absorbs the ~0.14 m by which tracked
feet land below the nominal floor, not a correction for a head-centered origin.
No change requested on this point.

ivany-nv:
Verified this at 5faf2019 with a clean build and install tree (not the source tree), plus a headless replay run driven by fixtures from `teleop_ros2_mcap_generator`. Install layout, generated pyproject, `python -m` from the install tree, legacy/converted split, clang-format and pre-commit all pass. Two behaviour fixes confirmed independently: the discovery filter picks `hands_*.mcap` over a newer `controllers_*.mcap` (before 0981f53f it picked the controller recording), and `/recordings/` is anchored to the repo root without over-ignoring nested paths.

User-facing text defects, both in files this PR already touches:

** `docs/source/references/mcap_record_replay.rst:177` and `:192` say "From the example directory:" and then give a repo-root-relative path.**

```
$ cd examples/mcap_record_replay && uv pip install -e ./examples/mcap_record_replay
error: Distribution not found at: file://<repo>/examples/mcap_record_replay/examples/mcap_record_replay

$ cd <repo> && uv pip install -e ./examples/mcap_record_replay
Resolved 27 packages
```

Sphinx does not catch this and it ships in the rendered HTML. It also contradicts 887db467's own rationale ("its README has you run from the repo root") — if the rst heading is right, that `.gitignore` rule is unnecessary.

Unrelated to this PR, for whoever picks it up: #1039–#1047 are still stacked on the pre-rewrite 091b91cb, so the `/recordings/` fix has not propagated — #1041 still writes `Path.cwd() / "recordings"` with no root ignore rule.

jiwenc-nv:
> Verified this at [5faf201](https://github.com/NVIDIA/IsaacTeleop/commit/5faf2019ac34bf26795deabfbcd31d1308c42f4c) with a clean build and install tree (not the source tree), plus a headless replay run driven by fixtures from `teleop_ros2_mcap_generator`. Install layout, generated pyproject, `python -m` from the install tree, legacy/converted split, clang-format and pre-commit all pass. Two behaviour fixes confirmed independently: the discovery filter picks `hands_*.mcap` over a newer `controllers_*.mcap` (before [0981f53](https://github.com/NVIDIA/IsaacTeleop/commit/0981f53fff937d25310a7d4653c94775cb7200ae) it picked the controller recording), and `/recordings/` is anchored to the repo root without over-ignoring nested paths.
> 
> User-facing text defects, both in files this PR already touches:
> 
> ** `docs/source/references/mcap_record_replay.rst:177` and `:192` say "From the example directory:" and then give a repo-root-relative path.**
> 
> ```
> $ cd examples/mcap_record_replay && uv pip install -e ./examples/mcap_record_replay
> error: Distribution not found at: file://<repo>/examples/mcap_record_replay/examples/mcap_record_replay
> 
> $ cd <repo> && uv pip install -e ./examples/mcap_record_replay
> Resolved 27 packages
> ```
> 
> Sphinx does not catch this and it ships in the rendered HTML. It also contradicts [887db46](https://github.com/NVIDIA/IsaacTeleop/commit/887db467fcd7d7cf777389764614e1e23b4fd24b)'s own rationale ("its README has you run from the repo root") — if the rst heading is right, that `.gitignore` rule is unnecessary.
> 
> Unrelated to this PR, for whoever picks it up: #1039–#1047 are still stacked on the pre-rewrite [091b91c](https://github.com/NVIDIA/IsaacTeleop/commit/091b91cb28b566e10c80bf580450337b54e9cd58), so the `/recordings/` fix has not propagated — #1041 still writes `Path.cwd() / "recordings"` with no root ignore rule.

good catch. fixed.

examples/mcap_record_replay/python/isaacteleop_examples/mcap_record_replay/replay_full_body.py
```diff
@@ -62,13 +63,21 @@ def resolve_mcap(path_arg: str | None) -> Path:
...
-        sys.exit(
-            f"[replay] error: no .mcap files in {recordings}. "
-            "Run record_full_body.py first or pass a path."
-        )
+        candidates = list(recordings.glob("*.mcap"))
+        if candidates:
+            print(
+                f"[replay] warning: no full_body_*.mcap in {recordings}, "
+                "falling back to the newest .mcap of any kind -- it may not "
+                "match this replay's channels."
```
aristarkhovNV: don't do a fallback, people may miss the log output and see a wrong replay

examples/mcap_record_replay/python/isaacteleop_examples/mcap_record_replay/replay_controller.py
```diff
@@ -67,13 +69,21 @@ def resolve_mcap(path_arg: str | None) -> Path:
...
     if not candidates:
-        sys.exit(
-            f"[replay] error: no .mcap files in {recordings}. "
-            "Run record_controller.py first or pass a path."
-        )
+        candidates = list(recordings.glob("*.mcap"))
+        if candidates:
+            print(
+                f"[replay] warning: no controllers_*.mcap in {recordings}, "
+                "falling back to the newest .mcap of any kind -- it may not "
```
aristarkhovNV: do not fallback

aristarkhovNV (approved):

examples/mcap_record_replay/python/isaacteleop_examples/mcap_record_replay/replay_full_body.py
jiwenc-nv: Agreed, dropped the fallback in 8dc80b6cf -- fails now instead of degrading silently.

examples/mcap_record_replay/python/isaacteleop_examples/mcap_record_replay/replay_controller.py
jiwenc-nv: Agreed, dropped the fallback in 8dc80b6cf -- fails now instead of degrading silently.

PR #1039, opened 2026-08-28
examples(teleop_ros2): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1038.

Fifteen modules imported their siblings as top-level modules — `constants`, `messages`, `geometry`, `assets`, `session_config` and the rest — which resolved only because the invoked script's own directory landed on `sys.path`. That broke under `python -m`, broke when copied into another project, and claimed an unusually generic set of names. This moves the tree to the layout in `examples/README.md`.

**Two workarounds propping this up are deleted**, which is the evidence the packaging is real rather than relocated:

- `pythonpath = ["."]` in the pyproject
- the example source directory on the ctest `PYTHONPATH`

The tests reach the in-tree source through a `conftest.py` instead, mirroring `mujoco_xr`'s, so a bare `pytest` in that directory works for the first time. Six test import lines change; no assertion does.

The node keeps its filename and gains a `__main__.py`, so the container entry point becomes `python -m isaacteleop_examples.teleop_ros2` and still forwards `--ros-args`. The install tree drops a level, so the Dockerfile `WORKDIR` follows, as do the three invocations in `build-ubuntu.yml`.

## Type of change

- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships 15 modules under `isaacteleop_examples/teleop_ros2/` with no namespace `__init__.py`. Editable-installed into a clean venv, the ROS-free modules (`constants`, `tensor_group_helpers`) import with the CWD outside the repo; `import constants` fails, so the flat names are gone rather than relocated. The new `conftest.py` was checked to resolve to `examples/teleop_ros2/python` with the package present and no namespace `__init__.py`. All test files compile. `pre-commit` clean.

**Not run: `ctest -L teleop_ros2`** — no ROS 2 on this host, so `geometry_msgs` and `rclpy` are unavailable. That suite and the container entry point are what CI should be trusted for here; please treat the CI run as the gate rather than my local checks.

## Checklist

<sub>No new tests: this is a move. The existing suite is the check — it exercises every moved module and passes with only its import lines changed.</sub>

## Summary by CodeRabbit

- **New Features**
  - Added installable packaging for the ROS 2 teleoperation example.
  - Added a module-based launch entry point for running the example consistently across environments.
  - Updated container execution to use the packaged example.

- **Bug Fixes**
  - Corrected package resolution and asset-root discovery when running the example after installation.
  - Updated ROS 2 launch and topic-verification commands to use fully qualified package paths.

- **Documentation**
  - Clarified that robot assets are not downloaded at runtime.
  - Updated setup and validation instructions for the packaged workflow.

- **Tests**
  - Improved test configuration for both source-tree and installed-package environments.


sgrizan-nv (approved):

ivany-nv:
A heads-up about this PR's base rather than its own content: the branch is still
built on the pre-rebase copy of the `mcap_record_replay` packaging commit, so it
carries versions of files that #1038 has since fixed.

This head (`774a9f0f4`) sits on `091b91cb2 examples(mcap_record_replay): package
under isaacteleop_examples` and does not contain #1038's current head
`887db467`. The same holds all the way up the stack — #1047 also resolves to
`091b91cb2`.

So the three commits #1038 picked up on 2026-09-08 are absent here:

- `681bc222b` — bind viser viewers to every interface by default
- `0981f53ff` — filter replay auto-discovery by recording type
- `887db467f` — ignore `/recordings/` written from the repo root

Checkable on the two heads:

```console
$ git show 774a9f0f4:examples/mcap_record_replay/python/isaacteleop_examples/mcap_record_replay/replay_full_body.py | sed -n 66p
    candidates = list(recordings.glob("*.mcap"))

$ git show 887db467f:examples/mcap_record_replay/python/isaacteleop_examples/mcap_record_replay/replay_full_body.py | sed -n 66p
    candidates = list(recordings.glob("full_body_*.mcap")) or list(

$ git show 774a9f0f4:.gitignore | grep -c '^/recordings/'
0
```

Two consequences:

1. Merging up the stack without a rebase would revert those fixes — the
   type-prefixed auto-discovery in particular, which is the one that stops
   `replay_full_body` from picking a `controllers_*.mcap`.
2. Because the merge-base is old, this PR's diff currently renders as 50 files
   / +310 −137, most of which is the stale `mcap_record_replay` content rather
   than the `teleop_ros2` change actually under review.

A rebase of #1039..#1047 onto the current #1038 head would clear both. I held
off reviewing the `teleop_ros2` change itself for now, since the diff as shown
is not what would land — happy to pick it up once the base is current.

ivany-nv:
**Verdict: a rebase onto #1038's current head looks unavoidable.** #1038 was rebased after this stack branched, and both merge orders now conflict on the same three files — including on the very fix #1038 added, where a careless resolution would quietly undo it.

Concrete evidence for the baseline point raised earlier. The stack's `091b91cb2` is **not** an ancestor of #1038's current head `887db467`. It was replaced by `f7b37f939`, followed by three more commits — one of which, `0981f53ff` "filter replay auto-discovery by recording type", is the fix for the defect reported in the #1038 review.

Both merge orders conflict, on the same three files (`replay_controller.py`, `replay_full_body.py`, `replay_hand.py`):

| order | result |
|---|---|
| #1038 then stack | #1038 clean, stack **3 conflicts** |
| stack then #1038 | stack clean, #1038 **3 conflicts** |

So sequencing alone will not avoid it. The part worth flagging: the conflict sits on the fix itself — a merged file holds both `glob("full_body_*.mcap") or glob("*.mcap")` and the older bare `glob("*.mcap")`, so whoever resolves it could revert the fix without noticing.

PR #1040, opened 2026-08-28
examples(camera_viz): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1039.

Sixteen import sites reached siblings as top-level modules — `pipeline`, `sources`, `transports`, `placements`, `codec` — resolving only because the invoked script's own directory landed on `sys.path`. Those are among the most generic names in the repo, and they broke under `python -m` and when copied into another project. This moves the tree to the layout in `examples/README.md`.

**The entry points are the risky half of this diff, not the imports:**

- `camera_viz.sh` runs `-m isaacteleop_examples.camera_viz` and `-m isaacteleop_examples.camera_viz.camera_streamer`
- the systemd unit's `ExecStart` follows
- `setup` now installs the example into its own venv, so `-m` resolves there, and `isaacteleop_examples.camera_viz` joins the setup smoke imports so a broken install fails setup rather than the first `run`
- the codec directory moves with the package; `build.sh` locates itself, so it needed no change

The test `conftest.py` inserts the namespace root instead of the example directory, so `pipeline` and `sources` stop being importable as top-level names there too. Five test import lines change; no assertion does.

## Type of change

- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships 27 modules under `isaacteleop_examples/camera_viz/` — `pipeline`, `placements`, `sources`, `transports`, `codec` — with no namespace `__init__.py`. Editable-installed into a clean venv, the modules import with the CWD outside the repo; `import pipeline` fails, so the flat names are gone rather than relocated.

`test_paired_frame_source_race.py` and `test_runner_health.py` pass (5 tests) through the new `conftest.py`. The other two need a GPU. `pre-commit` clean.

**Not covered, and this is where review effort belongs:** nothing here exercises `camera_viz.sh`, the systemd unit or the deploy path. Worth running `setup` then `run configs/synthetic.yaml --mode window`, and a `deploy` to a real robot box to confirm the unit's new `ExecStart` starts.

## Checklist

<sub>No new tests: this is a move. The existing suite is the check — it exercises the moved modules and passes with only its import lines changed.</sub>

## Summary by CodeRabbit

- **New Features**
  - Camera visualization and streaming tools can now be installed and run as packaged Python modules.
  - Added a module entry point for launching the camera visualization application.
  - Updated service and setup workflows to use the packaged tools.

- **Documentation**
  - Updated help commands, deployment instructions, project layout, source links, and native codec guidance.
  - Clarified split-mode capture and camera streaming usage.

- **Refactor**
  - Reorganized the camera visualization example into a standard package layout while preserving existing runtime behavior.


ivany-nv:
Every `:code-dir:` in this block was updated, but the `:code-file:` on `camera_streaming.rst:435` still points at the pre-move path:

```
- **The viewer** (:code-file:`camera_viz.py <examples/camera_viz/camera_viz.py>`) creates a
```

`_code_file_role` in `docs/source/conf.py:201` only builds a GitHub URL with no existence check, so `sphinx-build -W` won't catch it -- it publishes a dead link. Verified against `main`, where that path does exist, so it's new here. (#1005 would catch this class once it lands.)

The rest checks out: `camera_viz.sh`, the systemd unit template, `_install_deps.sh` (including the `CODEC_DIR` move and the added editable install), and `codec/build.sh` being `dirname $0`-relative.

ivany-nv:
**Verdict: one line missed in an otherwise complete update.** A single `:code-file:` role still points at the pre-move path, so that link will 404 in the built docs.

`docs/source/references/camera_streaming.rst:437`

```rst
- **The viewer** (:code-file:`camera_viz.py <examples/camera_viz/camera_viz.py>`) creates a
```

The target moved to `examples/camera_viz/python/isaacteleop_examples/camera_viz/camera_viz.py`. The rest of the file was updated — the tree diagram and the `pipeline/` and `codec/` `:code-dir:` roles all moved correctly; this one seems to have slipped through.

The role builds a GitHub URL from the path without checking that the file exists, so nothing in the build will flag it. Measured against a clean `main` worktree so it is attributable rather than pre-existing: 109 role targets and 2 dead on `main`, the same 109 with exactly this one added afterwards.

PR #1041, opened 2026-08-28
examples(noitom): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1040.

The three modules imported each other as top-level modules, and the README documented the consequence rather than fixing it — every Isaac Lab command carried `PYTHONPATH=~/IsaacTeleop/examples/noitom:$PYTHONPATH`. This moves the tree to the layout in `examples/README.md` and deletes the prefix.

The example gains its **first `pyproject.toml`**. `isaaclab` and `gymnasium` are deliberately *not* dependencies: `noitom_tasks.py` runs inside an Isaac Lab environment that already provides them, and neither installs from PyPI. The README now says to install into that interpreter, which is what replaces the `PYTHONPATH` prefix:

```bash
./isaaclab.sh -p -m pip install -e ~/IsaacTeleop/examples/noitom
```

and the callback becomes `--external_callback isaacteleop_examples.noitom.noitom_tasks.register_tasks`.

One import lived inside a function body, indented, so it needed catching separately from the module-level ones — worth knowing if you review the other trees.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships the four modules under `isaacteleop_examples/noitom/` with no namespace `__init__.py`. An AST sweep of the tree confirms no sibling resolves as a top-level import any more.

**Not covered, and I cannot cover it:** nothing here has a Noitom mocap suit or an Isaac Lab install. The dependency list is inferred from imports, and this example never had one before — a missing runtime dep would only show on the machine with the suit attached. Please run both README paths (the MCAP recording and the Isaac Lab teleop task) before merging.

## Checklist

- [x] I have read and understood the [contribution guidelines](../CONTRIBUTING.md)
- [x] I have run the linter and formatter with `SKIP=check-copyright-year pre-commit run --all-files`
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix/feature works (or explained why not)
- [x] I have signed off all my commits (`git commit -s`) per the [DCO](../CONTRIBUTING.md#signing-your-work)

<sub>No tests: this is a move, and no CI has the hardware.</sub>

## Summary by CodeRabbit

- **New Features**
[...]

PR #1042, opened 2026-08-28
examples(oglo_tactile): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1041.

`oglo_teleop_record.py` imported `oglo_heatmap` as a top-level module, which resolved only because the invoked script's own directory landed on `sys.path`. This moves the tree to the layout in `examples/README.md`.

The example gains its **first `pyproject.toml`**, so the README's loose `pip install pillow numpy` / `pip install cupy-cuda12x` become an install of the example and its `gpu` extra. `cupy` stays optional — without it the renderer still writes PNGs and only the in-headset overlay is lost.

**One path fix the move forces.** `_default_plugin_bin()` walked up `parents[2]` to the repo root; three levels deeper that lands on the example's `python/` directory. It is now `parents[5]`, with a note that it only resolves from a source checkout — an installed copy lands in `site-packages`, where the existing PATH fallback takes over.

## Type of change

- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships both modules under `isaacteleop_examples/oglo_tactile/` with no namespace `__init__.py`. Editable-installed into a clean venv, both import with the CWD outside the repo; `import oglo_heatmap` fails. `parents[5]` verified to land on the repo root. `pre-commit` clean.

**Not covered:** nothing here has an OGLO glove. The dependency list is inferred from imports and this example never had one before, so a missing runtime dep would only show on the machine with the gloves attached.

## Checklist

<sub>No tests: this is a move, and no CI has the hardware.</sub>

## Summary by CodeRabbit

- **New Features**
  - The OGLO tactile recording example can now be installed as a package.
  - Optional GPU support is available for headset overlays, while recording remains available without GPU dependencies.
  - The example supports Python 3.11–3.13 and can be launched through its installed command/module.

- **Bug Fixes**
  - Improved executable discovery so the recording demo works reliably both from source checkouts and installed environments.


ivany-nv (approved):
Read the diff line by line. Two changes beyond the move: the `oglo_heatmap` import becoming relative, and `here.parents[2]` becoming `here.parents[5]`. Checked the arithmetic — from `python/isaacteleop_examples/oglo_tactile/`, the repo root is five parents up, so that is right, and the comment about an installed copy falling through to the PATH lookup matches the code.

Not installed by `install_python_example()` and I have no OGLO hardware here, so nothing to run.

PR #1044, opened 2026-08-28
examples(haptic_feedback): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1043.

No flat sibling imports here — the two examples are independent — but the tree kept its `pyproject.toml` under `python/` and its README documented `uv run <script>.py`, so it needed the same conversion as the rest for one install command to work across every example.

```bash
uv pip install -e ./examples/haptic_feedback
python -m isaacteleop_examples.haptic_feedback.controller_haptic_example
```

`device/haptic_feedback.rst` follows: it pointed at `examples/haptic_feedback/python/` and told the reader to run from that directory.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships both modules under `isaacteleop_examples/haptic_feedback/` with no namespace `__init__.py`. `pre-commit` clean.

**Not covered:** both examples need a controller or a haptic glove plus a running CloudXR/OpenXR runtime.

## Checklist

- [x] I have read and understood the [contribution guidelines](../CONTRIBUTING.md)
- [x] I have run the linter and formatter with `SKIP=check-copyright-year pre-commit run --all-files`
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix/feature works (or explained why not)
- [x] I have signed off all my commits (`git commit -s`) per the [DCO](../CONTRIBUTING.md#signing-your-work)

<sub>No tests: this is a move with no import changes at all.</sub>

## Summary by CodeRabbit

- **Documentation**
  - Updated haptic feedback example instructions to install the package and run examples as Python modules.
  - Standardized example naming and updated referenced paths.

- **New Features**
  - Added packaging configuration for installing the haptic feedback example with supported Python versions and required dependencies.

- **Chores**
  - Moved the example installation destination to the package root.
  - Removed the previous Python-specific project configuration.
  - Added licensing headers to the example module.

PR #1045, opened 2026-08-28
examples(cloudxr_mujoco_teleop): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1044.

No flat sibling imports — the two scripts are independent — but the tree ran as `uv run <script>.py` from its own directory, so it needed the same conversion as the rest for one install command to work across every example.

```bash
uv pip install -e ./examples/cloudxr_mujoco_teleop
python -m isaacteleop_examples.cloudxr_mujoco_teleop.visualize_poses_mujoco_example
```

`vive_assets/` moves **inside** the package: both scripts resolve `ASSETS_DIR` from `__file__`, so the meshes have to travel with the modules or the move would silently break them. The wheel now carries them (6 asset files), matching how `mujoco_xr` ships its own assets.

The `REUSE.toml` override for `examples/cloudxr_mujoco_teleop/**` still covers the nested paths, so the HTC copyright is unaffected.

## Type of change

- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships 3 modules and all 6 `vive_assets` files under `isaacteleop_examples/cloudxr_mujoco_teleop/`, with no namespace `__init__.py`. `pre-commit` clean.

**Not covered:** both need a headset and a CloudXR runtime. The asset move is the part worth eyeing — if the meshes fail to load, `ASSETS_DIR` is the reason.

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Updated setup and launch instructions for the CloudXR MuJoCo teleoperation example.
  * Revised example naming and directory layout documentation.

* **Chores**
  * Added package build configuration for distributing the example.
  * Added licensing information to the example module.


ivany-nv (approved):
Read the diff line by line: every Python file is a 100% rename, so the only new content is the `pyproject.toml` (`only-include` + `sources`, no `__init__.py` on the namespace) and the README path updates.

Not installed by `install_python_example()`, so there is no install tree to check.

PR #1046, opened 2026-08-28
examples(oxr): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1045.

No flat sibling imports — these scripts are independent — but the tree kept its `pyproject.toml` under `python/` and ran as `uv run <script>.py`, so it needed the same conversion for one install command to work across every example.

```bash
uv pip install -e ./examples/oxr
python -m isaacteleop_examples.oxr.test_modular
```

**The GPU CI keeps invoking these by path, not by `-m`.** The test image copies `examples/oxr/python/` to `/app/tests` and runs `python <script>`, so the entries in `CXR_PYTHON_GPU_TESTS` just gain the namespace prefix. With no sibling imports the path-based invocation stays correct. Moving that container to `-m` would mean changing `Dockerfile.test`, the compose runner and the list together, on a path I cannot reproduce locally — worth doing, but not blind and not here.

The example carried **two overlapping READMEs**, at the example root and under `python/`. The `python/` one was a subset, so it goes and the root one absorbs its run commands.

## Type of change

- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships 12 modules under `isaacteleop_examples/oxr/` with no namespace `__init__.py`. `pre-commit` clean.

**Not covered:** everything here needs a headset and the CloudXR runtime. The CI list change is the risky part — please confirm the GPU job still finds all three scripts.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - OpenXR Python examples are now installable as a package with editable installation support.
  - Examples can be run consistently as Python modules using the documented commands.

- **Documentation**
  - Updated tracking, camera, plugin, and testing instructions to reflect the new example layout.
  - Consolidated setup and usage guidance in the top-level OpenXR README.

- **Bug Fixes**
  - Improved plugin discovery for examples used from both source checkouts and installed layouts.
  - Updated automated test paths to match the reorganized examples.


ivany-nv:
`run_tests_with_cloudxr.sh` now uses `isaacteleop_examples/oxr/test_*.py` relative to `/app/tests`, described as a copy of `examples/oxr/python/`. Nothing in this repo creates `/app/tests`, so I can't check it here -- worth confirming against whatever builds the test image rather than letting the GPU job find out.

The scripts themselves are fine either way: they only import `isaacteleop.*`, no intra-example imports, so running them as bare scripts still works.

PR #1047, opened 2026-08-28
examples(teleop): package under isaacteleop_examples

jiwenc-nv:
## Description

Part of #985. Stacked on #1046.

No flat sibling imports — the eleven examples are independent — but the tree kept its `pyproject.toml` under `python/`, so it needed the same conversion for one install command to work across every example.

```bash
uv pip install -e ./examples/teleop
python -m isaacteleop_examples.teleop.gripper_retargeting_example_simple
```

**Two things the move forces, both silent if missed:**

- `config/` moves inside the package. `dex_bimanual_example` and `full_bimanual_reordering_example` resolve it from `__file__`, so the dex-retargeting YAML has to travel with the modules; the wheel now carries all five files.
- Six scripts walked up four parents to reach the tree root — the install prefix for `PLUGIN_ROOT_DIR`, the checkout for the `so101_leader` build path. Three levels deeper that lands on the example's own `python/` directory, so they are `parents[5]` now. The install tree keeps the same shape, so one depth serves both.

This example has the widest reference footprint of the series: five docs pages, `src/plugins/so101_leader/README.md`, and a docstring in `joint_state_retargeter.py`. Runnable commands became `-m`; prose paths that gained nothing from the full nesting were shortened to `examples/teleop`.

## Type of change

- [x] Documentation update

## Testing

`x86_64` / Ubuntu, Python 3.12. Wheel ships 11 modules and all 5 `config/` files under `isaacteleop_examples/teleop/`, with no namespace `__init__.py`. `parents[5]` verified to land on the repo root. `pre-commit` clean.

**Not covered:** every example here needs a headset, and several need the `so101_leader` or synthetic-hands plugin. The `parents[5]` change is the part worth checking on a real run — a wrong depth shows up as a plugin that cannot be found, not as an import error.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Teleoperation examples are now packaged for installation and can be launched as Python modules.
  * Examples support both source-checkout and installed-package layouts.

* **Documentation**
  * Updated setup, retargeting, CloudXR, controller, pedal, and gripper instructions to use the reorganized example paths and commands.
  * Updated related plugin and API documentation links.


ivany-nv (commented):
**Summary**

This moves the teleop example scripts into `isaacteleop_examples.teleop`, promotes the example's `pyproject.toml` to the example root, and rewrites the references across five docs pages, a plugin README, and one docstring. The `parents[5]` rewrite you called out as the part worth checking holds in both the checkout and the install tree, and neither red check is yours — `build-ubuntu` fails on `cloudxr_test_background` and the base head is red too, on an unrelated Depthai download. One merge blocker: two of the rewritten `python -m` commands lost the install step they now depend on.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`docs/source/getting_started/quick_start.rst:342`](https://github.com/NVIDIA/IsaacTeleop/blob/de2ea9521e16187ef01efa903743613f113dc27b/docs/source/getting_started/quick_start.rst#L342) - `python -m isaacteleop_examples.teleop.gripper_retargeting_example_simple` fails with `ModuleNotFoundError: No module named 'isaacteleop_examples'` for a reader following the page in order, because step 2 installs only the `isaacteleop` wheel and that wheel declares `packages = { isaacteleop = "src/python/isaacteleop" }`. The pre-move command ran the script by path and needed no examples install, so this is a regression on the primary onboarding path. [`docs/source/references/cloudxr.rst:190`](https://github.com/NVIDIA/IsaacTeleop/blob/de2ea9521e16187ef01efa903743613f113dc27b/docs/source/references/cloudxr.rst#L190) has the identical gap; the other 13 `python -m isaacteleop_examples…` lines in `docs/` all sit directly under a `uv pip install -e ./examples/<name>`, including the one this PR added at `joint_space.rst:109`. Add that install line above both commands. |
| 💡 SUGGESTION | [`examples/teleop/`](https://github.com/NVIDIA/IsaacTeleop/tree/de2ea9521e16187ef01efa903743613f113dc27b/examples/teleop) - The converted example ships no root `README.md`, so its ten demo modules have no listed run commands anywhere; [`examples/README.md:23`](https://github.com/NVIDIA/IsaacTeleop/blob/de2ea9521e16187ef01efa903743613f113dc27b/examples/README.md#L23) assigns that list to the README for exactly this multi-demo, no-`__main__.py` shape, and 13 of the 15 examples with a root `pyproject.toml` have one — the only other exception is `rebot`, the carve-out the same file documents. Every one of the ten sibling PRs in this stack updates or adds its example README, and #1037 created `examples/retargeting/README.md` from nothing for the same reason — #1047 is last in the stack, so no later PR closes it. Nothing enforces this in CI, so treat it as convention rather than a gate. Add a README modelled on `examples/retargeting/README.md`. |
| 🧹 NIT | none |

- `docs/source/getting_started/quick_start.rst:342` - Add `uv pip install -e ./examples/teleop` as the first line of the step-6 `.. code-block:: bash`.
- `docs/source/references/cloudxr.rst:190` - Add `uv pip install -e ./examples/teleop` above the `python -m` line in the "Run teleop examples against the service" block.
- `examples/teleop/README.md` - Create it modelled on `examples/retargeting/README.md`: one-paragraph description, the `uv pip install -e ./examples/teleop` line, and a table of the ten modules with their `python -m isaacteleop_examples.teleop.<mod>` commands.

ivany-nv:
**Verdict: the stack looks good.** Merged onto current `main` and built it end to end, and everything I could check mechanically came back clean — 13/13 wheels coexisting in one environment, 89/89 modules compiling, 13/14 examples conforming to the new convention. Two small things are worth a look and I've put them on #1040 and on this PR separately.

Verified the whole stack as a unit, since #1039–#1047 form one linear chain and `1e0926fd4` carries the cumulative content. Everything below was run, not reasoned about.

**Merged into current `main` (`5b5bceb72`) and built.** `main` had moved 5 commits ahead, including `7513288d0`, which deletes `examples/mujoco_xr/` wholesale (23 files at the stack's branch point) and renames the app to `examples/robot_viz`. No stack commit touches `examples/mujoco_xr`, so the merge is clean: 0 conflicts, the deletion is preserved, and `add_subdirectory(examples/mujoco_xr)` is correctly dropped from the root `CMakeLists.txt` even though the stack side still carried that line. `configure` / `build` / `install` all exit 0, with no errors from this tree — the only `error` matches in the build log are miniz's `zError` unused-function warning from a fetched dependency.

**Every example distribution coexists in one environment.** This is the point of #985 and the collision scenario `examples/README.md` warns about, so it seemed worth testing directly rather than per-example:

- 13/13 wheels build with `uv build --wheel`
- 13/13 are rooted at `isaacteleop_examples/` — none claims a bare top-level name, i.e. the `packages` failure mode the convention doc describes does not occur
- 13/13 contain no `isaacteleop_examples/__init__.py`
- installing all 13 into one venv yields **one** namespace path and **13/13** resolvable subpackages — no collision, no shadowing
- `find_spec` + `py_compile` across every submodule: **89/89**, 0 failures

**Convention conformance.** Each rule in `examples/README.md` (#1035) checked mechanically against all 14 examples: distribution naming, hatchling `only-include`/`sources` rather than `packages`, namespace without `__init__.py`, `[tool.uv.sources]` placement, relative intra-example imports. The checker was red-green tested first — a synthetic clean tree reports zero issues, and injecting each violation in turn makes the matching rule fire, 9 for 9 — so "clean" here means something. 13/14 clean; the one deviation is `examples/teleop`, noted separately on this PR. `mujoco_xr` is exempt from the `only-include` rule, which is worded "With hatchling" — it needs scikit-build-core for a compiled extension and maps paths via `[tool.scikit-build.wheel] packages` with a comment explaining why.

**The `parents[5]` change (#1041, #1046, #1047) is correct.** Files moved two levels deeper, so `.parent`×4 (= `parents[3]`) becomes `parents[5]`. Measured rather than derived: 14/14 uses resolve to the repo root in a checkout, 5/5 to the install prefix in the install tree, where `plugins/` does exist. (Minor: the commit message says "three levels deeper"; it is two. The code is right.)

**Documentation.** All 22 `python -m isaacteleop_examples....` targets in `examples/`, `docs/`, and `scripts/` resolve. Sphinx `:code-file:` / `:code-dir:` targets: 109 on a clean `main` worktree with 2 dead (both `deps/cloudxr/*`, pre-existing), 109 post-merge with 3 — the one addition is on #1040.

**What this does not cover:** no headset was attached for this round, so nothing here exercises live tracking; and six examples (`camera_viz`, `cloudxr_mujoco_teleop`, `lerobot`, `noitom`, `oglo_tactile`, plus `teleop`) never enter the install tree, so they were verified as wheels and source, not from an installed prefix.

ivany-nv:
**Verdict: the PR itself checks out; this is a pre-existing wrinkle the diff happens to touch.** `examples/teleop/CMakeLists.txt` is never read by CMake, so the `install_python_example` line in it — including the line this PR changes — has no effect. Worth a decision at some point, not necessarily in this PR.

The root `CMakeLists.txt` never `add_subdirectory`s `examples/teleop`, so the example does not reach the install tree (confirmed: the install prefix has 9 examples, `teleop` not among them).

To be clear this predates the PR — `main` likewise omits `examples/teleop` from the root list, and its `examples/teleop/CMakeLists.txt` already had `DESTINATION examples/teleop/python`. Mentioning it only because this PR edits exactly that line, changing the destination to `examples/teleop`, and because `examples/README.md` ties that call to shipping in the install tree: "If the example ships in the install tree, add `install_python_example(DESTINATION examples/<name>)` to its `CMakeLists.txt`." As it stands the file reads as if it installs something. If `teleop` is meant to be source-only, a one-line comment saying so would save the next reader the same trip; if it is meant to install, `add_subdirectory(examples/teleop)` is the missing piece. Either is fine by me, and either could just as easily be a follow-up.

Two claims in the commit message do check out: the six scripts' `parents[5]` resolves correctly in both trees, and the dex retargeting config does travel with the package — all five files land at `isaacteleop_examples/teleop/config/dex_retargeting/` in the wheel and are still resolvable once installed.

PR #1060, opened 2026-08-29
ci: stop third-party artifact downloads reddening unrelated PRs

### LINKED ISSUE #1049: ci: third-party artifact downloads fail intermittently, reddening unrelated PRs
CI fails intermittently on third-party artifact downloads, unrelated to the changes under test. Measured over one afternoon on a 14-PR stack: **22 failed `Build Ubuntu` runs, 48 failed jobs**.

| Job | Count | Cause |
| --- | ---: | --- |
| `build-ubuntu` | 18 | `Couldn't download depthai-device-fwp.tar.xz` |
| `build-ubuntu` | 3 | `Couldn't download depthai-bootloader-fwp.tar.xz` |
| `build-ubuntu` | 5 | `Failed to download CloudXR Runtime SDK` (NGC 404) |
| `wheel-gate` | 22 | aggregator only — mirrors `build-ubuntu` |
| `test-teleop-ros2` | observed | CloudXR SDK 404, exit 4 |

It hits `main` too: `Build Ubuntu` failed on `c5fe6624c` with the CloudXR error, green either side.

Three things make one flaky download expensive:

1. **`fail-fast: true` on pull requests** (`build-ubuntu.yml:36`) turns a single 404 into 1 failure plus 11 cancelled jobs and a red `wheel-gate`.
2. **`curl --retry` does not retry 4xx.** `scripts/download_cloudxr_runtime_sdk.sh:105` passes `--retry 3 --retry-delay 5`, inert against the 404 actually seen.
3. **DepthAI device binaries have no mirror or cache** — fetched per job by `DepthaiDownloader.cmake`, so every matrix entry is an independent chance to fail.

Directions: cache or mirror both artifacts, retry 4xx explicitly, or let `test-teleop-ros2` skip when the SDK is unavailable.

jiwenc-nv:
CI reddens on things unrelated to the change under test: every `build-ubuntu` matrix entry fetches the DepthAI device firmware and the CloudXR SDKs over the network, so one transient failure at `artifacts.luxonis.com` or NGC fails the job. Measured over one afternoon on a 14-PR stack: 22 failed `Build Ubuntu` runs, 48 failed jobs. `main` is hit too.

Cache both artifacts, and retry the one download that curl's `--retry` can't cover.

- **DepthAI** — restore `_deps/depthai-build/resources` before Configure, keyed on the `DEPTHAI_COMMIT` pin. Upstream's `DownloadAndChecksum` skips any file already on disk, so a hit makes the fetch a no-op.
- **CloudXR** — cache `deps/cloudxr/*.tar.gz|*.tgz` inside `setup-cloudxr-sdk`, which both `build-ubuntu` and `test-teleop-ros2` use. Both scripts already prefer a local tarball over NGC.
- Restore/save are split rather than the combined `actions/cache`: that one saves on failed jobs too, and `file(DOWNLOAD)` leaves a truncated file behind on error — a half-written `.fwp` would satisfy the "already downloaded" check forever.
- `download_cloudxr_runtime_sdk.sh` retries the whole listed→unlisted sweep, not each request. A 404 on an individual remote name is how the script discovers the published name, so per-request retry would add delay to every successful run.

`fail-fast` is left alone: it amplifies flakes but doesn't cause them, and disabling it burns runner minutes on genuine compile errors.

Closes #1049

## Testing

- Stubbed `curl`: all-404 → 3 sweeps, 12 requests, exit 1; first-candidate hit → 1 request, 0 s added.
- Cold- and warm-cache CI runs on this PR are the real check — the cache keys can only be verified against the Actions cache service.

## Summary by CodeRabbit

- **Performance**
  - Faster, more reliable builds by caching CloudXR SDK and DepthAI firmware downloads.

- **Bug Fixes**
  - Improved CloudXR SDK retrieval with fallback resources and automatic retries.
  - Partial downloads are cleared between attempts to prevent failed installations.
  - Enhanced failure reporting after repeated download attempts.

### DIFF
--- a/.github/actions/setup-cloudxr-sdk/action.yml
+++ b/.github/actions/setup-cloudxr-sdk/action.yml
@@ -30,6 +30,20 @@ runs:
         echo "CloudXR Runtime SDK version: ${CXR_RUNTIME_SDK_VERSION}"
         echo "CloudXR Web SDK version: ${CXR_WEB_SDK_VERSION}"
 
+    # Both download scripts prefer a tarball already sitting in deps/cloudxr/ over
+    # fetching from NGC, so a cache hit makes the download a no-op. NGC 404s
+    # transiently and reddens otherwise-good runs; this keeps the pipeline off the
+    # network on the common path. Restore only — the save is gated on the downloads
+    # succeeding, so a partial fetch is never cached.
+    - name: Restore CloudXR SDK tarballs
+      id: cache
+      uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+      with:
+        path: |
+          deps/cloudxr/*.tar.gz
+          deps/cloudxr/*.tgz
+        key: cloudxr-sdk-${{ runner.os }}-${{ runner.arch }}-rt${{ steps.parse.outputs.runtime_version }}-web${{ steps.parse.outputs.web_version }}
+
     - name: Download CloudXR Runtime SDK
       shell: bash
       env:
@@ -44,3 +58,12 @@ runs:
         CXR_WEB_SDK_VERSION: ${{ steps.parse.outputs.web_version }}
       run: |
         ./scripts/download_cloudxr_sdk.sh
+
+    - name: Save CloudXR SDK tarballs
+      if: steps.cache.outputs.cache-hit != 'true'
+      uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+      with:
+        path: |
+          deps/cloudxr/*.tar.gz
+          deps/cloudxr/*.tgz
+        key: ${{ steps.cache.outputs.cache-primary-key }}
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -110,6 +110,21 @@ jobs:
         ccache --version
         ccache --show-config
 
+    # depthai-core downloads the device/bootloader *.fwp blobs into its build
+    # tree at configure time, once per matrix entry; artifacts.luxonis.com fails
+    # often enough to redden unrelated PRs. Its DownloadAndChecksum skips any
+    # file already on disk, so a restored resources/ makes the fetch a no-op.
+    # Keyed on the DEPTHAI_COMMIT pin, which selects the firmware version.
+    # Split restore/save on purpose: the combined action also saves on a failed
+    # job, and CMake's file(DOWNLOAD) leaves a truncated file behind on error —
+    # that would be cached and then pass the "already downloaded" check forever.
+    - name: Restore DepthAI device resources
+      id: depthai-resources
+      uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+      with:
+        path: ${{ env.BUILD_DIR }}/_deps/depthai-build/resources
+        key: depthai-resources-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('cmake/DepthAIVcpkgManifest.cmake') }}
+
     - name: Configure CMake
       run: |
         coverage_args=()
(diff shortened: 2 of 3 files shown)


ivany-nv (approved):
Approving. The reasoning holds up — I verified the two load-bearing claims against depthai-core at the pinned commit `8d32a1f` rather than taking them on trust:

- **Cache path is right.** depthai-core's root `CMakeLists.txt` sets `DEPTHAI_RESOURCES_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/resources"`, and `src/plugins/oak/CMakeLists.txt` adds it via `add_subdirectory(${depthai_SOURCE_DIR} ${depthai_BINARY_DIR})`, so that resolves to `build/_deps/depthai-build/resources`.
- **The truncation worry is real, not theoretical.** `DownloadAndChecksum` opens with `if(EXISTS "${output}") ... return()` and never re-verifies the checksum, so the split restore/save is justified. Firmware download failure ends in `message(FATAL_ERROR "Aborting.")`, so Configure fails and the save is correctly skipped.

I also ran `download_cloudxr_runtime_sdk.sh` locally under bash 3.2 with a stubbed curl and reproduced your numbers: all-404 gives exit 1, 3 sweeps, 12 requests, 10s; the realistic happy path (`-external` 404s, plain name serves) is 4 requests and 0s added; no leftover files in `deps/cloudxr` on any failure path.

Two things worth a follow-up, neither blocking:

**1. `resources/` holds one file the "Configure succeeded ⇒ everything checksummed cleanly" comment doesn't cover.** depthai also writes `depthai-shared-commit-hash-<id>.txt` there, and it sets `DEPTHAI_SHARED_COMMIT_HASH_ENFORCE OFF` under `$ENV{CI}` — only the bootloader one is `ON`. So a failed device-side hash download is non-fatal, `DownloadAndChecksum` leaves the partial `${output}` on disk, Configure still succeeds, and the truncated `.txt` gets saved. From then on the `EXISTS` skip means it's never re-fetched and the host/device shared-commit check silently degrades to a spurious `WARNING` for the life of that key. That's the failure mode this PR set out to prevent, one file over. Narrowing the cached path to `.../resources/*.tar.xz` would close it.

**2. The cross-PR benefit only arrives after this merges — worth a line in the description.** Caches are branch-scoped: one written on `jiwenc-nv/ci-stability` is readable by that branch and its descendants, not by other PRs. Only default-branch caches are visible everywhere, so the 14-PR-stack scenario stays red until this lands on `main` and a `main` push run populates the keys. Otherwise someone may see the next PR still miss and conclude this didn't work.

Minor notes:

- The body says "keyed on the `DEPTHAI_COMMIT` pin"; the key is `hashFiles('cmake/DepthAIVcpkgManifest.cmake')`. Conservative in the safe direction and consistent with the existing `Cache vcpkg` key — just the wording.
- Expect five "Unable to reserve cache" lines per arch on the first run, since the depthai key deliberately omits `build_type`/`python_version` and all six matrix entries race for it. `actions/cache/save` treats that as informational, so it's harmless — just don't let it get read as a failure.
- Cache budget is a non-issue: I measured CloudXR at 11.8 + 12.1 + 0.3 ≈ 24 MB/arch and the depthai firmware at ~6 MB/arch, so ~60 MB against the 10 GB repo limit. No risk of evicting the ccache/vcpkg entries.
- For the pinned 6.3.0 the listed `cloudxr-runtime` resource 404s on all four names and the unlisted one serves, so "3 attempts" is effectively three attempts at the unlisted resource. Free, since curl doesn't retry 4xx.

Nice touch that `deps/cloudxr/.gitignore` already ignores exactly `*.tar.gz` and `*.tgz`, matching the cached globs, and that nothing in the tree globs those tarballs version-agnostically — a stale cached tarball stays inert.

yanziz-nvidia (approved):

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**
This PR adds version/architecture-scoped artifact caches and retries the complete CloudXR NGC fallback sweep. CI is green, and focused testing confirmed the expected 12-request, three-sweep failure path. Other than the existing findings in previous comments, one cache-integrity risk remains.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | none |
| 💡 SUGGESTION | [`.github/actions/setup-cloudxr-sdk/action.yml:62`](https://github.com/NVIDIA/IsaacTeleop/blob/214c014e3d1c67c0cb621d9a4e2f0211ed966e84/.github/actions/setup-cloudxr-sdk/action.yml#L62) - The cache is saved immediately after download, while the downloaders only require nonempty files and the runtime archives are not extracted until the later build. If NGC returns a nonempty malformed payload with a successful HTTP status, this creates a poisoned cache before the current job detects the problem; subsequent jobs restore it and skip downloading. Validate every matched archive using published digests, or at least archive and required-member checks, before saving it. |
| 🧹 NIT | none |

- `.github/actions/setup-cloudxr-sdk/action.yml:62` - Before saving the CloudXR cache, validate each runtime, experimental runtime, and Web SDK archive and remove/fail on invalid files.

PR #1067, opened 2026-09-01
docs(ecosystem): note Meta Quest lacks full-body robot control support

jsepulveda-nvidia:
## Summary
Modified 2 major documentation pages available to the public:
https://nvidia.github.io/IsaacTeleop/main/overview/ecosystem.html
- The Meta Quest 2/3/3S entry on the ecosystem/devices page did not call out that it tracks head and hands only, not full-body.
- Adds a requirements line noting it's not supported for full-body robot control, with a link to the Body Tracking guide for equipment with Motion Trackers.

https://nvidia.github.io/IsaacTeleop/main/device/body_tracking.html
- Changed the note at the top discussing the Meta skeleton and that no translation exists from it's WebXR Skeletal format.
- Remove the "Quest Body Tracking (Limited Support)" section entirely.

## Test plan
- [ ] Build docs and check the Requirements panel for the Meta Quest 2 / 3 / 3S row renders as expected

## Summary by CodeRabbit

* **Documentation**
  * Added Meta Quest compatibility guidance, noting it is not supported for full-body robot control.
  * Linked to the Body Tracking guide and recommended Motion Trackers for full-body teleoperation.

### DIFF
--- a/docs/source/_data/devices.yaml
+++ b/docs/source/_data/devices.yaml
@@ -79,10 +79,6 @@ devices:
         - Isaac Teleop Web Client — runs in the headset browser
       requirements:
         - No additional software
-        - Not supported for full-body robot control
-        - label: Body Tracking guide
-          doc: /device/body_tracking
-          note: for full-body teleoperation, use equipment with Motion Trackers
       acquire:
         - label: meta.com
           url: https://www.meta.com/quest/
--- a/docs/source/_data/devices.yaml
+++ b/docs/source/_data/devices.yaml
@@ -79,6 +79,10 @@ devices:
         - Isaac Teleop Web Client — runs in the headset browser
       requirements:
         - No additional software
+        - Not supported for full-body robot control
+        - label: Body Tracking guide
+          doc: /device/body_tracking
+          note: for full-body teleoperation, use equipment with Motion Trackers
       acquire:
         - label: meta.com
           url: https://www.meta.com/quest/
--- a/docs/source/device/body_tracking.rst
+++ b/docs/source/device/body_tracking.rst
@@ -28,9 +28,7 @@ and the BD skeleton format used on the server.
 
    Meta Quest 3/3S also expose body tracking to WebXR via inside-out body
    tracking (IOBT), and the CloudXR client will stream it under the WebXR
-   Body Tracking extension. However, this can not currently be used for full
-   body control as a retargetting of the WebXR skeleton to the robot's full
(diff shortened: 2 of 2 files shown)


docs/source/_data/devices.yaml
```diff
@@ -79,6 +79,10 @@ devices:
         - Isaac Teleop Web Client — runs in the headset browser
       requirements:
         - No additional software
+        - Not supported for full-body robot control
```
yanziz-nvidia: I don't think we shall explicitly call this out? Looking at other patterns we only say what's supported (modes above already mentioned)

docs/source/device/body_tracking.rst
```diff
@@ -27,10 +27,10 @@ and the BD skeleton format used on the server.
 .. note::
 
    Meta Quest 3/3S also expose body tracking to WebXR via inside-out body
```
yanziz-nvidia, line 29: should we just remove the description?

docs/source/_data/devices.yaml
jiwenc-nv: +1.  No need to explicitly statement things that are out of scope.

jiwenc-nv (changes requested):

docs/source/device/body_tracking.rst
jiwenc-nv: agree. no need to describe the implementation details in the doc here.

docs/source/device/body_tracking.rst
```diff
@@ -27,10 +27,10 @@ and the BD skeleton format used on the server.
...
-   tracking (IOBT), and the CloudXR client will stream it. However the Quest
```
jiwenc-nv, line 30: just remove the this "However the Quest..." wording here.

docs/source/device/body_tracking.rst
```diff
@@ -27,10 +27,8 @@ and the BD skeleton format used on the server.
...
-   skeleton is mapped to the PICO BD 24-joint layout in the current version,
-   and accuracy is lower since there are no physical trackers. See
-   `Quest Body Tracking (Limited Support)`_ for details.
+   tracking (IOBT), and the CloudXR client will stream it under the WebXR
```
yanziz-nvidia, line 30: this means a dependency to release next version right? Instead of pre-announcing this, I think we can decouple into two parts, one is to remove the things not supported (thus just remove here); and two we can add a new PR when next version is about to merge and release and then add this description.

PR #1072, opened 2026-09-02
docs: correct the viz agent notes and Televiz page against the code

ivany-nv:
Seventeen claims across two viz docs that the code contradicts. Each was checked against the tree at `2a0854c16`; none needed an author decision. Two commits, one per file, so either can be dropped cleanly.

## Why now

The 2026-08-22 doc audit found 13 of 68 falsifiable claims false in `src/viz/AGENTS.md` and deliberately left them — viz was mid-refactor and a correction would have gone stale before review. That refactor has been quiet since 2026-08-24 and the file has not been touched since, so the findings were re-verified against today's code: **12 still stood, 1 had self-healed, and 5 more turned up.**

`televiz.rst` was nominally in scope of that audit but produced one incidental finding. 729 lines describing a module whose own agent notes were 19% false is not credible as clean, so it was re-read on its own rather than batched. ~170 claims checked, 5 defects — 3 here, 2 held back below.

## `src/viz/AGENTS.md` (14)

**Build and CI** — misleads anyone reasoning about either:

- `BUILD_VIZ` documented as a hard default `OFF` "matching `BUILD_PLUGIN_OAK_CAMERA`". It auto-detects — `ON` when Vulkan and CUDAToolkit are both found (`CMakeLists.txt:74-86`) — and the cited analogue is the opposite convention (`:62`).
- `BUILD_VIZ_XR` documented as the option gating `viz_xr`, in two places. **No such option exists**; `add_subdirectory(xr)` is unconditional (`src/viz/CMakeLists.txt:35`).
- CI documented as running `ctest -L gpu`. It never has — the GPU job invokes each packaged binary with the Catch2 `[gpu]` filter (`build-ubuntu.yml:435,440`) because it downloads bare executables with no build tree. This **contradicted the file's own CI section**.

**Module boundaries and the dependency DAG** — what an agent consults this file for before touching a `CMakeLists`:

- `viz_layers` links PUBLIC `viz::core` + `viz::session`, PRIVATE `viz::shaders` (`layers/cpp/CMakeLists.txt:29-35`). The doc had the edge running the other way and placed `LayerBase` in `viz/layers/`; it is in `viz/session/`.
- `viz_session` is not offscreen-only and does not link `viz_layers`. Offscreen + window/GLFW + XR, plus `Swapchain` and `TileLayout`; links `viz::core`, `viz::xr`, `oxr::oxr_utils`, `glfw`.
- `viz_xr` does not do swapchain wrapping or type conversion — both live in `viz_session`'s `XrBackend`. It does own the frame loop.
- `viz/xr/` does not expose `xr_backend.hpp`; that is a `viz_session` header.
- There is **no `viz::detail` namespace** anywhere under `src/viz`.
- CUDA links `CUDA::cudart_static`, not `CUDAToolkit::cudart` — the static choice is deliberate, so the reason is now stated.

**Layers** — left behind by the `ImageLayerBase` refactor:

- Five types ship, not one. The mailbox moved off `QuadLayer` onto the shared `ImageLayerBase` and is `kSlotCount = kMaxFramesInFlight + 2`, not three slots. `ProjectionLayer` sits directly on `LayerBase`.
- The deferred zero-copy note said the change would be "local to `QuadLayer`"; it would now touch `ImageLayerBase` and the layers on it.
- It pointed at `DESIGN.md` for open questions. No such file exists.

**One broken link introduced by #980** while it fixed the fixture-layer paths: `../../../tests/` is above the repo root from `src/viz/`. Two occurrences, now `../../tests/`, matching what `src/core/AGENTS.md:8` already uses.

## `docs/source/getting_started/televiz.rst` (3)

- The `VizSessionConfig` table showed "—" for seven Defaults and marked `mode` "Required." All seven have working in-class defaults and a bare `VizSessionConfig()` yields a valid 1024×1024 offscreen session. The table already showed real defaults for two other fields, so the dashes read as "no default" by contrast.
- `format` described as "typically `kRGBA8`". It is the **only** accepted value — anything else throws (`image_layer_base.cpp:59-62`), and `PixelFormat` has two enumerators. The page's own "Submitting frames" bullet already said flatly RGBA8.
- The install section listed `glslangValidator` as part of the `BUILD_VIZ` auto-enable condition. Only Vulkan and CUDA gate the default; a missing `glslangValidator` **fails configuration outright** rather than dropping the module (`CheckBuildDeps.cmake:60-63,90-91`). The sentence inverted the failure mode and contradicted this page's own C++ section.

## Held back — these need an author call, not a doc fix

Two findings from the same pass are **not** in this PR, because in each case the doc describes something the code does not do and only the owner knows whether that is unfinished or abandoned:

- **`FrameTimingStats`**: four of its six fields are never written. `update_timing_stats` assigns only `avg_frame_time_ms` and `render_fps` (`viz_session.cpp:315-326`), so `target_fps`, `missed_frames`, `gpu_time_ms` and `stale_layers` are permanently `0` — and the page's worked example branches on two of them. `examples/camera_viz/pipeline/runner.py:261-264` already guards all four with truthiness tests.
- **`SessionState`**: `kStopping` and `kLost` are never assigned. `state_` is set in exactly three places — `kReady`, `kRunning`, `kDestroyed`. The XR state machine is tracked inside `OpenXrSession` and never propagated, so an XR app polling `get_state()` for session loss, as the page instructs, will never see it.

Raising both separately.

## Verification

`sphinx-build -W --keep-going` over `docs/source` — the command `make current-docs` and CI run: **build succeeded, zero warnings.** Repo pre-commit suite passes on both commits. Docs only; no code, no build config.

## Summary by CodeRabbit

- **Documentation**
  - Updated build instructions to note the `glslangValidator` requirement and the option to disable VIZ builds when unavailable.
  - Documented default display, window, XR, clear-color, and GPU-timing settings.
  - Clarified supported texture layer formats and resulting validation behavior.
  - Expanded contributor guidance for VIZ modules, dependencies, build detection, namespaces, include paths, CI, and OpenXR integration.

### DIFF
--- a/docs/source/getting_started/televiz.rst
+++ b/docs/source/getting_started/televiz.rst
@@ -38,8 +38,10 @@ The published wheels (Linux x86_64 / aarch64, CPython 3.11–3.13) bundle the co
    import isaacteleop.viz as televiz
 
 You only need to :doc:`build from source </getting_started/build_from_source/index>` when
-developing Isaac Teleop itself — that build enables Televiz automatically when Vulkan, the CUDA
-Toolkit, and ``glslangValidator`` are present.
+developing Isaac Teleop itself — that build enables Televiz automatically when Vulkan and the
+CUDA Toolkit are detected. ``glslangValidator`` is additionally required to compile the shaders:
+if it is missing, configuration fails rather than quietly dropping the module — install it, or
+pass ``-DBUILD_VIZ=OFF``.
 
 Overview
 --------
@@ -138,10 +140,11 @@ Session configuration
      - Default
      - Description
    * - ``mode``
-     - —
-     - ``DisplayMode.kXr`` / ``kWindow`` / ``kOffscreen``. Required.
+     - ``DisplayMode.kOffscreen``
+     - ``DisplayMode.kXr`` / ``kWindow`` / ``kOffscreen``. Set it explicitly; the default is
+       offscreen.
    * - ``window_width`` / ``window_height``
-     - —
+     - ``1024``
      - Render size for window and offscreen modes. Ignored in XR (the runtime dictates per-eye
        resolution; query it with ``get_recommended_resolution()``).
    * - ``app_name``
@@ -153,16 +156,17 @@ Session configuration
        components (e.g. ``TeleopSession`` trackers) need them. Televiz already enables its own
        rendering extensions. See `Sharing the XR session`_.
    * - ``xr_near_z`` / ``xr_far_z``
-     - —
+     - ``0.05`` / ``100.0``
      - Near / far planes for the XR projection.
    * - ``xr_system_wait_seconds``
-     - —
+     - ``0``
      - How long to wait for the OpenXR system (headset) to become available at create time.
+       Zero fails fast when no headset is present.
    * - ``clear_color``
-     - —
-     - Background color as an ``(r, g, b, a)`` sequence in ``[0, 1]``.
+     - ``(0, 0, 0, 0)``
+     - Background color as an ``(r, g, b, a)`` sequence in ``[0, 1]``. Transparent by default.
    * - ``gpu_timing``
-     - —
+     - ``False``
      - Enable GPU timestamp queries, surfaced via ``get_gpu_timing()``.
 
 Construct the session with the factory; never call the class directly:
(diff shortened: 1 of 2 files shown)


farbod-nv (approved):
thanks

PR #1083, opened 2026-09-08
Merge changes for adding xsens fullbody support to Nvidia Isaac Teleo…

P-B-M:
## Description

Gives Isaac Teleop an Xsens route: MVN Studio's "Isaac Teleop" network-streamer
preset -> UDP -> a tensor collection -> the stock `FullBodyTracker`.

**The reader is a vendor row, not a new tracker type.** `FullBodyTracker` is already
vendor-dispatched and `FullBodySource` already threads a `TrackerVendor` through the retargeting
engine, so `body.xsens` needs no new pybind type, no new Python source node, and works with stock
`FullBodySource`:

```python
FullBodySource("body", vendor=TrackerVendor("body.xsens", {
    "collection_id": "xsens_full_body", "max_flatbuffer_size": "4096"}))
```

**The plugin needs no vendor SDK.** MVN Studio converts its own 23-segment skeleton to the
vendor-neutral 24-joint `XR_BD_body_tracking` layout and emits a `core::FullBodyPose` FlatBuffer --
the schema this repo already defines -- so the plugin forwards the payload bytes verbatim rather
than re-serialising.

```bash
xsens_full_body_plugin --collection-id=xsens_full_body --address=0.0.0.0 \
    --port=9764 --max-flatbuffer-size=4096
```

Every flag shown is its default, so a bare invocation matches MVN's preset. `--address` picks the
interface to bind; `--address=127.0.0.1` confines the pusher to loopback. The older positional form
still parses, warns, and cannot be mixed with flags.

### Notes for reviewers

- **22 of 24 joints valid is correct.** MVN has no finger tracking, so `LEFT_HAND` and `RIGHT_HAND`
  carry a copy of the wrist pose flagged invalid, and `all_joint_poses_tracked` is structurally
  always false.
- **A sequence gap is not proof of packet loss:** `seq` is sent-only, and MVN skips it for frames it
  declines to build. A step back to 0 is a new MVN session.
- **A sample time moving backwards is a timeline rewind** (scrubbing or restarting a recording), not
  a stale frame, and is never dropped -- doing so discarded every frame of a looped playback.
- The header's `sample_time_ns` is on MVN's send-host clock, so it is not published as the local
  common clock; the plugin stamps that itself and forwards MVN's device clock verbatim alongside it.

## Type of change

- [ ] Documentation update

## Testing

**Platform:** Linux (Ubuntu, x86_64), GCC C++20, Isaac Teleop main, CloudXR runtime 6.3.0-rc4.
The plugin is Linux-only and its CMake target degrades to a skip message elsewhere.

**Unit tests** -- 242 assertions across three suites, all registered with ctest and all run by the
default build (`BUILD_TESTING` is `ON`):

```bash
cmake -S . -B build && cmake --build build -j"$(nproc)"
ctest --test-dir build -R 'xsens_frame_decision|xsens_plugin_options|vendor validation' \
    --output-on-failure
```

- `xsens_frame_decision` -- 114 assertions. Depends on nothing but flatbuffers and the schema, so
  it also builds standalone without a configured tree (the `g++` line is in the plugin README).
- `xsens_plugin_options` -- 120 assertions. Needs no runtime at all.
- `vendor validation` -- 22 assertions in the existing suite, 8 of them new for `body.xsens`.

**Mutation-checked.** 16 deliberate breaks in the frame logic, all 16 caught. The vendor row was checked against both ways
it can be mis-registered, each previously silent: dropping the `validate_vendor` delegation fails 4
assertions (params fall through to "vendor params are not supported yet"), and dropping the
dispatch entry fails 5 (the id becomes unknown).

**Live, against MVN Studio and a real CloudXR runtime:**

- Playing a recording: 22/24 joints valid, pelvis Y steady at hip height with X/Z moving, zero
  malformed, unverified or stale frames over the run.
- Runtime killed mid-stream and restarted: recovered, stream resumed.
- Runtime left down: gave up cleanly, exit 1 with counters printed.
- 3 forced hard `recv` errors: re-bound, stream unbroken.
- Duplicate seq 0: `delivered=7 stale=1 resets=1`, where it used to be `delivered=8 resets=2`.
- Sample time +1 ns: delivered, not dropped.
- The published payload hash across a clean run is unchanged by the refactor.

**Linter/formatter:** `SKIP=check-copyright-year pre-commit run --all-files` clean, and
`clang-format --dry-run --Werror` clean on all 15 changed C++ files.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an Xsens MVN full-body tracking plugin that streams validated pose data over UDP into OpenXR.
  * Added configurable collection IDs, network addresses, ports, and payload-size limits.
  * Added automatic recovery for network interruptions and runtime publishing failures.
  * Added telemetry counters and diagnostics for delivered, rejected, dropped, and recovered frames.

* **Documentation**
  * Added setup, configuration, protocol, diagnostics, and troubleshooting documentation.

* **Tests**
  * Added coverage for frame validation, sequencing, timestamps, options, and vendor configuration.


src/plugins/xsens_full_body/main.cpp
```diff
@@ -0,0 +1,140 @@
...
+        return 1;
+    case ParseOutcome::Ok:
+        break;
+    }
+
+    if (options.used_legacy_positionals)
+    {
+        std::cerr << argv[0] << ": warning: the positional argument form is deprecated; use"
+                  << " --collection-id=, --port= and --max-flatbuffer-size= instead" << std::endl;
+    }
```
aristarkhovNV, line 87: perhaps shouldn't introduce legacy path from day 0

aristarkhovNV (approved):

PR #1089, opened 2026-09-11
Add a unified Python + C++ logging system

lotusl-code:
# Unified logging for Python and C++

Every process this project starts — the host's interpreter, each pybind11
extension, each `fork()`+`execvp()`'d plugin, the detached CloudXR service — now
feeds one logger tree rooted at `isaacteleop`, with one line format, one console
threshold, and one set of files per session. Output that no logger can reach,
because a vendor library formats its own lines onto a file descriptor, is
captured separately and never takes the host's terminal away from it.

37 commits, 187 files, +5418 / −722, based at `f29bff1a6`.

## Problem

Before this branch, diagnostics were emitted by whatever was nearest to hand:
`std::cout` and `printf` in C++, `print()` in Python, `logging.basicConfig()` in
three entry points that each configured the root logger differently, and a
`dup2()` in `cloudxr/runtime.py` that sent file descriptor 1 to `/dev/null` and
descriptor 2 to a file of its own. There was no way to raise or lower verbosity
for a run, no persistent record of a session, and no way to see what a plugin
subprocess had said. The loggers that did exist used bare names such as
`"robot_viz"` and `"wss-proxy"`, which are siblings of `isaacteleop` in Python's
hierarchy rather than descendants, so no handler attached to this tree could
ever see their records.

## Module architecture

```mermaid
flowchart TB
    subgraph PY["Python half — src/python/isaacteleop/logging_config"]
        direction TB
        setup["_setup.py<br/>install() — runs once on import"]
        console["_console.py<br/>the one console StreamHandler"]
        filemod["_file.py<br/>the one RotatingFileHandler<br/>10 MiB x 5, always DEBUG and below"]
        fwd["_forwarding.py<br/>leader receiver / child handler"]
        nfd["_native_fd.py<br/>raw fd 1 / fd 2 capture"]
        napi["_native_api.py<br/>host-facing capture controls"]
        coremod["_core.py<br/>TRACE=5, log_dir(), level names,<br/>private-directory creation"]
        setup --> console
        setup --> filemod
        setup --> fwd
        setup --> nfd
        napi --> nfd
        console --> coremod
        filemod --> coremod
        fwd --> coremod
        nfd --> coremod
    end

    subgraph CPP["C++ half — src/core/log_bridge"]
        direction TB
        hdr["cpp/inc/log_bridge/logger.hpp<br/>the only public header<br/>Logger::get(), LoggerKind"]
        lcore["log_bridge_core — STATIC<br/>logger.cpp · sink_config.cpp<br/>socket_sink.cpp · level_mapping.cpp<br/>no pybind11, no Python.h"]
        lpy["log_bridge_py → _log_bridge<br/>python_bridge_sink.cpp<br/>install_python_sink()"]
        hdr --- lcore
        lpy -->|links| lcore
    end

    spdlog["spdlog v1.17.0<br/>deps/third_party"]
    consumers["~20 targets<br/>src/core · src/viz · 13 plugins<br/>5 pybind11 extensions"]

    spdlog --> lcore
    consumers -->|target_link_libraries| lcore
    fwd <-.->|"same wire format —<br/>change both halves together"| lcore

    classDef pyStyle fill:#e8f0fe,stroke:#4285f4
    classDef cppStyle fill:#fef7e0,stroke:#f9ab00
    class setup,console,filemod,fwd,nfd,napi,coremod pyStyle
    class hdr,lcore,lpy cppStyle
```

The `cpp/` and `python/` halves are deliberately disjoint. A standalone plugin
executable or a Catch2 binary links `log_bridge_core` and can never reach the
Python bridge, which is what makes it safe for `install_python_sink()` to
*replace* a logger's sinks rather than add to them: the two populations are
structurally separate, not switched by policy.

When `BUILD_PLUGIN_OAK_CAMERA` is on, spdlog is built against vcpkg's `fmt`, so
spdlog and DepthAI share one `fmt` instead of colliding.

## Data flow, part 1 — records

The first process of a session — the "leader" — owns the real console and file
handlers and starts a receiver on a Unix domain socket, publishing its path in
`ISAACTELEOP_LOG_SOCKET`. Every process that inherits that variable is a
"forwarding child": instead of local handlers it gets one handler that ships each
record to the leader, which re-emits it through its own tree. Only the leader
ever touches a terminal or a log file for `isaacteleop` records.

```mermaid
flowchart LR
    subgraph CHILD["Forwarding child — inherited ISAACTELEOP_LOG_SOCKET"]
        direction TB
        cpy["isaacteleop.* Python record"]
        ccpp["isaacteleop::Logger record"]
        cfh["ForwardingHandler<br/>_forwarding.py"]
        csink["SocketForwardSink<br/>socket_sink.cpp"]
        cpy --> cfh
        ccpp --> csink
    end

    frame["4-byte big-endian length prefix<br/>+ UTF-8 JSON<br/>name · levelno · msg · created · process · exc_text"]
    cfh --> frame
    csink --> frame

    subgraph LEADER["Leader process — started the receiver, published the path"]
        direction TB
        rx["receiver thread<br/>ThreadingUnixStreamServer"]
        lpy["isaacteleop.* Python record"]
        lcpp["isaacteleop::Logger record<br/>in a pybind11 extension"]
        tree["the isaacteleop logger tree"]
        ch["console handler"]
        fh["rotating file handler"]
        lpy --> tree
        rx --> tree
        tree --> ch
        tree --> fh
    end

    frame -->|"AF_UNIX SOCK_STREAM<br/>best-effort, dropped if unreachable"| rx
    lcpp -->|"SocketForwardSink →<br/>this process's own receiver"| rx

    term(["terminal"])
    recfile[("YYYYmmdd-HHMMSS.isaacteleop.PID.log")]
    ch --> term
    fh --> recfile

    classDef sink fill:#e6f4ea,stroke:#34a853
    class term,recfile sink
```

Both senders map severities through the same table, so a forwarded record cannot
change level in transit. `exc_text` comes from the Python sender only: a
traceback object does not survive JSON, so it is rendered before it is sent.

A C++ process with no interpreter at all forwards exactly like a Python child
does, because `sink_config.cpp` keys off the same variable and speaks the same
frame. Note the in-process case above: a record from a pybind11 extension also
travels over the socket, to a receiver in its own process.

Where Unix domain sockets do not exist (Windows), `socket_path()` answers `None`,
every process takes the leader branch and keeps its own console and file
handlers. That is the pre-forwarding behaviour and a correct degradation.

## Data flow, part 2 — non-logger output

Some diagnostics cannot be routed through a logger at all. The CloudXR/Monado
OpenXR runtime and the Manus SDK format their own lines and write them straight
to a descriptor; the runtime exports no log hook and does not implement
`XR_EXT_debug_utils`. The descriptor is the only seam.

This library does not take its host's descriptors. The split is by ownership:
fd 1 and fd 2 of the *host* are rebound only inside an explicitly entered block
that restores them; a process this library *launched* has its descriptors set at
launch, because they are not the host's.

```mermaid
flowchart LR
    subgraph HOST["Host process — fd 1 / fd 2 belong to the host"]
        direction TB
        hstream["print()<br/>sys.stderr.write()<br/>uncaught tracebacks"]
        hraw["raw os.write(1,2)<br/>host C libraries<br/>subprocesses the host starts"]
        itnative["OpenXR runtime writes<br/>during session construction<br/>and teardown"]
    end

    gate{"is a<br/>capture_native_output()<br/>block open?"}

    hstream -->|"stream objects are moved<br/>onto terminal duplicates"| term
    hraw --> gate
    itnative -->|"TeleopSession always<br/>holds one open here"| gate

    gate -->|"no — the default state<br/>fd 1 / fd 2 untouched"| term
    gate -->|"yes — dup2 in, dup2 out on exit"| nat

    subgraph LAUNCHED["Processes this library launched — their fds are ours to set"]
        direction TB
        plug["plugin process<br/>after fork(), before execvp():<br/>open() + dup2() on its own fds"]
        svc["CloudXR service worker<br/>Popen(stdout=native_capture_fd())"]
    end
    plug --> nat
    svc --> nat

    term(["terminal"])
    nat[("YYYYmmdd-HHMMSS.isaacteleop.PID.native.log<br/>opened O_APPEND — one file for both<br/>descriptors, so interleaving survives")]
    nat -.->|"echoed live only when the<br/>console threshold is TRACE"| term

    classDef sink fill:#e6f4ea,stroke:#34a853
    class term,nat sink
```

A file, never a pipe: a pipe refuses writes past 64 KiB until a reader drains it,
a reader in this process needs the GIL between reads, and the native call doing
the writing holds it — `oxr_bindings.cpp` releases none — so the two deadlock. A
write to a file needs nothing else to run.

Captured bytes are always persisted. The `open()` in the plugin child is between
`fork()` and `execvp()`, where only async-signal-safe calls are legal;
`open`/`dup2`/`close` are on that list and `getenv()` is read before the fork.

## Where each kind of output ends up

For a leader process running a `TeleopSession`. "terminal" means the fd 1 / fd 2
the session started with.

| Source | Mechanism | Destination |
| --- | --- | --- |
| Host application's `print()` | Python stream object | terminal |
| Host's raw `os.write(1/2, …)` | descriptor | terminal, except inside an open capture block |
| Host's own subprocesses | inherited descriptors | terminal, except when started inside an open capture block |
| `isaacteleop.*` Python record | logger | terminal + `…isaacteleop.PID.log` |
| `isaacteleop::Logger`, in-process C++ | spdlog → `SocketForwardSink` → leader's receiver | terminal + `…isaacteleop.PID.log` |
| `isaacteleop::Logger`, plugin process | spdlog → socket → leader's receiver | terminal + `…isaacteleop.PID.log` |
| CloudXR / Monado runtime, in-process | descriptor, inside a capture block | `…isaacteleop.PID.native.log` |
| Manus SDK, in the plugin process | descriptor, child-side `dup2` | the same `.native.log` |
| CloudXR service worker | `stdout=native_capture_fd()` | the same `.native.log` |

Two files per session, both under `log_dir()`. A capture file nothing ever wrote
to is removed at exit.

`log_dir()` is `/tmp/isaacteleop-<uid>/logs` by default — per-uid, created
`0700`, every component checked for ownership — or whatever `ISAACTELEOP_LOG_DIR`
names. The C++ half resolves the same pair identically.

One line format, both halves:

```
[2026-09-15 12:22:36.448] [INFO ] [isaacteleop.cloudxr.service] [pid:12345] message
```

## User interfaces

Nothing has to be called to get working logging: `isaacteleop/__init__.py`
installs the handlers once, on import. The public API is for changing what that
default does.

```mermaid
flowchart LR
    subgraph API["isaacteleop.logging_config — public API"]
        direction TB
        k1["set_console_level(level)"]
        k2["set_console_filter(pattern, target)"]
        k3["set_logger_colors(mapping)"]
        k4["set_propagate_to_root(enabled)"]
        k5["capture_native_output()"]
        k6["set_native_capture_mode(mode)"]
        k7["set_native_echo(enabled)"]
        k8["native_capture_path()<br/>native_capture_fd()<br/>native_capture_mode()"]
    end

    ch["console handler<br/>threshold · keyword filter · colour"]
    fh["rotating file handler<br/>always DEBUG and below —<br/>never affected by the console knobs"]
    root["link to Python's root logger<br/>on by default"]
    cap["raw fd 1 / fd 2 capture<br/>off · scoped · process"]
    mir["live echo of captured bytes"]
    intro["read-only: where the capture file is,<br/>a descriptor on it for your own child"]

    k1 --> ch
    k2 --> ch
    k3 --> ch
    k4 --> root
    k5 --> cap
    k6 --> cap
    k7 --> mir
    k8 --> intro
    k1 -.->|"at TRACE, also turns<br/>the live echo on"| mir
    k1 -.->|"does not touch"| fh

    classDef ro fill:#f1f3f4,stroke:#9aa0a6
    class intro ro
```

`set_native_capture_mode()` applies the transition on return — selecting
`process` rebinds fd 1 and fd 2 there and then, and leaving it restores them —
and exports the choice so child processes inherit it.

### C++

```cpp
#include <log_bridge/logger.hpp>

auto logger = isaacteleop::Logger::get("isaacteleop.<module>.<ClassName>");
logger->info("tracking {} devices", count);
```

Memoized on the name; the same name always returns the same instance.
`LoggerKind::Application` (the default) starts at `debug`;
`LoggerKind::ThirdParty` starts at `trace`, for loggers that wrap vendor output,
so vendor chatter stays silent unless someone lowers the threshold to `TRACE`.
`Logger::get()` returns a bare `std::shared_ptr<spdlog::logger>` — spdlog's own
API is the interface, and there is no wrapper to learn. Link
`log_bridge::log_bridge_core`.

### Environment

Five variables, read identically by both halves. This is the whole external
contract.

| Variable | Meaning |
| --- | --- |
| `ISAACTELEOP_LOG_DIR` | where log files land |
| `ISAACTELEOP_LOG_LEVEL` | console threshold for out-of-process code |
| `ISAACTELEOP_LOG_SOCKET` | set by the leader; its presence is what makes a process forward instead of owning handlers |
| `ISAACTELEOP_NATIVE_CAPTURE` | `off` / `scoped` (default) / `process` |
| `ISAACTELEOP_NATIVE_CAPTURE_FILE` | published by the leader; read by processes with no interpreter so they can point their own stdio at the same file |

## Migration

Roughly 278 raw-output statements across `src/core`, `src/viz` and the plugins
were replaced with about 245 logger calls, and 13 diagnostic `print()` sites in
`src/python/isaacteleop/` moved to `logging`. Loggers using bare names —
`robot_viz`, `wss-proxy`, `oob-teleop-*`, `mujoco_xr`, `camera_streamer` — were
renamed under `isaacteleop.`, and the three `logging.basicConfig()` call sites
were removed.

Deliberate terminal UX was not migrated: CLI usage text, interactive prompts,
operator banners, progress lines a log file would ruin. Two sets of raw-output
sites are exempt for correctness rather than style, and say so in the source:

- the four `std::cerr` sites in `plugin.cpp` between `fork()` and `execvp()`,
  where only async-signal-safe calls are legal, and `Logger::get()` allocates and
  touches spdlog's global registry;
- `schema_compat.cpp`, whose compatibility warning stays on `std::cerr` because
  a test counts how many times it fires by swapping `std::cerr`'s streambuf.

The repository-wide rules are in the root `AGENTS.md` under "Logging", with
subsystem-internal rules in `src/python/isaacteleop/logging_config/AGENTS.md` and
`src/core/log_bridge/AGENTS.md`.

## Behavioural changes to be aware of

- An application that calls `logging.basicConfig()` itself will see `isaacteleop`
  records twice, once in this tree's format and once in its own, because this
  tree propagates to the root by default. `set_propagate_to_root(False)` makes
  this tree the sole route; attach your handler to the `isaacteleop` logger
  rather than to the root.
- Every session writes to `/tmp/isaacteleop-<uid>/logs` unless
  `ISAACTELEOP_LOG_DIR` says otherwise. Files rotate at 10 MiB with five
  backups, per process.
- `cloudxr/runtime.py`'s `dup2()` of fd 1 to `/dev/null` is gone. Output that
  used to be discarded there is now persisted.
- A detached CloudXR service owns its own logs rather than forwarding into a
  leader that may already have exited.

## Known limitations

- **`install_python_sink()` reaches one shared object, not the process.**
  `log_bridge_core` is static and every pybind11 extension carries its own copy
  of the bridge pointer and of spdlog's registry; Python loads extensions
  `RTLD_LOCAL`, so nothing unifies them. Loggers created inside `_oxr` or
  `_robot_twin` still reach Python, by the socket route, which is why the split
  is invisible on Linux; on Windows, where that transport does not exist, those
  modules keep their own console and file sinks instead. Closing it for real
  means giving the extensions shared logging state. Written up in
  `src/core/log_bridge/AGENTS.md`.
- **A capture block is process-wide while it is open.** A descriptor has no
  narrower scope. For the length of a block, a raw descriptor write by another of
  the host's threads is captured too — and a process the host starts inside the
  block keeps the capture file for its whole life, because the restore applies
  only to this process's descriptor table. Blocks are therefore kept around
  native construction and teardown rather than around a whole session, and
  `ISAACTELEOP_NATIVE_CAPTURE=off` removes even that, at the cost of no longer
  persisting in-process vendor output.
- **Whether the OpenXR runtime writes to a descriptor during steady-state
  operation, rather than only during session construction and teardown, has not
  been determined.** If it does, those lines reach the terminal under the default
  mode. The remedy is a host-side capture block around the main loop, or
  `ISAACTELEOP_NATIVE_CAPTURE=process`.


src/python/isaacteleop/logging_config/_native_fd.py
```diff
@@ -0,0 +1,138 @@
...
+        return
+    label = _FD_LABELS[fd]
+    directory = log_dir()
+    directory.mkdir(parents=True, exist_ok=True)
+    timestamp = time.strftime("%Y%m%d-%H%M%S")
+    sink_fd = os.open(
+        directory / f"{timestamp}.isaacteleop.{os.getpid()}.native-{label}.log",
+        os.O_WRONLY | os.O_CREAT | os.O_APPEND,
+        0o644,
+    )
```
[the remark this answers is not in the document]
lotusl-code: No, it's closed at _native_fd.py:44

deps/third_party/CMakeLists.txt
```diff
@@ -98,6 +98,25 @@ set(FLATBUFFERS_BUILD_FLATC ON CACHE BOOL "Build flatc compiler" FORCE)
...
+FetchContent_Declare(
+    spdlog
+    GIT_REPOSITORY https://github.com/gabime/spdlog.git
+    GIT_TAG        v1.17.0
+    GIT_SHALLOW    TRUE
+)
+set(SPDLOG_BUILD_EXAMPLE OFF CACHE BOOL "Skip spdlog examples" FORCE)
+set(SPDLOG_BUILD_TESTS OFF CACHE BOOL "Skip spdlog tests" FORCE)
+set(SPDLOG_INSTALL OFF CACHE BOOL "Skip spdlog install target" FORCE)
+FetchContent_MakeAvailable(spdlog)
```
sgrizan-nv, line 117: With `BUILD_PLUGIN_OAK_CAMERA=ON`, this introduces spdlog’s bundled fmt alongside the external fmt used by DepthAI. DepthAI links both `fmt::fmt` and `spdlog::spdlog`, mixing incompatible headers. The [current CI job](https://github.com/NVIDIA/IsaacTeleop/actions/runs/34557184168/job/103132218447) fails in `fmt/chrono.h`, `fmt/ostream.h`, and `fmt/std.h`. Please configure both consumers to use the same fmt dependency, or reuse compatible vcpkg-provided spdlog/fmt targets.

src/python/isaacteleop/logging_config/_native_fd.py
```diff
@@ -0,0 +1,170 @@
...
+    if fd == 2:
+        sys.stderr = saved
+        console_handler.setStream(saved)
+    else:
+        sys.stdout = saved
+    pump = threading.Thread(
+        target=_pump,
+        args=(fd, read_fd, sink_path),
+        name=f"isaacteleop-native-{label}",
+        daemon=True,
```
sgrizan-nv, line 132: The pipe is drained by a Python thread, which needs the GIL between reads. If native code retains the GIL while filling the pipe, the writer blocks waiting for the drain thread, while the drain thread waits for the writer to release the GIL. Existing OpenXR bindings retain the GIL during native calls. I reproduced this with a GIL-held 1 MiB native write: it completes without capture and hangs with capture enabled. Please make the drain independent of the GIL.

src/python/isaacteleop/logging_config/_forwarding.py
```diff
@@ -0,0 +1,237 @@
...
+        while len(buf) < n:
+            chunk = self.rfile.read(n - len(buf))
+            if not chunk:
+                return None
+            buf.extend(chunk)
+        return bytes(buf)
+
+
+class ThreadingUnixStreamServer(
+    socketserver.ThreadingMixIn, socketserver.UnixStreamServer
```
sgrizan-nv, line 197: `socketserver.UnixStreamServer` is unavailable on Windows, so this unconditional class definition raises `AttributeError` during import. Since `isaacteleop.__init__` imports the logging configuration, this breaks ordinary package imports and the import-based stub generation used during Windows builds. Please guard both the class definition and receiver startup, retaining local console/file logging when Unix sockets are unavailable.

src/python/isaacteleop/logging_config/_setup.py
```diff
@@ -0,0 +1,55 @@
...
+        # gating it in this process still has to key off that same variable.
+        env_level_name = os.environ.get("ISAACTELEOP_LOG_LEVEL")
+        env_level = (
+            _LEVEL_NAMES.get(env_level_name.lower(), logging.INFO)
+            if env_level_name
+            else logging.INFO
+        )
+        _native_fd.gate(env_level, _console.ensure_handler())
+        _forwarding.ensure_handler(socket_path)
+        return
```
sgrizan-nv, line 50: The detached CloudXR service launched by `background.spawn()` inherits `ISAACTELEOP_LOG_SOCKET` from the launcher, so this branch gives it only a forwarding handler. The launcher exits after startup, removing the receiver while the service continues running. Subsequent records are silently dropped, with no local console or file fallback. I reproduced records arriving before launcher exit and disappearing afterward. Please clear the inherited forwarding destination when launching an independent service, or provide a receiver-loss fallback.

src/python/isaacteleop/logging_config/_native_fd.py
```diff
@@ -0,0 +1,170 @@
...
+    read_fd, write_fd = os.pipe()
+    saved = os.fdopen(os.dup(fd), "w", buffering=1)
+    os.dup2(write_fd, fd)
+    os.close(write_fd)
+    _saved[fd] = saved
+    if fd == 2:
+        sys.stderr = saved
+        console_handler.setStream(saved)
+    else:
+        sys.stdout = saved
```
sgrizan-nv, line 127: These assignments replace `sys.stdout` and `sys.stderr` even when the application has installed custom streams. Because capture runs during package import, importing IsaacTeleop inside `contextlib.redirect_stdout(StringIO())` causes subsequent `print()` calls to escape the requested capture; I reproduced that behavior. Notebook and test-runner output streams are affected for the same reason. Please preserve custom Python streams and only replace streams that actually follow the descriptors being redirected.

examples/camera_viz/transports/_nv_encode_gst.py
```diff
@@ -22,7 +22,7 @@
 
 from ._nv12_convert import build_rgba_to_nv12_kernel, launch_rgba_to_nv12
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("isaacteleop.camera_viz.transports._nv_encode_gst")
```
aristarkhovNV, line 25: why not keep the built-in dunder

lotusl-code: In python loggers, the name determines the report chain. Our log system is designed to route all loggers to the root logger in _core.py called "isaacteleop" to apply formatting, filtering, level and so controls.
The loggers in example can't get the "isaacteleop" prefix as other places in the codebase, because examples are not isaacteleop sub-packages. So it can only reach our log system by hardcoding loggernames

deps/third_party/CMakeLists.txt
lotusl-code: Fixed by a2449b04e64d66b7b39068c622f474c736887eba. When BUILD_PLUGIN_OAK_CAMERA=ON, spdlog will reuse the vcpkg fmt.

src/python/isaacteleop/logging_config/_forwarding.py
lotusl-code: Fixed by 36b69d3b3e1de18e54e02c639ed249c76e38c261, added multiple guards on POXIS-only calls

src/python/isaacteleop/logging_config/_setup.py
lotusl-code: Fixed by f9e0ab6118ddb5cdb27289814ca5d3c5946d947a, removed ISAACTELEOP_LOG_SOCKET for cloudxr, its logs will be saved as before (under ~/.cloudxr)

src/python/isaacteleop/logging_config/_native_fd.py
lotusl-code: Fixed by a31b1dad14414e7799e7270cdf32e2a65fb5068f, solved the deadlock by write-read from a log file instead of a pipe. The log file will be deleted in the end if it's empty

src/python/isaacteleop/logging_config/_native_fd.py
lotusl-code: Fixed by a36cc1f85fcd78c5479e815e5cfb7d474915370c, now will only route fd

src/core/log_bridge/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20)
+
+# Core Logger library (static; consumed in-tree via log_bridge::log_bridge_core).
+# No pybind11/Python.h anywhere in this target -- every standalone plugin
+# executable and Catch2 test binary links this without pulling in libpython.
+add_library(log_bridge_core STATIC
```
sgrizan-nv, line 9: `log_bridge_core` is linked statically into `_log_bridge` and the other native extensions, while its bridge pointer lives in module-local static storage. Calling `install_python_sink()` through `_log_bridge` therefore does not configure the copies used by `_oxr`, `_deviceio_session`, etc.

On Unix, the socket fallback can mask this by forwarding records back to the same process. On Windows, the other extensions retain their independent C++ sinks.

I think it makes sense to add an integration test that emits from another compiled extension and verifies that a Python handler receives the record. It's possible that it may be necessary to give the extensions shared logging state, or explicitly install the bridge in each extension

src/python/isaacteleop/logging_config/_console.py
```diff
@@ -0,0 +1,168 @@
...
+        handler.setFormatter(
+            _LoggerNameColorFormatter(LINE_FORMAT, datefmt=DATE_FORMAT)
+        )
+        handler.setLevel(logging.INFO)
+        root = logging.getLogger(ROOT_LOGGER_NAME)
+        root.setLevel(
+            TRACE
+        )  # handlers filter; the logger itself must stay maximally permissive
+        if _forwarding.socket_path() is None:
+            root.addHandler(handler)
```
sgrizan-nv, line 109: The `isaacteleop` logger retains `propagate=True` after installing its own handlers. If the embedding application has configured Python's root logger, records also reach those handlers, causing duplicate output and bypassing IsaacTeleop's console level and filters.

I reproduced this with an application root handler: after `set_console_level("error")`, an IsaacTeleop INFO message still appeared through that handler.

Could we set `propagate=False` when IsaacTeleop owns the handlers, or provide an explicit mode for integrating with application-owned logging? A repository convention against `basicConfig()` cannot cover applications using this library.

src/python/isaacteleop/logging_config/_forwarding.py
```diff
@@ -0,0 +1,267 @@
...
+        return _receiver_socket
+    with _lock:
+        if _receiver_socket is not None:
+            return _receiver_socket
+        directory = ensure_log_dir()
+        timestamp = time.strftime("%Y%m%d-%H%M%S")
+        path = str(directory / f"isaacteleop.{timestamp}.{os.getpid()}.sock")
+        if os.path.exists(path):
+            os.unlink(path)
+        server = ThreadingUnixStreamServer(path, RequestHandler)
```
sgrizan-nv: The socket path includes the entire `ISAACTELEOP_LOG_DIR`, but Unix-domain socket paths have much shorter limits than ordinary filesystem paths.

I reproduced successful log-file creation followed by `OSError('AF_UNIX path too long')` when starting the receiver. Since receiver startup runs during `import isaacteleop`, a valid log-directory setting prevents importing the package.

Could we create the socket in a short, private runtime directory independently of where log files are stored?

sgrizan-nv (commented):
The separation between console handling, file storage, process forwarding, and native output capture makes sense. Keeping the C++ core independent of Python is also appropriate for standalone plugins.

The architectural boundary I would reconsider is logging lifecycle ownership. Importing `isaacteleop` currently creates files, starts a receiver thread, redirects process-wide native stdout/stderr, and changes the environment inherited by children. These effects extend beyond the library's own logging calls.

Could session or application startup explicitly initialize this infrastructure, with a corresponding shutdown operation? Library modules could continue emitting named records, while the application decides when to take ownership of logging and native output capture.

jiwenc-nv (approved):

src/python/isaacteleop/logging_config/_native_fd.py
```diff
@@ -0,0 +1,229 @@
...
+        )
+    except OSError:
+        return  # leave the descriptors on the terminal rather than fail the import
+
+    # Both duplicates taken before either dup2, or the second save would
+    # duplicate the capture file instead of the terminal.
+    for fd in _FD_LABELS:
+        _saved[fd] = os.fdopen(os.dup(fd), "w", buffering=1)
+    for fd in _FD_LABELS:
+        os.dup2(sink_fd, fd)
```
sgrizan-nv: Let's try to avoid redirecting process-wide native stdout/stderr in IsaacTeleop. These `dup2()` calls affect every native library in the host process, plus subprocesses inheriting these descriptors - not just IsaacTeleop or CloudXR diagnostics. Preserving Python's `sys.stdout` and `sys.stderr` does not preserve those native output destinations.

sgrizan-nv (approved):

src/python/isaacteleop/logging_config/_native_fd.py
sgrizan-nv: A concrete in-process example is an unrelated C library calling `printf()`:

```python
import ctypes
import isaacteleop

libc = ctypes.CDLL(None)
libc.printf(b"Output from an unrelated C library\n")
libc.fflush(None)

print("Output from Python", flush=True)
```

After importing IsaacTeleop, the C library’s output is redirected into `*.native.log`, while the Python output remains in the terminal. The same applies to unrelated C extensions and native libraries using fd 1 or 2. This is why preserving `sys.stdout` and `sys.stderr` does not address the process-wide side effect of the `dup2()` calls.

yanziz-nvidia:
reviewed by yanziz-review-bot

**Summary**

Other than the existing findings in previous comments, I found one new startup regression. Logging tests passed in an isolated source harness (32 passed, 3 skipped), and all three packaging checks passed; full C++/hardware validation was not run locally. CI still reports a CloudXR termination-test failure.

| Severity | Finding |
|---|---|
| 🚫 BLOCKER | [`camera_streamer.py:31`](https://github.com/NVIDIA/IsaacTeleop/blob/bde92e7274ab751a6c8bb06d213d61d3dcec1b18/examples/camera_viz/camera_streamer.py#L31) — **Breaks sender-only deployments.** The new unconditional IsaacTeleop import conflicts with `_install_deps.sh`, which deliberately omits IsaacTeleop in `--sender-only` mode. The sender now exits before argument parsing or camera startup. Reproduced with `--help`: `ModuleNotFoundError: No module named 'isaacteleop'`. Preserve standalone logging when IsaacTeleop is unavailable. |
| 💡 SUGGESTION | none |
| 🧹 NIT | none |

- `examples/camera_viz/camera_streamer.py:31` — Make logging initialization work without IsaacTeleop in sender-only installations. Add a startup smoke test using that dependency set.

src/python/isaacteleop/logging_config/_native_fd.py
sgrizan-nv: An equivalent example using an actual CPython C extension:

```c
// native_noise.c
#include <Python.h>
#include <stdio.h>

static PyObject *emit(PyObject *self, PyObject *args) {
    printf("Output from an unrelated C extension\n");
    fflush(stdout);
    Py_RETURN_NONE;
}

static PyMethodDef methods[] = {
    {"emit", emit, METH_NOARGS, "Write to native stdout."},
    {NULL, NULL, 0, NULL}
};

static struct PyModuleDef module = {
    PyModuleDef_HEAD_INIT, "native_noise", NULL, -1, methods
};

PyMODINIT_FUNC PyInit_native_noise(void) {
    return PyModule_Create(&module);
}
```

```python
# setup.py
from setuptools import Extension, setup

setup(
    name="native-noise",
    ext_modules=[Extension("native_noise", ["native_noise.c"])],
)
```

```bash
python setup.py build_ext --inplace

python - <<'PY'
import isaacteleop
import native_noise

native_noise.emit()
print("Output from Python", flush=True)
PY
```

The Python line remains visible, while the unrelated extension’s `printf()` output is redirected into IsaacTeleop’s `*.native.log`.

PR #1090, opened 2026-09-11
ROS: separate hand provider from plugin lifecycle

sgrizan-nv:
## Description

Separate hand-tracking provider identity from plugin process lifecycle. Native tracking uses controller aim poses for controller teleop, while MANUS and Wuji use their provider-injected wrist poses so provider-owned calibration is preserved.

Replace `hand_tracking_plugin` with `hand_tracking_provider` and `start_hand_tracking_plugin`, preserve provider selection during MCAP replay, reject plugin startup during replay, and require plugin search paths only for managed startup. Update the MCAP matrix and verifier to validate provider-specific EE routing.

## Type of change

- [x] Documentation update

## Testing

- Built `examples/teleop_ros2/Dockerfile` with the `test` target.
- Ran the full teleop ROS 2 test suite: 60 passed.
- Ran all 9 MCAP scenarios, including native, MANUS, and Wuji provider routing.
- Verified formatting, Python syntax, and `git diff --check`.

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have added tests that prove my fix/feature works
- [x] I have signed off all my commits per the DCO

## Summary by CodeRabbit

- **New Features**
  - Added separate hand-tracking provider selection for Native, Manus, and Wuji tracking.
  - Added independent control over whether the hand-tracking plugin starts.
  - Updated teleoperation profiles and topic verification for provider-specific wrist and hand end-effector behavior.
  - Expanded ROS2 replay and live CloudXR scenarios across supported providers.

- **Documentation**
  - Updated teleoperation setup instructions, parameters, Docker examples, and Manus/Wuji startup guidance.

- **Bug Fixes**
  - Added validation for incompatible provider, replay, tracking-mode, and plugin-path configurations.
  - Preserved Manus calibration in controller end-effector output.

### DIFF
--- a/examples/teleop_ros2/python/geometry.py
+++ b/examples/teleop_ros2/python/geometry.py
@@ -22,7 +22,6 @@ def apply_manus_controller_to_hand_pose(pose: Pose, side: str) -> Pose:
     if side not in ("left", "right"):
         raise ValueError(f"side must be 'left' or 'right', got {side!r}")
 
-    # All MANUS calibration data is intentionally kept in this one function.
     hand_left_pico_rotation = np.array(
         [
             [-0.91777945, -0.18672461, -0.35044942],
@@ -54,14 +53,12 @@ def apply_manus_controller_to_hand_pose(pose: Pose, side: str) -> Pose:
             pose.orientation.w,
         ]
     )
-
     controller_to_hand_rot = Rotation.from_matrix(controller_to_hand_rot_mat)
 
     world_hand_pos = world_controller_pos + world_controller_rot.apply(
         controller_to_hand_trans
     )
     world_hand_rot = world_controller_rot * controller_to_hand_rot
-
     return to_pose(world_hand_pos, world_hand_rot.as_quat())
 
 
--- a/examples/teleop_ros2/python/geometry.py
+++ b/examples/teleop_ros2/python/geometry.py
@@ -22,6 +22,7 @@ def apply_manus_controller_to_hand_pose(pose: Pose, side: str) -> Pose:
     if side not in ("left", "right"):
         raise ValueError(f"side must be 'left' or 'right', got {side!r}")
 
+    # All MANUS calibration data is intentionally kept in this one function.
     hand_left_pico_rotation = np.array(
         [
             [-0.91777945, -0.18672461, -0.35044942],
@@ -53,12 +54,14 @@ def apply_manus_controller_to_hand_pose(pose: Pose, side: str) -> Pose:
             pose.orientation.w,
         ]
     )
+
     controller_to_hand_rot = Rotation.from_matrix(controller_to_hand_rot_mat)
 
     world_hand_pos = world_controller_pos + world_controller_rot.apply(
         controller_to_hand_trans
     )
     world_hand_rot = world_controller_rot * controller_to_hand_rot
+
     return to_pose(world_hand_pos, world_hand_rot.as_quat())
 
 
--- a/examples/teleop_ros2/python/node_parameters.py
+++ b/examples/teleop_ros2/python/node_parameters.py
@@ -21,12 +21,12 @@ from pathlib import Path
 import numpy as np
 from constants import (
     HAND_RETARGETERS,
-    HAND_TRACKING_PLUGINS,
+    HAND_TRACKING_PROVIDERS,
     TELEOP_MODES,
     TRACKED_HAND_RETARGETERS,
     WUJI_HAND_MODELS,
     HandRetargeter,
-    HandTrackingPlugin,
+    HandTrackingProvider,
     TeleopMode,
     resolve_hand_retargeter,
 )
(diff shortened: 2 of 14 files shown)


examples/teleop_ros2/python/geometry.py
```diff
@@ -22,7 +22,6 @@ def apply_manus_controller_to_hand_pose(pose: Pose, side: str) -> Pose:
     if side not in ("left", "right"):
         raise ValueError(f"side must be 'left' or 'right', got {side!r}")
 
-    # All MANUS calibration data is intentionally kept in this one function.
```
kajananchinniahNV, line 25: Nit why was this comment deleted

examples/teleop_ros2/python/geometry.py
```diff
@@ -54,14 +53,12 @@ def apply_manus_controller_to_hand_pose(pose: Pose, side: str) -> Pose:
             pose.orientation.w,
         ]
     )
-
```
kajananchinniahNV, line 57: Nit: the extra space deletions are probably not necessary

examples/teleop_ros2/python/node_parameters.py
```diff
@@ -344,73 +345,94 @@ def _load_hand_retargeter(
...
-        "hand_tracking_plugin",
-        HandTrackingPlugin.NONE.value,
+        "hand_tracking_provider",
+        HandTrackingProvider.NATIVE.value,
         ParameterDescriptor(
             description=(
-                "Optional hand-tracking input plugin managed by this node. "
-                "Use 'none' for native OpenXR or a manually launched plugin. "
-                "Valid managed plugins: 'manus' or 'wuji'."
+                "Hand-tracking data provider. Use 'native' for native OpenXR "
```
kajananchinniahNV: Update description to also note that native for controllers does nothing?

examples/teleop_ros2/python/node_parameters.py
```diff
@@ -344,73 +345,94 @@ def _load_hand_retargeter(
...
         )
 
-    node.get_logger().info(f"Managed hand-tracking plugin: {plugin}")
-    return plugin, search_paths
+    search_paths: tuple[Path, ...] = ()
+    if start_plugin:
+        search_paths = tuple(
+            dict.fromkeys(
+                Path(value).expanduser().resolve()
+                for value in os.environ.get("ISAAC_TELEOP_PLUGIN_PATH", "").split(
```
kajananchinniahNV, line 419: Nit: I think Isaac Teleop Plugin Path being a parameter might make more sense; however I'm currently using it as an environment vaiable myself so I'm fine either way

examples/teleop_ros2/python/node_parameters.py
```diff
@@ -344,73 +345,94 @@ def _load_hand_retargeter(
...
             "does not consume OpenXR hand tracking"
         )
 
-    if not any(path.is_dir() for path in search_paths):
-        raise FileNotFoundError(
-            "No existing plugin search directory was found in ISAAC_TELEOP_PLUGIN_PATH"
+    if start_plugin and provider == HandTrackingProvider.NATIVE:
+        raise ValueError(
+            "Parameter 'start_hand_tracking_plugin' requires "
+            "hand_tracking_provider:=manus or hand_tracking_provider:=wuji"
```
kajananchinniahNV: Do we need to fail? If it's set to native and hand tracking is set to start, we should just ignore it right? I'm fine with the explicit check if safety / debugging is preferred (in case a param doesn't update or whatever)

examples/teleop_ros2/python/node_parameters.py
sgrizan-nv: Agreed that a parameter could be useful, but I am keeping ISAAC_TELEOP_PLUGIN_PATH unchanged in this PR to preserve the existing workflow and avoid expanding the scope.

examples/teleop_ros2/python/geometry.py
sgrizan-nv: Good catch—this was unrelated churn. I restored the existing comment in 3a1879986.

examples/teleop_ros2/python/geometry.py
sgrizan-nv: Agreed. I restored the blank lines in 3a1879986 so geometry.py no longer contains those unrelated formatting changes.

examples/teleop_ros2/python/node_parameters.py
sgrizan-nv: The lifecycle option is now use_external_hand_tracking_plugin and defaults to false. With native, the default starts nothing. I kept an explicit error for native plus use_external_hand_tracking_plugin:=true because native has no external plugin and silently ignoring it would hide a configuration mistake.

examples/teleop_ros2/python/node_parameters.py
sgrizan-nv: Updated in 3a1879986. The description now states that native starts no plugin and has no effect when the configuration does not consume tracked hands.

examples/teleop_ros2/python/node_parameters.py
kajananchinniahNV: That sgtm - we can do it in a follow up in a later release since it's nbd

examples/teleop_ros2/python/node_parameters.py
kajananchinniahNV: Ok that makes sense to me, we can keep it like this

kajananchinniahNV (approved):

Issue #888, opened 2026-08-04
Consolidate CloudXR runtime lifecycle management: one supervised-launcher surface for apps, examples, and rig


jiwenc-nv:
## Summary

While reviewing #887 I formed the impression that the viz stack was re-inventing CloudXR lifecycle management. Checking the code, the premise needs one correction: `src/viz/` (the C++ libraries) contains **no** CloudXR lifecycle code at all, and `examples/camera_viz/camera_viz.py` already imports and uses `CloudXRLauncher` (`camera_viz.py:32`, `:394`, `:442`). What it re-invents is not the launcher — it is the **supervision policy around** the launcher: manual `__enter__`/`__exit__`, its own signal handlers, its own health-poll plumbing, its own teardown-ordering rules. So the smell is real, it just lives one layer up from where I first placed it.

Every app that wants a managed CloudXR runtime today has to hand-roll the same four things: start it, hold it open, notice when it dies, tear it down in the right order. `CloudXRLauncher` gives you the first and the last; the middle two are copy-pasted per app, and they have already drifted.

## Evidence

**1. Four independent implementations of "poll the launcher's health from the app's main loop."**

- `src/python/isaacteleop/cloudxr/__main__.py:260-272` — `while not stop: launcher.health_check(); time.sleep(0.1)` with its own SIGINT/SIGTERM flag.
- `examples/teleop_ros2/python/teleop_ros2_node.py:216` and `:225` — one check between session attempts, one per step, each with a comment explaining why the *other* one is insufficient.
- `src/python/isaacteleop/teleop_session_manager/teleop_session.py:922-924` — `_check_plugin_health()`, the same pattern for plugin contexts.
- PR #887 — `VizRunner.wait(health_check=...)`, threading a callable through the runner so the render-thread join loop can poll it.

The protocol name has already diverged: `CloudXRLauncher.health_check()` vs. plugin contexts' `check_health()`. Two names for one concept is a good sign the protocol was never designed, only repeatedly rediscovered.

**2. Signal-handler collision in `camera_viz`.**

[...]

PR #1093, opened 2026-09-12
refactor!: drop the tmux rig launcher

jiwenc-nv:
## Description

`isaacteleop.rig` — the YAML-driven tmux launcher — and the three `rigs/*.yaml` it drove are unused. Nothing in the tree imports the module, and every setup they launched is documented standalone. Removing them takes a 245-line reference page, three test modules and a wheel subpackage with it.

One fact only the rig documented is kept rather than dropped: the C++ full-body recorder's output directory, the one `replay_full_body.py` defaults to, moves into the body-tracking page. The same page now also names the CloudXR env the full-body readers need, which the rig used to source per pane.

Breaking: `isaacteleop.rig` leaves the wheel. Start the runtime with `python -m isaacteleop.cloudxr` and each producer/consumer yourself. No deprecation window — the module has no known users and no stability promise.

No related issue.

## Type of change

- [x] Documentation update

## Testing

- `cmake -B build` configures clean; `ctest -N` lists 71 tests and no `rig_*`.
- `sphinx-build -b html -W --keep-going` succeeds — no dangling toctree entry, no dangling `:ref:` to the removed page.
- Grep leaves only unrelated English uses of the word ("rigid", "multi-actuator rigs", the MuJoCo rig).

## Checklist

- [x] I have added tests that prove my fix/feature works (or explained why not) — pure deletion; the build, `ctest` and the `-W` docs build are the proof

## Summary by CodeRabbit

- **Breaking Changes**
  - Removed the Rig Launcher, including its command-line interface, configuration support, and bundled rig definitions.
  - Rig-based workflows for full-body tracking, robot visualization, and SE3 tracking are no longer available.
  - CloudXR must now be started separately where required.

- **Documentation**
  - Updated full-body tracking and recording instructions to use direct runtime commands.
  - Clarified that unmatched collection or tensor identifiers result in no samples rather than errors.
  - Removed obsolete Rig Launcher references and documentation.


jiwenc-nv:
related to #888

docs/source/device/add_device.rst
```diff
@@ -221,7 +221,8 @@ the collection and prints samples. Pattern (see :code-file:`examples/schemaio/pe
 4. Loop: call ``session->update()``, then read ``tracker->get_data(*session)``. If the
    returned handle is non-empty, use the latest sample; otherwise sleep briefly and repeat.
 
-Use the same ``collection_id`` (and optionally ``tensor_identifier``) as the plugin. See
+Use the same ``collection_id`` (and optionally ``tensor_identifier``) as the plugin. A
+mismatch is not an error — the printer simply never receives a sample. See
```
sgrizan-nv: How is this related to the rig removal?

sgrizan-nv (approved):

docs/source/device/body_tracking.rst
```diff
@@ -203,10 +203,8 @@ joint ``is_valid`` flags to determine which joints have valid poses.
 For a minimal C++ reader see ``examples/schemaio/full_body_printer.cpp``, which
 creates the tracker, queries the required OpenXR extensions, and prints the
 joint data each frame through ``DeviceIOSession``. The Python equivalent is
-``examples/oxr/python/test_full_body_tracker.py``. Running
-``python -m isaacteleop.rig rigs/full_body.yaml`` runs this printer and the C++
-MCAP recorder together in one tmux window, against the CloudXR runtime it makes
-sure is serving first (see :ref:`rig-launcher`).
+``examples/oxr/python/test_full_body_tracker.py``. Both read the runtime
+directly, so start the CloudXR runtime and connect the headset first.
```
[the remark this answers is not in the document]
jiwenc-nv: Agreed, and fixed in 90feb1347 — this was a real gap my rewrite introduced.

The rig sourced the CloudXR env in every pane, so deleting it and saying only "start the runtime first" lost the requirement. Both readers create `OpenXRSession` directly rather than embedding `CloudXRLauncher`, so they need `XR_RUNTIME_JSON` / `NV_CXR_RUNTIME_DIR` in their own shell.

The passage now follows the pattern the other device pages already use (`wuji_glove.rst`, `manus.rst`, `haptikos.rst`): service start, `source ~/.cloudxr/run/cloudxr.env`, the binary, plus links to `dedicated-cloudxr-runtime` and `load-cloudxr-environment-variables`.

I left out the build commands — `build_from_source` already covers building and installing the example binaries, and repeating that here would be a second place to keep current.

docs/source/device/add_device.rst
jiwenc-nv: Fair challenge — it is the one line in the diff that is not a deletion or a consequence of one.

The reason it is here: the removed files were the only place in the tree that documented this failure mode. `rigs/se3_tracker.yaml` carried it as a comment ("producers and consumers rendezvous on this ... a mismatch is SILENT no-data") and `docs/source/references/rig.rst` repeated it in an `.. important::` block. `add_device.rst` already said "use the same `collection_id` as the plugin" but never said what happens when you do not — and the answer is nothing at all, no error, which is exactly the part worth writing down. So the line is a rescue, not new content.

Happy to drop it and land it separately if you would rather this PR be pure deletion — just say which.

aristarkhovNV (approved):

docs/source/device/add_device.rst
jiwenc-nv: removing.  AI is a bit crazy :(

jiwenc-nv: Done — reverted in c4a1416a8. `add_device.rst` is back to its base state and the PR is now pure deletion plus the doc sites that pointed at the deleted files.

For the record, so it is not silently lost: the `collection_id` mismatch failure mode is no longer documented anywhere once this lands. I can open a small follow-up for it, or leave it — your call.

PR #1094, opened 2026-09-12
feat(schema): give raw hand fingertips their own keyed SE(3) pose schema

jiwenc-nv:
## Description

Raw glove fingertips have no schema of their own, so the Manus plugin shipped its five SE(3) tips as 35 scalar `JointState` entries named `j0`..`j34`, where `j[7i:7i+7]` held one pose. The contract was a doc comment and a slicing rule rather than a type, and `JointStateOutput` means "one actuated DOF, rad or m" — a quaternion component is neither, and `JointStateRetargeter` nominally accepted data it cannot interpret.

`JointSe3PoseOutput` carries them as what they are: a sparse vector of poses keyed by `JointName`, sorted so `LookupByKey` works, where a joint absent from the vector means untracked. `JointName` is allocated in blocks (1100-1199 raw hand tips; 1000, 2000 and 3000 reserved for OpenXR hands and full body) so values never collide across tracker families. The table structure mirrors `joint_state.fbs`, the schema it pairs with: "joint" is the XR sense here — a tracked frame, as in `hand.fbs` and `XrHandJointEXT` — and the robotics sense there, an actuated DOF.

Manus pushes it on the same `manus_sensors_left` / `manus_sensors_right` collections with tensor identifier `joint_se3_pose`; hosts read a tip by name. Each collection carries exactly one hand, so side stays out of the payload.

Schema-side note worth a reviewer's eye: `JointSe3PoseOutput::Pack()` emits the keyed struct vector with `CreateVectorOfStructs` and does **not** sort it, so writers must build through `CreateVectorOfSortedStructs` or the reader's `LookupByKey` searches an unsorted vector. Both writers here do; a test pins it.

Additive on disk: a new golden, no existing `.bfbs` changes, so recordings already made stay readable.

No linked issue.

## Type of change

- [x] Documentation update

The breaking part is narrow and deliberate: the tensor identifier on `manus_sensors_*` moves from `joint_state` to `joint_se3_pose`, so any out-of-tree consumer slicing `j0`..`j34` off those collections stops resolving. Nothing in-tree consumed them.

## Testing

- `ctest`: 372/372 pass on Linux/aarch64 (Jetson AGX Orin), 8 OpenXR cases skipped for want of an HMD.
- `[schema_conform]` failed as intended before the golden existed and passes after; `schema_golden_update` added `golden/joint_se3_pose.bfbs` and modified no other golden.
- New: `test_joint_se3_pose.cpp` / `.py` (lookup, sortedness, sparse miss, 32-byte struct size) and a `ReplaySession` round-trip covering recording traits → MCAP → `check_schema_compat()` → generated tracker for both hands. The round-trip was mutation-checked: collapsing the two hands onto identical data fails it.
- `SKIP=check-copyright-year pre-commit run --all-files` clean.

**Not verified here:** the Manus plugin itself never compiled — no Manus SDK on this host, so configure reports "Skipping Manus plugin build". The FlatBuffers sequence from `push_sensor_side()` was extracted verbatim and compiled + run against the generated headers with a stub transform (five tips found by key, `UNKNOWN` misses, 212 B buffer), but the file needs a build on an SDK host. Live-glove checks — tips physically plausible, left/right not swapped — are also outstanding.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added named, sparse fingertip pose tracking for Manus devices.
  - Added Python access to joint pose data, including joint lookup, device IDs, and timestamps.
  - Added recording, replay, and live visualization examples with configurable collections, playback rate, looping, and output paths.

- **Documentation**
  - Updated sensor output guidance to use named joint poses and direct fingertip lookups.
  - Added references for recording and replay workflows.

- **Tests**
  - Added coverage for serialization, lookup behavior, device streams, replay, timestamps, and empty or missing data.


src/core/schema/python/joint_se3_pose_bindings.h
```diff
@@ -0,0 +1,105 @@
...
+        "from joints is not tracked.")
+        .def(py::init(
+                 [](const std::vector<JointSe3Pose>& joints, const std::string& device_id)
+                 {
+                     JointSe3PoseOutputT native;
+                     native.joints = joints;
+                     // The wire contract is sorted-and-unique; sort here so Python callers cannot
+                     // hand LookupByKey a vector it would silently mis-search.
+                     std::sort(native.joints.begin(), native.joints.end(),
+                               [](const JointSe3Pose& a, const JointSe3Pose& b) { return a.joint() < b.joint(); });
```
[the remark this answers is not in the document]
sgrizan-nv: +1

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
```diff
@@ -973,34 +974,25 @@ void ManusTracker::push_sensor_side(bool is_left, core::SchemaPusher& pusher)
...
-        };
-        for (int k = 0; k < kManusSensorPoseFloats; ++k)
-        {
-            auto joint = std::make_shared<core::JointStateT>();
-            joint->name = "j" + std::to_string(sensor * kManusSensorPoseFloats + k);
-            joint->position = pose[k];
-            joint->valid = true;
-            out.joints.push_back(std::move(joint));
-        }
+        joints.emplace_back(kManusSensorJoints[static_cast<size_t>(sensor)],
```
sgrizan-nv, line 982: Since this is the actual SDK-facing conversion path, can we build and smoke-test the Manus plugin against the real Manus SDK before merge?

The PR description says this translation unit was not compiled on the test host because the SDK was unavailable. The schema/replay tests are useful, but they don't validate this mapping itself.

For the smoke test I'd specifically verify:
- left/right are not swapped;
- sensor index -> thumb/index/middle/ring/little is correct;
- position units/frame are correct;
- quaternion ordering is correct.

src/core/schema/fbs/joint_se3_pose.fbs
```diff
@@ -0,0 +1,80 @@
...
+    joint: JointName (key);
+
+    // Position [m], orientation quaternion [x, y, z, w], in the producer's frame -- which the
+    // producer MUST document, as for Se3TrackerPose.
+    pose: Pose;
+}
+
+// Per-frame output of one tracking device.
+// All fields are always present whenever this table itself is present.
+table JointSe3PoseOutput {
```
sgrizan-nv, line 67: In the Hand API discussion we had roughly:

```text
GenericPoseArray {
    type: TrackerType
    poses: [GenericPose]
}
```

with types such as `HAND_OPENXR`, `HAND_RAW`, `FULL_BODY`, etc.

Here the family is inferred from the `JointName` numeric range and collection identity instead. That works for the current Manus streams, but it also means a single `JointSe3PoseOutput` can structurally contain keys from different tracker families. How do we plan to handle this?

jiwenc-nv: let me add it back.

jiwenc-nv: Good catch — fixed in 7e9fdaee8.

You're right that inferring the family from a numeric range plus collection identity left a mixed payload structurally representable. There's now a `JointType` enum whose values *are* the `JointName` block bases, and `JointSe3PoseOutput` carries `type`:

```
enum JointType : int {
  UNKNOWN = 0,
  HAND_OPENXR = 1000,
  HAND_RAW    = 1100,
  // reserved: FULL_BODY_SMPL = 2000, FULL_BODY_SOMA = 3000
}
```

Two things that fall out of making the values the block bases rather than arbitrary ordinals like the original `TrackerType`:

- `type` is redundant with the keys *by construction*, so it is checkable against them rather than a parallel claim that can quietly drift.
- It still names the family when the vector is empty. Sparse means absent here, so an all-untracked frame has no keys to infer anything from — that case is why this is a field and not just a derivation.

One trap worth flagging for anyone deriving a family later: the blocks are not a uniform width (100 for hands, 1000 for bodies), so `value / 100 * 100` is wrong for body joints. The rule is "the greatest `JointType` not exceeding the value". That's stated in the schema and pinned by a test.

`type` took `id: 0`, renumbering `joints` and `device_id`. Normally off-limits, but this schema has never shipped — its golden was added in this same PR — so no recording carries the old ids. I confirmed the conform test rejects the renumbering *before* refreshing the golden, rather than refreshing to make it quiet.

src/core/schema/fbs/joint_se3_pose.fbs
```diff
@@ -0,0 +1,98 @@
...
+    HAND_RAW_RING_TIP = 1103,
+    HAND_RAW_LITTLE_TIP = 1104,
+}
+
+// One keyed joint pose. 32 bytes, 4-byte aligned, no internal padding.
+struct JointSe3Pose {
+    // Lookup key for the joints vector.
+    joint: JointName (key);
+
+    // Position [m], orientation quaternion [x, y, z, w], in the producer's frame -- which the
```
jiwenc-nv, line 60: @aristarkhovNV @sgrizan-nv @lotusl-code i think this is probably NOT ideal for every producer to describe their own reference frame.  but seems we don't have a good way to align with it.

src/core/schema/fbs/joint_se3_pose.fbs
sgrizan-nv: This makes sense to me.

One small thought: since type is now explicitly carried and the relationship to the `JointName` block is checkable, would it be worth validating that the keys actually belong to the declared `JointType` in the writer / construction path?

Not necessarily a blocker, but it could make the invariant a bit harder to violate accidentally rather than relying only on the schema comment.

src/core/schema/fbs/joint_se3_pose.fbs
sgrizan-nv: Yeah, I agree that having every producer define its own reference frame doesn't scale very well.

Would it make sense for the reference-frame convention to be part of the `JointType` contract where we can define one, rather than the individual producer? Then something like `HAND_OPENXR` would have one well-defined convention, while `HAND_RAW` could explicitly remain vendor-native when there isn't a meaningful common frame.

Not sure we need to solve a general frame/transform system in this PR, but it may be useful to distinguish "defined by the pose type" from "defined by each producer".

src/plugins/manus/core/manus_hand_tracking_plugin.cpp
jiwenc-nv: did a manual smoke tests for manus.  all four points checked out.

src/core/schema/fbs/joint_se3_pose.fbs
jiwenc-nv: the simplest and most effective way of doing this is to map all into the world space defined by `OpenXR`.  A couple successful examples:

1) hand: we always require pllugin developers to fuse glove data into the world frame (starting with manus). 
2) pico full body tracking does the same thing.

the only reason I'm a bit reluctant to standardize on this: what if there is no HMD in the loop?

docs/source/device/manus.rst
sgrizan-nv: nit: The detailed sensor section is updated to `JointSe3PoseOutput`, but the page intro still says the flex-sensor tip poses are published as `JointStateOutput` tensors.

sgrizan-nv (approved):

src/core/schema/fbs/joint_se3_pose.fbs
sgrizan-nv: I think OpenXR world space makes sense for the normalized/OpenXR-facing types, but I'm not sure we can require it for all raw data.

For example, `HAND_RAW` may come from a device that has no HMD/OpenXR world reference at all, or it may intentionally represent the vendor-native measurement frame. Requiring the producer to transform that into OpenXR world space would also make the "raw" dataset depend on an additional calibration/fusion step.

jiwenc-nv: Let's handle that in a separate discussion: https://github.com/NVIDIA-dev/jiwenc-dev/issues/18

examples/mcap_record_replay/python/replay_joint_se3_pose.py
```diff
@@ -0,0 +1,213 @@
...
+    tracked_ids = [
+        cid
+        for cid, ch in summary.channels.items()
+        if ch.topic.endswith("/joint_se3_pose_tracked")
+    ]
+    counts = stats.channel_message_counts
+    best = max((counts.get(cid, 0) for cid in tracked_ids), default=0)
+    if best <= 1:
+        return default
+    return best / (span_ns / 1e9)
```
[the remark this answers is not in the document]
jiwenc-nv: fixed.

src/core/schema/fbs/joint_se3_pose.fbs
jiwenc-nv: added some extra checks.

docs/source/device/manus.rst
jiwenc-nv: fixed.

src/core/schema/python/joint_se3_pose_bindings.h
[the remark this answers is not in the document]
jiwenc-nv: fixed.

src/core/schema/python/joint_se3_pose_bindings.h
```diff
@@ -0,0 +1,136 @@
...
+                     if (duplicate != native.joints.end())
+                     {
+                         throw py::value_error("joints: duplicate JointName " +
+                                               std::string(EnumNameJointName(duplicate->joint())));
+                     }
+                     // One frame carries one family. type is redundant with the keys by
+                     // construction, so disagreement is a writer bug rather than a choice.
+                     for (const JointSe3Pose& joint : native.joints)
+                     {
+                         if (type != JointType_UNKNOWN && joint_family(joint.joint()) != type)
```
sgrizan-nv: Small edge case: with this condition, `type = UNKNOWN` disables the family check, so a non-empty payload with e.g. `HAND_RAW_THUMB_TIP` can still be serialized as `UNKNOWN`.

Since `type` is meant to name the family for the joints in the frame, would it make sense to allow `UNKNOWN` only when `joints` is empty?

jiwenc-nv: updated the check.

